PR #6, opened 2016-04-28
Tseries

raoulcollenteur:
- TseriesBase class added, with additional attributes.
- Stresses are checked for missing values and constant frequency. User
can define frequency and how to fill nan-values.


raoulcollenteur:
Should be a pull request into dev -- not master!

Issue #32, opened 2017-02-14
GXG functions grayed out


tomvansteijn:
Methods for calculating Dutch groundwater statistics GHG and GLG are included in the Statistics class. 
Why are these commented out? Can I add percentile based methods? I have forked the repo.

rubencalje:
I think it was commented out because the GHG- and GLG-method now does not take the frequency of the data into account. We can solve this by sampling the series at the 14th and 28th of every month (which has my preference), or use a percentile-based method like you propose.

tomvansteijn:
I have implemented a percentile-based method in my fork. Needs some testing. Can implement the classic method as well, if you like I can send a pull request  when its ready.

Tom van Steijn MSc
Hydrologist
T +31 88 348 23 34 | M +31 6 54 24 50 79 | E tom.van.steijn@rhdhv.com<mailto:tom.van.steijn@rhdhv.com> | W www.royalhaskoningdhv.com<http://www.royalhaskoningdhv.com/>

raoulcollenteur:
Tom,

The GxG methods were indeed commented out because they need some further work on how these statistics are calculated.. It would be great if you could add percentile based methods and create a pull request for that. I think Pandas has some built-in functions that can be called? 

Cheers, Raoul

raoulcollenteur:
Closed the issue as I think it is all solved for now.

tomvansteijn:
Good. You can let me know if there is more work to be done on this issue or another.

PR #34, opened 2017-02-15
added percentile based GXG methods to stats.py issue #32

tomvansteijn:


raoulcollenteur:
Hi Tom,

First of all, thanks for this pull request and the stats methods you added. I just had a sneak peak and they look good, well-documented and tested (reminding me I should write some tests for the other methods..:-))

I see you have branched of from the master branch. However, all work is done in the development branch (dev) and we only accept PR's on the branch. The master branch only holds the current (stable) version. This means you have to submit your changes to the dev branch of your own fork, and then create a pull request from that to the dev branch of this repo.

As you'll see we have made quite some changes in the meantime. Most notably the get_methods for there time series. These are now 1) implemented in the Model class and 2) not private methods anymore. But I think you can get it all working with some refactoring easily.

I'll have some time to look further into the code tomorrow!

PR #35, opened 2017-02-16
added quantile based GXG methods as per issue #32

tomvansteijn:
I kept the stats.summary() method. Remove it when you feel like.


raoulcollenteur:
I checked out your changes, and most of them seem to be working fine. the summary method was just for convenience so it's OK to keep it / get it back in if you like it :)

Before merging, three things are necessary:

We try to follow the PEP8 guidelines for naming methods, lowercase and words separated by underscores (e.g. gGHG => g_ghg).

It would be nice if the standard signature of the stats methods remains the same, so please change:

`def qGVG(self, key='simulated', tmin=None, tmax=None)`

to:

`def qGVG(self, tmin=None, tmax=None, key='simulated')`

Finally, travis indicates some errors in the final test: https://travis-ci.org/pastas/pastas/builds/202169122

Let me know if you need some help on any of this!

tomvansteijn:
Ok thanks! I have made the changes. Travis crashed because I used pytest instead of nosetests for my test. I have removed the pytest variable.

tomvansteijn:
Honestly I cannot figure out how to let Travis find the data file I use for test_gxg and test_qgxg. It does seem to find the files in your tests, and the filepath I use is almost the same. Any advice?

raoulcollenteur:
Found it! it's actually not in your code but in test_003.py that changes the paths somehow.. I will discuss with @rubencalje tomorrow to see if we can change / delete it (since it is quite slow) and then perform the merge.

I saw you added some private methods. We had a discussion last week and decided not to distinguish between them but I feel more and more the need to start using them. Will get back to you about this as well 👍

rubencalje:
Yes, Raoul was right, you can blame me. test_003 changed the folder. I fixed it with a commit in the dev-branch.

PR #47, opened 2017-05-30
debugging test_gxg and test_qgxg

tomvansteijn:
Changed to .loc indexing for gvg and qgvg calculation (method inspring and lambda the14or28).

Pandas supports boolean indexing a Series or DataFrame using a boolean series with a valid index in combination with .loc, or a boolean array of equal length in combination with .iloc. Apparently I used an Index object, containing booleans, in combination with .iloc, and Pandas 2.0 did something different with that.


tomvansteijn:
python2.7 fails because of new syntax in __init__.py

PR #69, opened 2018-06-12
Add string representation to pastas Model

Hugovdberg:
To get a quick summary of a pastas Model I added a string representation to the Model object. When printing the model it shows the class name followed by some basic properties of the model.


Hugovdberg:
There are actually two functions that can be used in this case: `str` and `repr`. Calling these functions calls the respective dunder method on the object. The default definition on objects is loosely as follows (default string for `repr` partly pseudo coded):

```python
class object:
    def __repr__(self):
        return '<{cls} at memory {mem}>'.format(cls=self.__class__.__name__, mem=memory_location)

    def __str__(self):
        return self.__repr__()
```

By overriding `__repr__` the output of both `str` and `repr` is updated. `repr` is meant for a more formal representation that ideally could be used to recreate the object, which is not viable in this case. `str` is meant for a nicely formatted, more informal, output. I think the extended output with the stressmodels and perhaps even settings should be part of the `__str__` method and not the `__repr__`.

PR #96, opened 2019-03-20
Fix pastas.stats.ccf 

dbrakenhoff:
Issue: 
- Original CCF calculated correlation for both positive and negative lag at the same time
- Resulting CCF was scaled by CCF.abs().max() meaning there was always a correlation of +/-1.0 even if series could not match perfectly.

Fix:
- CCF now only calculated for positive lag by setting a part of the matrix on which the bins are calculated to NaN
- Remove scaling by CCF.abs().max()

A Notebook in the examples/notebooks directory shows the old and new behavior.


raoulcollenteur:
Just checked it and it all still works the same for the ACF so merging the commit. Thanks for the contribution !

PR #98, opened 2019-03-22
fix datetime x axis error in results plot

OnnoEbbens:
a horizontal line is added to the model result plot:
`ax2.axhline(0.0, color='k', linestyle='--')`

The `.axhline()` uses default values `xmin=0` and `xmax=1` therefore changing the dtype of the x-axis from datetime to some nummerical value. This causes errors later on if you want to plot timeseries on the same axis.

I solved this by plotting the horizontal line after the timeseries are plotted.


raoulcollenteur:
Looks good. PR merged. Thanks for the contribution Onno!

PR #101, opened 2019-04-01
Update rfunc.py

saklop:
Included  two extra response functions; the FourParam and the DoubleExponential. The FourParam includes a step function calculated using quad instead of the numerical approximation for testing purposes. The default setting for the FourParam response is to use the Simpson integration to reduce calculation time.


raoulcollenteur:
Nice! I'll look at it today or tomorrow. It still needs some PEP8 formatting, especially limit the number of characters (79) per line. Can you change that? And is there an example notebook? Or maybe change the response function notebook that is already there. Thanks!

mbakker7:
I agree that it would be nice to extend the current response function
Notebook and show that the new functions work, and also show that the
four-parameter function can do the same as the Gamma and Hantush functions
(by fixing one parameter) and can do better when all parameters are
optimized.

saklop:
I have created a new commit for the rfunc file. The file should now be in PEP8 format. I will upload a notebook where the Four Parameter function is compared to the other response functions.

raoulcollenteur:
Looks good and ready for a merge once the notebook has been added and works :)

saklop:
I have slightly changed the rfunc file and added the FourParamQuad response function. This is available for testing but has a much longer calculation time. Also added two new notebooks. One with a synthetic example for the noise model testing and one which compares the Gamma and Hantush response functions with the new Four Parameter response function.

raoulcollenteur:
Everything seems to work. I saw you also added a very interesting notebook on the noise model. Cool! I will merge this pull request now. I will later add these classes to be automatically imported when importing Pastas in the init file.

Thanks for your work!

PR #109, opened 2019-05-16
Time standardisation

rubencalje:
As discussed on the meeting of the 8th of May (see [the presentation](https://docs.google.com/presentation/d/1NWe86KxzMd2_lTNHpBdYFXYgdKm380i-ubj96JXBup4/edit?usp=sharing)), here are some changes to standardise some time-properties of a model. These changes make sure:

- that the noise_alpha has the unit days, just like the warmup;
- that the parameter-values do not change when the frequency of the model is changed;
- that the user still has the option to convert series to daily values when needed (for example monthly well-extractions), with the 'to_daily'-setting.

### DIFF
--- a/pastas/noisemodels.py
+++ b/pastas/noisemodels.py
@@ -142,14 +142,12 @@ class NoiseModel(NoiseModelBase):
             Series of the noise.
 
         """
-        dt = get_dt(freq)
-        odelt = res.index.to_series().diff() / pd.Timedelta(dt, "D")
+        odelt = res.index.to_series().diff() / pd.Timedelta(1, 'd')
         odelt = odelt.iloc[1:]
-        alpha = parameters[0]
         noise = pd.Series(data=res)
+        alpha = parameters[0]
         # res.values is needed else it gets messed up with the dates
         noise.iloc[1:] -= np.exp(-odelt / alpha) * res.values[:-1]
-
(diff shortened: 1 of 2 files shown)


rubencalje:
Any comments? If not, I will merge this at the end of this week.

raoulcollenteur:
Why is there a new method to_daily in the TimeSeries class? Can this not be done by sample_up or sample_down?

rubencalje:
The sample_up and sample_down-settings are changed by this pull request so that they do not change the unit of the stress-series anymore.

The to_daily-method is to standardise the unit of a series that contains quantities (for example monthly extraction-values in m3/month). It recalculates the values to a daily unit (for example m3/day), but does not change the timestep of the series. The to_daily method also works for irregular timesteps. We do not need to know if we have to sample up or sample down. We choose a daily unit, as a day is used as a standard unit in several places (like the warmup, and with this pull request also the noise_alpha).

So it only changes the unit, not the timesteps. The to_daily setting is introduced so that a user can still use his monthly series of quantities in a stressmodel, as we discussed during the meeting on the 8th of may.

Now that I write this, maybe we can rename the setting and the method to 'to_daily_unit'. What do you think?

rubencalje:
I changed to_daily to to_daily_unit. If there are no more comments I will merge this pull request next week.

PR #127, opened 2019-07-25
Branch onno

OnnoEbbens:
I changed the 'add_recharge' method in project.py and added 'solve_models' and 'add_models' so you can apply these methods to all models at once instead of one model at the time. Changes:
- 'add_recharge' now has the option to add recharge to multiple models at once instead of just a single model.
- 'add_models' was added to add multiple models from the available oseries
- 'solve_models' was added to solve all models in the project at once with an option to ignore solve errors.

Besides changed default parameter bounds of pmin and pmax for rfunc One to -100 and +100 when up is None.


raoulcollenteur:
Hi Onno, thanks for your PR with some nice additions! Couple of small things to do before we can merge:

**Add_models method**

- [ ] Use the add_model code in the for-loop, now there is twice the code.

**add_recharge method**

- [ ] Move the method up after the other add methods

**All methods**

- [ ] Should we use `mls` instead of `ml_list`?
- [ ] Does it actually make sense to use the Model object or should we move to model name?
- [ ] Apply PEP8 formatting everywhere


Final note on the other changes:
> Besides changed default parameter bounds of pmin and pmax for rfunc One to -100 and +100 when up is None.

Let's do this in a separate PR request.

raoulcollenteur:
I changed a couple of things and did some commits to the PR. You'll have to pull first. Can you check if it all still works? 

What do you think about using model names instead of using model objects?

OnnoEbbens:
Thanks for checking and the changes you made.

- [ ] I changed `ml_list` to `mls` 

- [ ] I also agree with using model names instead of model objects in the `add_models`, `add_recharge` and `solve_models` methods. 

- [ ] I choose to return model objects for the `add_models` method because `add_model` also returns a model object. Looking closer at the other `add_..` methods I see that `add_model` is the only one returning something. That makes me wonder is it not a better idea to remove the return statement from the `add_model` and `add_models` methods? I did not add this to the PR because it will break backward compatibility.

Issue #108, opened 2019-05-09
Changing Hantush step function for WellModel


dbrakenhoff:
As discussed at the Pastas meeting, I would like to propose changing the Hantush step function to allow individual stresses in a WellModel to have a different "steady-state" influences as determined by the distance to the observation point. In the current implementation each well has the same steady-state influence regardless of its distance to the observation point. 

I'm posting this to check if there are other changes that need to be made as a result of this adjustment? This changes the value of the parameter A (as compared to the other response functions in Pastas), which perhaps should be described somewhere. 

The adjustment would change
 ```python
def step(self, p, dt=1, cutoff=None):
    ...
    return p[0] * F / (2 * k0rho)
````
to
 ```python
def step(self, p, dt=1, cutoff=None):
    ...
    return p[0] * F
````

raoulcollenteur:
Sorry for the late response. Can we implement the redial distance `r` as a parameter, that is fixed by WellModel but free by default? If so, I think this would be the preferred.

`A` should be the value of the step response as time goes to infinity. Is this possible as well?

Also, I have noticed the block response is not smooth with some parameters combinations, seen that before? E.g.,

`import pastas as ps
import matplotlib.pyplot as plt
b = ps.Hantush()
plt.plot(b.block([4, 5, 10]))`

Right after the peak the curve is not smooth..

dbrakenhoff:
In my previous attempt at something along these lines, I added `r` as an optional parameter in Hantush by checking the length of the passed array of parameters is. (see https://github.com/pastas/pastas/blob/52f049185439d8cbbed8bcb4c7f673f0427f4328/pastas/rfunc.py#L288) Which I agree is a bit vague.

If we add `r` as a parameter, how do we deal with that in the solver? The solver won't be able to distinguish between `r` and `lambda` under normal conditions (when both are free). When the WellModel is used, and `r` is set to a fixed value that isn't a problem.

dbrakenhoff:
> Also, I have noticed the block response is not smooth with some parameters combinations, seen that before?

FYI, the non-smooth block response for those parameters has been that way for all implementations of Hantush we've had so far. Not sure if that puts you at ease, but that's all I know...

> `A` should be the value of the step response as time goes to infinity. Is this possible as well?

[...]

PR #129, opened 2019-07-29
Add WellModel

### LINKED ISSUE #102: WellModel randomly results in different fit results
I've been trying to implement the WellModel in a project I'm working on. The idea is to put all the wells in the same aquifer in one WellModel which then calculates one response function for all the wells which is only scaled by the distance from each well to the observation point. 

The way I've done this is by passing an extra parameter to `pastas.Hantush` containing the distance to the well _r_. The parameter _rho_ passed to leastsquares is now equal to 1 / _lambda_, and is multiplied by the distance _r_ when `pastas.Hantush.step()` is called. 

I've made an [example notebook](https://github.com/pastas/pastas/blob/wellmodel/examples/notebooks/wellmodel_consistency_problem.ipynb) that highlights the problem. I've created simple models with one Well with and without recharge I've tried adding the well with StressModel and with the WellModel. The model with a WellModel and recharge randomly yields different fit results (see figure at the end of the notebook). I haven't been able to establish why this happens. It would be great if we could get this to work somehow... So any ideas are welcome!

dbrakenhoff:
Add WellModel to pastas:
- adds WellModel that supports fitting multiple stresses with one response (Hantush) scaled by radial distance
- implements suggestion from #108 
- closes #102 (has not re-occurred on my PC)


raoulcollenteur:
I will look at the code later today and probably merge the PR (just had a quick look and it al looks fine). I just need to read the issues #201 and #108 a bit and compare to the formulas we used in the Pastas paper and check if the notebook still works.

Do you have example notebook to go with this PR? That would be a nice addition :)

dbrakenhoff:
I modified the existing example_Wellmodel.py to show how it works. I think I can easily convert that into a notebook if that is preferred over a Python script.

pastas/model.py
```diff
@@ -595,6 +595,12 @@ def initialize(self, tmin=None, tmax=None, freq=None, warmup=None,
         # Initialize parameters
         self.parameters = self.get_init_parameters(noise, initial)
 
+        for icol in self.parameters:
```
raoulcollenteur: This seems unnecessary so I will remove it in a following commit.

raoulcollenteur:
As I was afraid this PR changes the result of the example 2 of the groundwater paper, which is not something we want. Most likely this is caused by the scaling of the step response.

Maybe we best add a new response function to rfunc.py?

dbrakenhoff:
Do you think the boundaries of the parameter or the different value of `A` are impacting the optimization of `A`? Because otherwise there is no reason for the result to change... I only moved a factor that wasn't contained in `A` into the parameter.

I can't solve the example from the paper with the LmFitSolver because of the covar error you mentioned elsewhere in #133. But I do get a fairly similar result when solving with LeastSquares and multiplying the parameters and stderr by the factor 2*k0(rho). Left is my result, and right is the old result. 

image

Having said all that, I do think a new rfunc is a good idea though, that way all the weirdness caused by the WellModel can be contained in a separate class. I think that's clearest from the user perspective. It can inherit from Hantush and only needs a extra few lines of code. If you agree, I'll set it up.

dbrakenhoff:
As you can see I went ahead with your suggestion for a new rfunc class. It's better this way, as the differences between Hantush and HantushWellModel are now more obvious (instead of being hidden in if-statements in the code).

raoulcollenteur:
Nice! I'll check it now and hopefully merge soon :)

raoulcollenteur:
Okay all looks good and ready to merge! The notebook looks very good now, really like the stack_plot. This could become a nice new plotting method acutally! Where all contribution are stacked. I think at some point we might want to provide the stresses from a csv-file and move away from the Menyanthes file but that can be done later. One small question before merging:

- What is the use/advantage of the sort_wells argument?

dbrakenhoff:
Theres no real need to sort the wells, I just wanted a nice list ordered by distance for myself and I figured others might too, so i added it as an option. I have a few models with quite a few wells in the WellModel and then it help keep things a little more organized if they're sorted by distance.

raoulcollenteur:
> Theres no real need to sort the wells, I just wanted a nice list ordered by distance for myself and I figured others might too, so i added it as an option. I have a few models with quite a few wells in the WellModel and then it help keep things a little more organized if they're sorted by distance.

Okay makes sense. I added the keyword to the dump method. In general all arguments used in the init method of a stressmodel need to be written in the dump method :)

Merging now.

PR #130, opened 2019-07-30
change default parameters for steptrend

OnnoEbbens:
Firstly I propose to add an option to set the parameter bounds in rfunc `One` to `-100` and `100`. Now there is only an option `up=True` to set them to `0` and `100` or `up=False` to set them to `-100` and `0`. You need the parameter bounds to be `-100` and `100` to fit positive and negative trends. At the moment you have to set the initial parameters manually.

Secondly, I propose to change the default parameter bounds of the `StepModel` to this new setting. Because by default you want to fit both negative and positive trends, even if you expect a positive trend. For example, if you think a positive trend will occur and you only try to fit positive trends you will end up with only positive trends even though at some points there is a negative trend. This might give you the wrong impression that only positive trends occur.

### DIFF
--- a/pastas/model.py
+++ b/pastas/model.py
@@ -102,7 +102,7 @@ class Model:
         }
 
         if constant:
-            constant = Constant(value=self.oseries.series.mean(),
+            constant = Constant(initial=self.oseries.series.mean(),
                                 name="constant")
             self.add_constant(constant)
         if noisemodel:
--- a/pastas/rfunc.py
+++ b/pastas/rfunc.py
@@ -150,16 +150,16 @@ class Gamma(RfuncBase):
     def get_init_parameters(self, name):
         parameters = DataFrame(
             columns=['initial', 'pmin', 'pmax', 'vary', 'name'])
-        if self.up is None:
-            parameters.loc[name + '_A'] = (1 / self.meanstress, 
-                                          -100 / self.meanstress,
-                                           100 / self.meanstress, 1, name)
-        elif self.up:
+        if self.up:
             parameters.loc[name + '_A'] = (1 / self.meanstress, 0,
                                            100 / self.meanstress, 1, name)
-        else:
+        elif self.up is False:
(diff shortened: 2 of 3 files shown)


raoulcollenteur:
Why not use -np.inf and np.inf or np.nan? Absolute values would depend on the units of the stress or oseries.. I think the best default option would be no lower/upper bounds (e.g. np.nan) depending on the `up` argument.

Would that work for you as well?

raoulcollenteur:
That said. We need to allow all response functions to allow `up=None`. Cause StepModel can take any of the response functions..

OnnoEbbens:
> Why not use -np.inf and np.inf or np.nan? Absolute values would depend on the units of the stress or oseries.. I think the best default option would be no lower/upper bounds (e.g. np.nan) depending on the `up` argument.
> 
> Would that work for you as well?

For me it will work but I think this might cause significant issues with the solving algorithm because you create an unbounded problem. If we implement this we should test thoroughly first therefore I did not implement this yet.

> 
> 
> That said. We need to allow all response functions to allow `up=None`. Cause StepModel can take any of the response functions..

I did this, I am not 100 % sure if the dump method was adjusted correctly.

raoulcollenteur:
> For me it will work but I think this might cause significant issues with the solving algorithm because you create an unbounded problem. If we implement this we should test thoroughly first therefore I did not implement this yet.

This should not be a problem, as we already do that with the Constant for each Model. Can you change to upper and lower boundary to np.nan?

As most of the times up==True, can you change the order of the if/elif/else statement to:

```
if self.up:
  #set lower bounds here
elif self.up is False:
  #set upper bound here
else:
  #set np.nan bounds here
```

raoulcollenteur:
I just made some changes to Constant and One. If you can do the other rfuncs we can merge the PR.

raoulcollenteur:
I'll make the changes now.

Issue #133, opened 2019-07-31
Make the solver an attribute of the Model class


raoulcollenteur:
I would like to store the Solver object as an attribute of the Model class. If something goes wrong now within the call to the solver we can't access the whole object which is annoying for debugging. This will also help in developing new solvers I plan to work on next month.

This would also mean we don't have to initiate a solver object every time we solve. This will only be an internal change for now.
Within the solve method it would look something like this:

```
self.solver = Solver(args)
self.solver.solve(args) or ml.solver.minimize(args)
```
and then we can access the solver as:

```
ml.solver
```

Anyone against this proposal?

Ps. looks like I'll need to block a day or two to do some programming on Pastas next week so other enhancements I could implement in the solver are welcome. :)

raoulcollenteur:
Possibly add the long standing issue #40 to this.

raoulcollenteur:
lmfit 0.9.13 seems to cause problems when MinimizerResult.covar does not exist. Needs to be fixed as well.

dbrakenhoff:
Semi-related to this: can we add the objective function value to the fit report? 

I did this in my WellModel PR because of the earlier issues with optimization, and this makes it easy to compare between two solve results. It's also not weird to provide the value of the objective function in the fit report, I think. And there was a nice space for it next to the type of solver in the fit report.

raoulcollenteur:
I'll start a new branch for these changes in a bit. Let's make a To Do list. 

- [ ] Store solver as attribute of model
- [ ] Add objective function value top fit_report
- [ ] Fix covar bug for LmFit
- [ ] ....

mbakker7:
And move objective function out of solver class

dbrakenhoff:
I was discussing some pastas results with Ruben (specifically the solver quitting after 2 iterations resulting in evp = 0.0 when solving with freq="14D", while solving on a daily basis yields an EVP of 95%) and we figured it would be nice to have an option in the solver to log the parameter values for each iteration.

So for example, if the log_level was set to debug, log the starting values and the subsequent parameter values least squares decides on, and also maybe log some of the output that least squares generates after it's done (e.g. parameters hitting bounds etc.)

raoulcollenteur:
This is already possible by using using the callback function implemented in the Solver but you'd have to define your own callback method. I used this for the example you mentioned.

dbrakenhoff:
Cool, that works.

raoulcollenteur:
[...]

PR #138, opened 2019-08-19
Solver

raoulcollenteur:
This PR solves many of the issues mentioned in Issue #133. The following things are changed:

1. Store solver as attribute of model (ml.fit)
2. Add solve method to solvers (ml.fit.solve)
3. Fix covar bug for LmFit (ml.fit.pcov)
4. Calculate parameter correlations from covariance matrix (ml.fit.pcor)
5. Add correlations to fit_report (added)
6. Make all "vary" parameters real booleans (no more 0's and 1's)

I chose not to implement to objective function as a separate function, we can do this in another branch/project as it involves some more work that can be done separate from this PR.

### DIFF
--- a/pastas/model.py
+++ b/pastas/model.py
@@ -88,6 +88,7 @@ class Model:
         self.constant = None
         self.transform = None
         self.noisemodel = None
+        self.solver = None
 
         # Default solve/simulation settings
         self.settings = {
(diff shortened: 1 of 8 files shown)


pastas/model.py
```diff
@@ -1367,22 +1385,25 @@ def fit_report(self, output="full"):
...
-
-        report = "{header}{basic}{parameters}".format(
-            header=header, basic=basic, parameters=parameters
-        )
+        if output == "full":
+            cor = dict()
+            pcor = self.fit.pcor
+            for idx in pcor:
+                for col in pcor:
+                    if (np.abs(pcor.loc[idx, col]) > 0.5) and (idx != col) \
```
dbrakenhoff: I set the value here to 0.5 (was 0.3), as stated in the description in fit_report() later on. Was 0.5 or 0.3 the intended value?

raoulcollenteur: Good point! 0.5 was the intended value. I think we can work on the representation of these values later and make it an input argument.

PR #140, opened 2019-08-26
get `tmin` from `ml.settings` if not provided to `ml.solve()`

OnnoEbbens:
Currently, if I set `tmin` in `ml.settings` and use `ml.solve()` without providing the `tmin` argument the model is not solved with the `tmin` from the settings, instead a new `tmin` is calculated from the oseries. Current behavior is:

```
# Set tmin
ml.settings['tmin'] = pd.Timestamp('2010-1-1')

# Solve
ml.solve()

# get tmin
print(ml.settings['tmin'])
>>> Timestamp('1985-11-14 00:00:00')
```

I think this is undesired, so I modified `model.py`. If `tmin is None` the `tmin` from the `ml.settings` is used (if `ml.settings['tmin'] is not None`). Also I added an example for this. Not sure if this is the best way, maybe I should make a test?

New behavior will be:
```
# Set tmin
ml.settings['tmin'] = pd.Timestamp('2010-1-1')

# Solve
ml.solve()

# get tmin
print(ml.settings['tmin'])
>>> Timestamp('2010-1-1 00:00:00')
```


raoulcollenteur:
Not sure what is wise here. Why do you change the settings dict directly and not use the tmin argument in the solve method? Editing the settings dict manually should be prevented...

OnnoEbbens:
I want to create a lot of models in bulk from a file with model settings. I prefer making the models first with all settings and then solving them all at once with the `Project.solve_models()` method.

raoulcollenteur:
but then you can use the tmin argument in Project.solve_models() right? Otherwise this should be made possible.

I am not really keen on supporting manual changes of the settings dict, it is hacky as nothing is checked when this is done, bringing a lot of unexpected errors (e.g. try ml.settings["tmin"]="2010"). We could create set methods for the settings dict, but that would involve some more work.

raoulcollenteur:
Another thing. This PR changes the default behaviour by using the stored tmin and tmax. Although this default might make sense, we should then implement it for all settings.

Issue #141, opened 2019-08-27
Making values from ml.settings default when nothing is provided for solve?


raoulcollenteur:
@OnnoEbbens proposed to use the tmin and tmax stored in the settings dict when no values for these arguments are provided in the solve method (see PR #140). This would change the default behaviour we have right now, which is to derive it from the data.

This would allow manual settings of the settings dict. Doing that bypasses any built-in checks and causes (weird) errors downstream, so I in my opinion this should be discouraged. We could support this, but then I think we need to subclass the dict class and create a enhanced settings dictionary with specific set_methods.

This is a lot of work and the benefit is minimal I think. Setting the settings through solve arguments should remain the way it is done.

dbrakenhoff:
I see your point about not modifying ml.settings directly. But consider this example:

We currently have a project in which
- We have data from an online source we do not control, so we store our modifications (i.e. removing certain periods) elsewhere. This is done by storing the observations series name and the tmin/tmax and removed periods in a different table. 
- We save the pastas models without timeseries, as it needs to be possible to update the timeseries in models if the user so desires. When the models are loaded, the timeseries are looked up in a pastas.Project. 
- When we run our analysis, we load the models, using either the old or updated timeseries, and if necessary optimize the models. In this situation it's nice to have all your custom settings applied to the model immediately, without having to think about passing stuff to each ml.solve().

I guess a solution would be to modify pr.solve_models() to accept tmin and tmax arguments that contain the tmin/tmax per model. 

Still would be nice to have the option of "saving" your customized tmin/tmax/settings with your model instead of having to provide them to the model each time you want to optimize.

(Perhaps this question is something we can discuss at a developer's meeting...)

raoulcollenteur:
> I guess a solution would be to modify pr.solve_models() to accept tmin and tmax arguments that contain the tmin/tmax per model.

this method already takes tmin and max through the kwargs, so this should be possible right?

dbrakenhoff:
I think this can be closed?

PR #144, opened 2019-08-28
improve solve_models() with tmin/tmax options

dbrakenhoff:


dbrakenhoff:
implements partial solution for question in #141.

Hugovdberg:
This could in fact be made more general quite easily by checking the `kwargs` for `pandas.Series` objects.

```python
m_kwargs = {}
for key, value in kwargs.items():
    if isinstance(value, pd.Series):
        m_kwargs[key] = value.loc[ml_name]
    else:
        m_kwargs[key] = value
# Convert timestamps
for tstamp in ['tmin', 'tmax']:
    if tstamp in m_kwargs:
        m_kwargs[tstamp] = pd.Timestamp(m_kwargs[tstamp])

[...]
ml.solve(report=report, **m_kwargs)
```

Also, is there any reason the new arguments are not forced to be passed as keywords ([PEP 3102](https://www.python.org/dev/peps/pep-3102/)) since Python 3 is required anyway?

dbrakenhoff:
Good suggestion! We also realized this was something we should eventually make more general, but since I only needed tmin/tmax at the moment, I got lazy. 

Feel free to submit a PR with this change, otherwise I'll implement your suggestion sometime later this week.

PR #147, opened 2019-08-29
Add tests for pastas.Project

dbrakenhoff:
Add tests for pastas.Project bringing coverage up to about 80%.
Found some minor bugs while writing the tests.

@rubencalje can you check if my fix in 77e76751046b6379d68fafad4d98f5177ad796f9 is what you would expect? When loading a Project from file, the load would crash when attempting to read the head series. The data is not read as a pandas.Series but a pastas.TimeSeries, which results in an error when you attempt to access the index in order to call tz_localize(). 

I'm not sure if this is only needed for a pandas.Series or should also be done for a pastas.TimeSeries...

### DIFF
--- a/pastas/io/pas.py
+++ b/pastas/io/pas.py
@@ -35,7 +35,8 @@ def pastas_hook(obj):
                     obj[key] = TimeSeries(**value)
                 except:
                     obj[key] = value
-            obj[key].index = obj[key].index.tz_localize(None)
+            if isinstance(obj[key], Series):
+                obj[key].index = obj[key].index.tz_localize(None)
         elif key in ["time_offset", "warmup"]:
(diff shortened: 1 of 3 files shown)


rubencalje:
> @rubencalje can you check if my fix in 77e7675 is what you would expect? When loading a Project from file, the load would crash when attempting to read the head series. The data is not read as a pandas.Series but a pastas.TimeSeries, which results in an error when you attempt to access the index in order to call tz_localize().

Seems ok. I added the obj[key].index = obj[key].index.tz_localize(None) line as pandas 0.25.0 would give an error while reading a model from a pas-file. I should have encountered the same error as you though. Maybe the try-except statements before this line has something to do with it...

PR #149, opened 2019-08-30
fix `solve_models`

OnnoEbbens:
removed tmin and tmax as keyword argument in `solve_models` because they should be read as kwargs


raoulcollenteur:
I already changed this (including removing them from the docstrings in the dev-branch. Closing this issue.

PR #150, opened 2019-09-20
add stacked_results plot

dbrakenhoff:
add stacked_results plot
add stress name to contributions calculation for WellModel
update WellModel notebook


dbrakenhoff:
@raoulcollenteur I explicitly do not split the contributions for StressModel2. I wasn't sure if I needed to add additional checks for other StressModels that also contain multiple stresses... Or maybe we _do_ want to split StressModel2 as well... Can you verify that this implementation is ok?

Issue #156, opened 2019-10-11
parameters polder function


mbakker7:
As far as I know, the polder function is still experimental and needs some work. 
Three issues:
1. Right now, the parameters are chosen as x/(2*lab) and 1/cS. I am not sure that is the most logical, although it gives the most obvious link with aquifer parameters. Anybody against experimenting a bit with these?
2. Right now, there is no scaling factor. I understand why: in a semi-infinite aquifer, the ultimate rise is a function of the rise in the river and the aquifer parameters. But this function is intended to be used as well in other situations, for example finite aquifers. So I suggest to add a scale factor, like the other response functions.
3. I will implement a conservative (but quick) estimate of tmax.

rubencalje:
The idea about the current implementation of the Polder Function (instead of Gamma for example) is that you can get some properties of the system. It then helps if there are less parameters.

I changed the Polder function from 3 to 2 parameters, also because I got an email from Wilbert Berendrecht, that he had problems with the scaling factor: the parameters were too correlated. I never really used the Polder function however (because of point 3), so I cannot say it is ok the way it is now, and so I am not against experimenting. Maybe we can add the scaling factor optionally?

mbakker7:
Ruben,

I think the scaling factor should be in there. For one thing, this function
should work for general settings, not just for a semi-infinite aquifer. I
understand what you are saying, that the parameters are highly correlated
(which may be true) and if you want to get physical parameters out then it
is preferred to fix the scaling factor to 1. My suggestion is to add the
scaling factor, but, for now, fix it to 1.

Mark

rubencalje:
Hi Mark,

The calculation for tmax seems to work well, as is fixing of the scaling factor to 1 (by default the scaling factor is now not fixed by the way).

The initial value of the 'b'-parameter is set to 0.1 (like it always has been). This causes a very slow response, see [this graph from the docs](https://pastas.readthedocs.io/en/latest/_images/examples_003_response_functions.ipynb_3_0.png) for example. With the initial parameters, there is allmost no influence from my varying river level, and the optimisation is unable to find an influence I am pretty sure exists. Therefore I think we should set the initial value of this parameter to 1.

[...]

PR #159, opened 2019-10-22
Add maxtmax argument

rubencalje:


pastas/stressmodels.py
```diff
@@ -225,6 +225,12 @@ def get_nsplit(self):
             return self.nsplit
         else:
             return len(self.stress)
+        
+    def get_block(self, p, dt, tmin, tmax):
+        """Internal method to get the block-response from the respnse function"""
+        maxtmax = (tmax-tmin)/pd.to_timedelta(1,'d')
+        b = self.rfunc.block(p, dt, maxtmax=maxtmax)
+        return b
```
raoulcollenteur: I think it would be better to rename this to "get_block_response" or simply "block". And then maybe we refactor step to step_response as well while we are at it.

Also I think refund.block requires the cutoff?

pastas/rfunc.py
```diff
@@ -851,7 +857,7 @@ def get_tmax(self, p, cutoff=None):
     def gain(self, p):
         return 1.
 
-    def step(self, p, dt=1, cutoff=None):
-        t = self.get_t(p, dt, cutoff)
+    def step(self, p, dt=1, cutoff=None, maxtmax=None):
```
raoulcollenteur: Mark this as an internal method just like the block?

pastas/stressmodels.py
```diff
@@ -225,6 +225,12 @@ def get_nsplit(self):
             return self.nsplit
         else:
             return len(self.stress)
+        
+    def get_block(self, p, dt, tmin, tmax):
+        """Internal method to get the block-response from the respnse function"""
+        maxtmax = (tmax-tmin)/pd.to_timedelta(1,'d')
```
raoulcollenteur: This involves another hardcoding of daily in Pastas.... just noting.

raoulcollenteur:
Why is this change necessary actually? #156 #159

pastas/stressmodels.py
rubencalje: This is in stressmodels.py, so there is no step here. I use get_block() instead of block() to show the difference between the block-method in rfunc.py. The rfunc gets the cutoff from self, as it was passed earlier already from the stressmodel.

pastas/rfunc.py
rubencalje: This is another file (rfunc.py instead of stressmodels.py)

rubencalje:
> Why is this change necessary actually? #156 #159

Yes, because of issue #156

rubencalje:
Strange error in Travis CI: RuntimeError: Kernel didn't respond in 60 seconds

In the pr-check of Travis this build is passed. I think this pr is ready to be merged.

Issue #74, opened 2018-06-12
Make private method private with leading underscore


raoulcollenteur:
I think we should make all non-public methods known by adding a leading underscore to the method name. This is suggested in PEP8 (https://www.python.org/dev/peps/pep-0008/), and is followed by all major packages (E.g. Numpy, Scipy, Pandas, Flopy). E.g. 

`Model._get_odelt()`

Also, we should not use private methods of other packages in Pastas, like the _base_and_stride taken from Pandas in utils.py. These methods can be dropped without notice, which can cause problems in future versions of Pastas.

This will make it more clear for the Pastas-users which methods they should use (They will pop up first on tab completion) and make maintenance of Pastas easier in the future.

Thoughts?

Hugovdberg:
There is actually a subtle difference between single and double underscores in front of the method name. Single underscore methods can be overriden by inherited objects (much like the `protected` keyword in other programming languages), while double underscore methods cannot be overridden.

Also, it isn't just part of PEP8, which just gives some basic style guides, and effectively says to use Python the way it was designed to be used. It is also part of the [Python Language Reference](https://docs.python.org/3/reference/lexical_analysis.html#reserved-classes-of-identifiers).

Using the underscores is an idiomatic way to keep the programming interface clean and keep flexibility in the implementation of the internals. As Raoul said most, if not all, major packages and programming tools respect and perhaps even expect this part of the language definition to be used.

I would definitely vote in favor of using the programming language's default method of what methods and functions are part of the external programming interface and what is not.

pastas:
Although Raoul and Hugo make good points, I think it is a pain in the butt
to use private methods with very little benefit. It rarely happens in
Pastas that anybody wants to call a function that they shouldn't. But that
is just my 5 cents.

Hugovdberg:
The problem is that it is unclear which methods they shouldn't use, except for the fact that they are marked as 'internal method' in the docstring.

I don't see how typing an underscore or two is a pain in the butt. What are the bad experiences you've had with private methods that make you think this way?

Hugovdberg:
Has anyone already started on making these interface changes? If not I hereby volunteer to implement this. I haven't heard compelling arguments against it so I'm assuming this is the way we want to go.

mbakker7:
My reason for not using private methods is that I don't see the advantage
while we are making things purposely obscure.

I don't think there is any problem now with people trying to access methods
that they shouldn't. So it doesn't solve a problem. So why 'hide' them now
[...]

PR #161, opened 2019-10-25
Uncertainty

raoulcollenteur:
Initial methods to add uncertainty calculations to Pastas Models.

For example, uncertainty calculations are called as follows after model optimisation:

```
ml.fit.prediction_interval()
ml.fit.ci_simulation()
```

- An example is added (example_uncertainty.py)
- Tests are added (test_solvers.py)


examples/example_uncertainty.py
```diff
@@ -0,0 +1,41 @@
...
+sm = ps.RechargeModel(prec=rain, evap=evap, rfunc=ps.Exponential,
+                      recharge="Linear", name='recharge')
+ml.add_stressmodel(sm)
+
+# Solve
+ml.solve()
+#
+df = ml.fit.prediction_interval()
+inside = (obs > df.loc[obs.index, 0.025]) & (obs < df.loc[obs.index, 0.975])
+print('results of prediction_interval2')
```
dbrakenhoff: I assume prediction_interval2 has become the default.

pastas/model.py
```diff
@@ -1253,8 +1253,31 @@ def get_transform_contribution(self, tmin=None, tmax=None):
         sim_org = ml.simulate(tmin=tmin, tmax=tmax)
         return sim - sim_org
 
+    def get_response(self, rfunc, name, parameters=None, dt=None, **kwargs):
```
dbrakenhoff: use of rfunc as variable for block or step is a bit confusing here?

dbrakenhoff: How elaborate are our docstrings for internal methods? And do we hide internal methods by using the prefix _?

examples/example_uncertainty.py
raoulcollenteur: Good point. pred2 has been removed. I'll change this.

pastas/model.py
raoulcollenteur: This is an internal method. I would love to change all internal methods to the method name preceded by a single or double underscore but this is another issue #74 which should not be addressed in this PR.

dbrakenhoff: maybe just the more general `func`? Can't think of anything better at the moment.

raoulcollenteur: changed the argument name to "block_or_step" which is a little long but more explicit. Also added a docstring for the method.

rubencalje:
Looks good. I do not have time to look into the code in detail though. Personally I would keep 'model' instead of 'ml', but if this helps to be consistent throughout Pastas it's better like this.

dbrakenhoff:
@mbakker7, could you review this PR (if you haven't already)? I'd like to start using/testing it from the dev branch.

Issue #152, opened 2019-09-26
first entry in ml.noise() is set to 0


dbrakenhoff:
I came across the following result as I was optimizing my timeseries models:

![bad_fit](https://user-images.githubusercontent.com/6552225/65669769-6ea1b480-e044-11e9-8408-16c8038c3169.png)

This occurs because the first entry of the noise calculation is set to 0 in the code on line 153 (in combination with a large `noise_alpha`):
https://github.com/pastas/pastas/blob/21a8464f4fc1b47cc67af9108cefc6069e146df4/pastas/noisemodels.py#L148-L155

This first value of the noise (before changing it to 0) is equal to that ca. 2m offset between the model simulation and the observations. Commenting out this line ensures that value of `constant_d` is such that the model is on top of the observations.

My question therefore is if this setting the first value of the noise to 0.0 is the correct way to go? Would removing that line cause any problems in other situations?

rubencalje:
We used to set the first value of the noise to 0, because we used the multiply method, and had to assign a fill-value. Now that Raoul has optimized the code, the line 'res.iloc[0] = 0 ' takes the form of the fill_value.

I think we can (and should) remove this line, so that the first value of the noise is equal to the first value of the residuals. In this way, all of the model error is taken into account during the optimization. I think this solves a small part of the problem we have with bad fits using a noisemodel.

raoulcollenteur:
Good point. I actually only put in that line to ensure I didn't change the outcome compared to the old code.

Leaving the residual values for the first time step might be an idea, although that also means the beginning of the simulation becomes very important. I thought @mbakker7 had some thoughts on a good first value for the noise?

rubencalje:
I think we should address this before the next release. therefore I made a pull request: #163

raoulcollenteur:
I think we should not change anything to the current noise model with regard to this. I think the noise model works fine as is, and where it does not work either 1) the noise model is not appropriate and another noise models is needed (e.g., second-order) or 2) the model structure is inappropriate (as is like in the example given by @dbrakenhoff here).

Just to be clear from my experience, I actually never see the problem above anymore once I change the model structure or gwl observation frequency. To me this strongly suggests that the problem lies elsewhere. I suggest we work on new noise models instead of making ad hoc changes, such as the one proposed by @rubencalje in PR #163. As suggested in the PR discussion, this could become a new noise model we can test.

We could then change the default noise model in the future once we agree on a "better" noise model.

rubencalje:
[...]

PR #163, opened 2019-10-30
Keep first entry in noise

rubencalje:
This pull request addresses issue #152

### DIFF
--- a/pastas/noisemodels.py
+++ b/pastas/noisemodels.py
@@ -150,7 +150,6 @@ class NoiseModel(NoiseModelBase):
         # res.values is needed else it gets messed up with the dates
         v = res.values[1:] - np.exp(-odelt / alpha) * res.values[:-1]
         res.iloc[1:] = v * self.weights(alpha, odelt)
-        res.iloc[0] = 0
         res.name = "Noise"
         return res
 


raoulcollenteur:
I am very, very strongly against this suggested change, as it fundamentally changes the objective function (to something that is wrong in my opinion and masking important underlying problems with the noise model and optimization).

If you want to use such a noise model I suggest to add a NoiseModel3 class or something similar, and write in the docstring that this is meant for testing).

rubencalje:
ok, that is a bit dramatic. It is just one value, where we have no information on previous errors, so we assume all error is noise . Right now the first model error in time is ignored, which can cause a large residual for the rest of the simulation for a large value of noise-alpha. I think we should take the first error into account for our objective function, just like we do for residuals after a large gap in a measurement series. The noise-value just after the gap is then also equal to the residual.

In NoiseModel2 we already do this, so there is no point in adding another noisemodel.

raoulcollenteur:
Also, the solution might be in a different objective function, for example by adding the average value of the residuals, but this is separate from the noise model.

rubencalje:
> In NoiseModel2 we already do this, so there is no point in adding another noisemodel.

I see NoiseModel2 was changed. In the master-branch we keep the first residual as noise, like I want. In the develop-branch however, the first value of the Noise in NoiseModel2 is also set to 0...

PR #168, opened 2019-11-21
Update master to v0.13.0

raoulcollenteur:


rubencalje:
Hi Raoul, great that a new version is coming out! I would like to take some time to do some tests next week, to see if older code still runs, before we release this version.  Can you wait that long?

raoulcollenteur:
Sure no problem. If there are some features you are using that are not tested yes we should maybe add them to the automatic testing scheme.

I might try and fix issue #169 in the meantime before merging then, but not necessarily.

raoulcollenteur:
Can I merge this PR tomorrow and update the master branch?

PR #173, opened 2020-03-06
Recharge

raoulcollenteur:
I would like to merge two new non-linear recharge models into the development branch. These models are developed to work together with the `RechargeModel` stress model. The first is a new model I developed and the second is the non-linear model developed by Berendrecht et al. (2006). The models are highly optimised using (Numba compiled) Numpy code for the actual computations. Installing Numba is highly recommended but NOT necessary.  Summarizing this commit includes:

- Two new non-linear recharge models, documented.
- Decorator that optionally enables Numba JIT (checks whether or not Numba is installed)
- Jupyter notebook introducing the new non-linear models.


rubencalje:
Hi Raoul, is it possible to explain in one or two sentences or formulas in the notes how the recharge models work, just like in the Linear model, so users do not have to read the entire publications?

I see the preferred unit of precipitation and evaporation is mm/d. Can we change this to m/d? The knmi methods for example generate precipitation and evaporation in m/d.

raoulcollenteur:
Hi Ruben, I added the basic formulas to the [docstrings](https://pastas.readthedocs.io/en/recharge/API-docs/pastas.recharge.recharge_func.html) (after 5 commits when I figured this annoying math thing out). I hope this helps users get a basic understanding.

I see the point with the units. I initially started with m/d, but this causes large differences in the values between parameters of several orders of magnitude (e.g., large rch_A 500, and small rch_ks, 0.005). Using mm/d ensures that the parameters are in the same order of magnitude. This makes the parameter estimation easier (although it shouldn't matter much when your method rescales automatically) so I would like to keep it this way.

mbakker7:
Guys,

I see now that some of the parameters may be significantly lower than 1.
This causes a problem with least_squares, as the derivative is incorrectly
computed, as explained here: https://github.com/scipy/scipy/issues/11258

I am not sure this has been fixed in the latest scipy. Something to bear in
mind. Lmfit and the old leastsq don't have this problem.

Mark

raoulcollenteur:
In the current setup (mm/d) the parameters are not significantly lower than one, so I suspect this would not be a problem.

ps. I can imagine this being a problem in other cases though, for example a groundwater pumping rate defined in million cubic meter per year, well_A would become significantly smaller than 1.

raoulcollenteur:
I uploaded a template for PRs to make it easier to check if a PR contains all the necessary code changes. This PR misses tests. I will add these

## Short Description

## Checklist before PR can be merged:
- [X] ~~closes issue #xxxx~~ / new feature
- [X] API is documented
- [X] PEP8 compliant code
- [ ] tests added / passed
- [X] passes Travis
- [X] Example Notebook (for new features)

examples/example_recharge.py
```diff
@@ -18,17 +18,21 @@
...
 
-# Create stress
-sm = ps.RechargeModel(prec=rain, evap=evap, rfunc=ps.Exponential,
-                      name='recharge')
 ml.add_stressmodel(sm)
 
 # Solve
-ml.solve()
-ml.plot()
+ml.solve(noise=False, report=False)  # This usually helps
```
dbrakenhoff: Maybe add explanation why solve is run twice?

raoulcollenteur:
Made some final changes, should be good now?

PR #179, opened 2020-03-17
Improved documentation website

raoulcollenteur:
This PR bring a totally new documentation website to Pastas, using a new theme. On the new website is a lot more clear where to find things due to better navigation options for the theme. The API docs have been strongly improved, summarising different modules and classes and providing a better overview of all the methods and classes available in Pastas (see e.g., https://pastas.readthedocs.io/en/improve_docs/api/statistics.html). These are now focused on Users instead of developers.

Basically no information was removed, and only information was added. There are still some sections to fill (help welcome) but most of the work is done. Most importantly, the new website allows us to better structure information and make more information available over time.


rubencalje:
Looks good! I think this will help people a lot. Two remarks:

- On my phone the menu button does not do anything. Do you know why this is the case?
- Some items are empty, like the about section. Maybe we could remove those for now?

doc/about/faqs.rst
```diff
@@ -0,0 +1,27 @@
...
+To cite a specific version of the Pastas software we recommend using the the
+DOI proivded through Zenodo for every major release of the software.
+
+- Collenteur, R., Bakker, M., Caljé, R. & Schaars, F. (XXXX). Pastas:
+  open-source software for time series analysis in hydrology (Version X.X.X).
+  Zenodo. http://doi.org/10.5281/zenodo.1465866.
+
+.. rubric:: What does "Pastas" stand for?
+
+Pastas is not an acronym for something. However we like a good Pasta at it's
```
dbrakenhoff: tiny dutchism I think: "at it's time"
Maybe: "we like a good Pasta every once in a while"?

pastas/read/menyanthes.py
```diff
@@ -24,6 +24,18 @@
...
+    Parameters
+    ----------
+    fname:
+    locations
+    datatype
+
+    Returns
+    -------
+
+    """
```
dbrakenhoff: to be completed

pastas/stressmodels.py
```diff
@@ -659,16 +659,15 @@ def simulate(p=None):
...
+    """
+    Convolution of one or more stresses with one response function.
 
     Parameters
     ----------
     stress : list
         list containing the stresses timeseries.
     rfunc : pastas.rfunc
-        WellModel only works with Hantush!
+        this model only works with Hantush!
```
dbrakenhoff: Only works with HantushWellModel

pastas/read/menyanthes.py
raoulcollenteur: Actually improving the docstrings was not part of this PR but should off course be done.

raoulcollenteur:
> * On my phone the menu button does not do anything. Do you know why this is the case?

No idea why that happens. Let's hope people don't check on their phone to often..

> * Some items are empty, like the about section. Maybe we could remove those for now?

I removed empty items but mostly filled them.

PR #181, opened 2020-03-19
Add model comparison plot, deprecation warning Project

dbrakenhoff:
- Added deprecation warning for pastas.Project

- Added `ml.plots.compare()` for comparing models on the same figure. The figure is based on `ml.plots.results()`. See example below.
- Added a test for this method to test_plot.py

image


dbrakenhoff:
The failure is being caused by c82df9372f74d2e055a75aee288170d1de953fee which only affects python 3.5 apparently.

raoulcollenteur:
> The failure is being caused by [c82df93](https://github.com/pastas/pastas/commit/c82df9372f74d2e055a75aee288170d1de953fee) which only affects python 3.5 apparently.

Nope, this has been happening a few more times now. Something changed that causes a timeout while running a notebook with Python 3.5.. it should work if you rerun travis

pastas/project/project.py
```diff
@@ -39,6 +39,8 @@
 from ..timeseries import TimeSeries
 from ..version import __version__
 
+from warnings import warn
```
raoulcollenteur: why not use logger.warning?

pastas/project/project.py
```diff
@@ -59,6 +61,10 @@ def __init__(self, name, metadata=None):
             Dictionary with any metadata information on the project.
 
         """
+        warn("The Project class will be deprecated in a future verion of "
+             "Pastas! Consider switching to pastastore: "
```
raoulcollenteur: We need to end a time here I think. End of 2020?

dbrakenhoff: sounds good

pastas/project/project.py
dbrakenhoff: changed

dbrakenhoff:
I'd be fine with it being a separate method. I did it this way because this is how the rest of the plot methods are accessed and thought it might be less easy to find if it was a separate method. Then again, I like the call signature better with `ps.plots.compare([ml1, ml2, ...])`, so I've changed it.

The warn method allows it to be of type DeprecationWarning, but I'll change it to the logger.warning.

Failure in travis is showing the following message:
`TypeError: stem() got an unexpected keyword argument 'use_line_collection'`
Odd, but let's see if it works now though.

raoulcollenteur:
We'll deal with that matplotlib error in the dev-branch. Merging and Closing this PR.

PR #197, opened 2020-05-19
Add FerrisKnowles response function for wells

raoulcollenteur:
# Short Description
This PR adds the FerrisKnowles response function, as described in Shapoori et al (2015). Not sure about how to use  and interpret this function though:
- It has an indefinite integral so it keeps lowering the heads.
- I didn't scale  it yet so no gain parameter.

No idea if we want this, but we can add it as an experimental response function.

## Checklist before PR can be merged
- [ ] is documented
- [X] PEP8 compliant code
- [ ] tests added / passed
- [ ] passes Travis

### DIFF
--- a/pastas/__init__.py
+++ b/pastas/__init__.py
@@ -2,12 +2,12 @@ import logging
 
 import pastas.recharge as rch
 from .model import Model
-from .noisemodels import NoiseModel, NoiseModel2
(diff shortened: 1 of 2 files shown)


mbakker7:
I think we should only add this if the function has a gain. Otherwise it is
like the Theis function, which never reaches steady state, in which case it
is not useful.

Issue #217, opened 2020-07-16
Modify `WellModel` to use new `Hantush`


mbakker7:
The `WellModel` needs to use the new `Hantush` function with parameters `A`, `a`, and `b`

dbrakenhoff:
I'm having trouble setting the optimization bounds using the new `Hantush` implementation.

The trouble is that `get_tmax()` returns infinity when <img src="https://render.githubusercontent.com/render/math?math=\rho = r / \lambda \gt 699">. I'm not entirely sure anymore if the `get_tmax()` was the problem, but in any case, the optimization doesn't work when rho exceeds this value. 

So, to ensure this doesn't happen the bound for parameter b has to be set so that this value isn't exceeded. Since parameter <img src="https://render.githubusercontent.com/render/math?math=a=cS"> is also contained in the value of <img src="https://render.githubusercontent.com/render/math?math=b=r^2 cS / 4 \lambda^2">, the bound of b is also dependent on parameter `a`. This makes it difficult to set the boundary, as the boundary is dependent on the value of one of the other parameters.

So, any ideas on how to solve this conundrum?

mbakker7:
It makes sense that it doesn't work when <a href="https://www.codecogs.com/eqnedit.php?latex=\rho&space;=&space;r/\lambda&space;>&space;10" target="_blank"><img src="https://latex.codecogs.com/svg.latex?\rho&space;=&space;r/\lambda&space;>&space;10" title="\rho = r/\lambda > 10" /></a>, as the drawdown will be negligible, and I can see this giving problems with finding tmax.

I also noticed that it is very difficult to set reasonable bounds for `b`: they can indeed be anything, depending on a. 

Considering what you are saying, it may be better to write the general four-parameter function as 

<a href="https://www.codecogs.com/eqnedit.php?latex=A&space;t^{n-1}&space;\exp(-t/a&space;-&space;ab/t)" target="_blank"><img src="https://latex.codecogs.com/svg.latex?A&space;t^{n-1}&space;\exp(-t/a&space;-&space;ab/t)" title="A t^{n-1} \exp(-t/a - ab/t)" /></a>

Then we can set reasonable bounds for `b`. The parameter `a` will then have dimensions of time and `b` will be dimensionless.

dbrakenhoff:
That sounds like a good idea. Does that require an adjustment to the calculation of the step in Hantush? Or only in the calculation of `rho`? I tried the latter, and i got a reasonable model result.

[...]

Issue #177, opened 2020-03-14
least_squares problem for parameters that are much smaller than 1


mbakker7:
The default method for parameter estimation in Pastas, `scipy.optimize.least_squares` has a bug when the absolute value of a parameter is much smaller than 1. See here: https://github.com/scipy/scipy/issues/11258
We should do something about that, at least until it is resolved.

raoulcollenteur:
We need to find a working example to test this somewhere.. anyone?

dbrakenhoff:
Check if implementation is correct in Pastas for small parameters

mbakker7:
Scipy > 1.8 doesn't install yet with conda. Gotta wait for that. Bumped the milestone.

raoulcollenteur:
I checked and scipy 1.8 is now available from Conda and Pypi. Bumped Scipy version in last commit so we can close this issue. :)

dbrakenhoff:
I think it's fine closing this, but to double check:

@mbakker7 you had a ttim script where you discovered this issue, can you confirm it is now no longer a problem?

raoulcollenteur:
@mbakker7 and @dbrakenhoff can you please check if we can close this issue and merge PR #496 ?

mbakker7:
This is fixed in scipy 1.8, so as long as we require that, this can be closed.

PR #222, opened 2020-07-20
Update Wellmodel and HantushWellModel

dbrakenhoff:
# Short Description
Update WellModel to use new Hantush.

- [x] closes issue #217
- [x] is documented
- [x] PEP8 compliant code
- [x] tests added / passed
- [x] passes Travis


dbrakenhoff:
I used `LmfitSolve` because of the issue in `least_squares` for small parameter values (as mentioned in #217 and #177).

Apparently lmfit is not installed in the environment on travis. Anyone mind if I add it?

mbakker7:
I think we should add lmfit back as a dependency.
The issue with small parameters in least_squares is a serious problem that may not be fixed in scipy anytime soon (I have no experience submitting PRs in scipy). But it should definitely be added to Travis, otherwise it can never be tested.

pastas/stressmodels.py
```diff
@@ -711,6 +711,10 @@ def __init__(self, stress, rfunc, name, distances, up=False, cutoff=0.999,
             raise NotImplementedError("WellModel only supports the rfunc "
                                       "HantushWellModel fow now!")
 
+        logger.warn("It is recommended to use LmFitSolve as the solver "
```
raoulcollenteur, line 714: Even when when changing the units of Q? If I divide by 1e6 there's often to problem.

mbakker7:
In lmfit, we should make leastsq the default method (which may be the
default for lmfit, but need to check).

pastas/stressmodels.py
mbakker7: "to problem" -> no problem or a problem?

mbakker7:
I made one final change to the Hantush function (just a factor of 2), so that it reaches the correct gain. 
Could you run WellModel with the final version of Hantush? After that (and adding lmfit to Travis) I think it is ready to be merged.

pastas/stressmodels.py
dbrakenhoff: The values of parameters A and b become even lower because the distance (`r`) is set to 1.

You can probably avoid this by changing units as you suggest, but then we'd need to recommend doing so somehow.

dbrakenhoff:
> In lmfit, we should make leastsq the default method (which may be the default for lmfit, but need to check).
In pastas we set the method to "leastsq" which uses scipy.leastsq (https://lmfit.github.io/lmfit-py/fitting.html)

I updated HantushWellModel and reran the notebook. Intriguingly, fitting the extractions separately now leads to a very similar result as the one with WellModel (which it wasn't doing before). That's a good development, but does make the case for implementing the WellModel less persuasive (in the notebook). The explanations in the notebook will have to be updated to reflect that.

dbrakenhoff:
> The issue with small parameters in least_squares is a serious problem that may not be fixed in scipy anytime soon (I have no experience submitting PRs in scipy). But it should definitely be added to Travis, otherwise it can never be tested.

Btw, i saw your issue on scipy was mentioned in a PR: https://github.com/scipy/scipy/pull/11263 
Not sure it includes the fix we need, but it hasn't been accepted yet.

pastas/stressmodels.py
raoulcollenteur: I meant "no problem"

raoulcollenteur:
> In lmfit, we should make leastsq the default method (which may be the default for lmfit, but need to check).
> […](#)

lmfit uses leastsq as default./

Issue #204, opened 2020-06-29
reading Dinoloket Peilschalen (level gauges)


tomvansteijn:
Pastas can only read groundwater levels from the Dutch data portal Dinoloket. Support for reading level gauges (peilschalen) is useful for timeseries analysis in the Netherlands. The CSV format for level gauges is slightly off, therefore a variation of the ps.read.DinoGrondwaterstand class is needed.

mbakker7:
I didn't know DinoLoket had surface water levels. That would be nice.
Can you submit a PR including an example Notebook where you use both groundwater levels and surface water levels?

rubencalje:
That would be nice. A PR would be welcome. If you do not have a scripts for reading Dinoloket level gauges yet, we do have something, but it will take some work to make it ready for Pastas. I can commit the raw script though.

tomvansteijn:
I added DinoPeilschaal in my pastas fork [here](https://github.com/tomvansteijn/pastas/blob/dev/pastas/read/dinoloket.py). However it is bit of an overhaul as I tried to make use of some generalisations for reading this type of files. Needs more testing as well.

tomvansteijn:
Ofcourse it is also possible to refactor and do a PR for the complementary code only, leaving the groundwater level reader class untouched.

mbakker7:
I think that would be the best option: a PR for the complimentary code only
and leaving the groundwater level reader class untouched.

tomvansteijn:
Second try, a new DinoPeilschaal class for reading level gauges. A minimum of changes compared to the groundwater level reader class.

https://github.com/tomvansteijn/pastas/blob/dev/pastas/read/dinoloket.py

Example notebook to be added.

PR #227, opened 2020-07-24
Dinoloket read method

tomvansteijn:
# Short Description
Added support for reading river level gauges in Dinoloket format. Changes in pastas\read\Dinoloket.py and a simple read test.

- [x] closes issue #204
- [x] is documented
- [x] PEP8 compliant code
- [x] tests added / passed
- [x] passes Travis


tomvansteijn:
Ofcourse codacy does not like the duplication in my code. Which is deliberate because I did not want to diverge from the coding style in the original reader class for groundwater levels.

A previous attempt is available in [this branch](https://github.com/tomvansteijn/pastas/blob/dev-read-dinoloket/pastas/read/dinoloket.py). No duplication, but this contains breaking changes.

dbrakenhoff:
I think you need to change the filepaths in the notebook to use the forward slash instead of backslash for the tests to pass.

raoulcollenteur:
@dbrakenhoff can you check the PR? I'll make a new Pastas version the end of this week, would be great if we can add this to the new version!

dbrakenhoff:
As Tom already mentioned, there's a lot of duplication, but that is unavoidable. I figured the notebook would be a good place to elaborate on all the read methods currently in Pastas, so I made some initial changes so the examples for KNMI, Waterbase can be added later. 

p.s. Waterbase was replaced by waterinfo as far as I know. And I don't know if read.waterbase is also capable of reading rws.waterinfo.nl files...? Something to look into.

dbrakenhoff:
Once travis passes, I will merge.

Issue #207, opened 2020-07-03
`ValueError: array of sample points is empty`


RDWimmers:
When downloading KNMI data the timestamp of the data is YYYYMMDDHHMMSS, with HHMMSS as 01:00:00.
The DINO timestamp has HHMMSS as 00:00:00.

When trying to solve the model `ml.solve()` the following error will accrue:

`ValueError: array of sample points is empty`

This error is given in `np.interp()` and relate to the different timestamp. There is no overlapping timestamp between DINO data and KNMI data. So there are no values to interpolate. See attachment.

By removing the HHMMSS the model does compute. `evap.index = pd.to_datetime(evap.index).date`
Maybe it is possible to check the data? Or find the closest value based on the `sim.timestamp`?

Thanks, and keep up the good work.

[dev.zip](https://github.com/pastas/pastas/files/4868808/dev.zip)

I guess this error is somehow similar in the mechanics as issue #55. However, this accusers for every model. Not only the `stressmodel2`.

raoulcollenteur:
Hi RDWimmers,

Thanks for the Bug report. I have quickly looked into it and it's appears to be a bit more complicated. The problem does not occur in `ml.residuals()` but already in `ml.simulate()`. ml.get_stress("recharge") still gives a result.

This is because the sim_index returned by ml.get_sim_index() does not take the offset of 1 hour into account. I'll keep this issue open, but it might take some time before it is addressed. The solution you proposed above is the fix  for now :-)

Cheers,
Raoul

raoulcollenteur:
Short script to test for future reference:

```
import pastas as ps
from pastas.read.knmi import KnmiStation
import pandas as pd


head = pd.read_csv("data/head_nb1.csv", index_col=0, parse_dates=True,
                  squeeze=True)

# fetch data from KNMI station
knmi = KnmiStation.download(stns=260, start="1985", end="2008")
evap = knmi.data["EV24"]
rain = knmi.data["RH"]

ml = ps.Model(head, name="Model", constant=False)
sm = ps.RechargeModel(rain, evap, rfunc=ps.Exponential, name="recharge")
ml.add_stressmodel(sm)

ml.solve(tmin="1982")
```

rubencalje:
The time_offset needs work indeed. You can set the time_offset manually, just before ml.solve(), by

`ml.settings['time_offset'] = pd.to_timedelta(1,'H')`

or let the model set it, by:

`ml._set_time_offset()`

[...]

Issue #231, opened 2020-07-31
[BUG] example_900.py errors


raoulcollenteur:
**Describe the bug**
 example_900.py errors on Travis, but not on my local computer. Should figure out what happens there.

raoulcollenteur:
example_900.py errors after updating to Pandas 1.1.x. No idea why. Anyone wants to fix this?

rubencalje:
It has to do with the _sample_weighted , which does not handle the time-offset correctly. I can fix this, but my fix will probably break other examples (without a time-offset, but with more irregular data). It will take some time to really fix this issue. For now, I will fix the example, by adding the freq_original to the evaporation-timeseries.

rubencalje:
Why this only happens after updating to Pandas 1.1.x I have no idea.

raoulcollenteur:
Fixed as per PR #237

PR #237, opened 2020-10-02
Improve time_offset

rubencalje:
See issue #207 and #231

# Short Description
This pull request finally solves the problem we have with the time-offset and also problems that arise when changing frequency. There are no new features, just existing ones that should work better. For example, you can now use any supported frequency (e.g. '7D') in example_900, so together with the time_offset of 9 hours. Also when a user now tries to use a model with two stresses that contain different time-offsets (for example daily precipitation at 9:00 and daily evaporation at 1:00, which is common in the netherlands), the user will receive an error explaining the problem, so he can solve it. In the past the model would fail because the simulation did not contain any values (see issue #207).

- [x] closes issue #207 and #231 
- [x] is documented (no new features)
- [x] PEP8 compliant code
- [x] tests added / passed (example_900.py)
- [x] passes Travis
- [x] Example Notebook (no new features)
- [x] API changes documented in Release Notes (no new features)

### DIFF
--- a/examples/example_900.py
+++ b/examples/example_900.py
@@ -18,7 +18,6 @@ knmi = ps.read.knmi.KnmiStation.fromfile(
 rain = ps.TimeSeries(knmi.data['RD'], settings='prec')
 
 evap = ps.read_knmi('data/etmgeg_380.txt', variables='EV24')
-evap.freq_original = 'D'
 if True:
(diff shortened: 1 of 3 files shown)


raoulcollenteur (approved):

raoulcollenteur:
Looks good! Gonna merge and close those two issues.

PR #278, opened 2021-03-11
Update pas.py

janjaappape:
Error when trying to write to file. np.integers are not JSON-able

# Short Description

- [ ] closes issue #xxxx
- [ ] is documented
- [ ] PEP8 compliant code
- [ ] tests added / passed
- [ ] passes Travis

### DIFF
--- a/pastas/io/pas.py
+++ b/pastas/io/pas.py
@@ -86,5 +86,7 @@ class PastasEncoder(json.JSONEncoder):
             return o.to_timedelta64().__str__()
         elif isna(o):
             return None
+        elif isinstance(o, np.integer):
+            return int(o)
         else:
             return super(PastasEncoder, self).default(o)


raoulcollenteur:
Hi @janjaappape, thanks for the PR, code suggestions are always welcome! It's maybe better to first make an Issue for this. 

I don't think we use numpy integers anywhere in Pastas actually, so this change may not be necessary. Maybe you used a wrong keyword argument somewhere (e.g. 1 vs. True)? 

Can you create an Issue with an example where this happens?
Cheers,
Raoul

OnnoEbbens:
I've ran into the same problem in the past when some values in my model, in particular metadata, were stored as a numpy type. Before I converted them to a Python standard type (int or float) when creating a pas file. I think it may we worth to allow for numpy types by modifying `pas.py` but I am not sure how this will affect writing and reading pas files. @dbrakenhoff might know more about this.

If we allow numpy types we could use the code below to convert any numpy types to standard Python types (if possible).
        
```
elif isinstance(o, np.generic):
    return o.item()
```

dbrakenhoff:
You can't store numpy dtypes in json, so if it is necessary to convert them we might as well go for as many as possible. Onno's solution looks nice if. But it would be good to see an example of where this goes wrong first to determine whether this fix is a necessary addition to the reading/writing of pas-files.

PR #286, opened 2021-03-23
Hantush

dbrakenhoff:
# Short Description

- [ ] ~closes issue #xxxx~
- [x] is documented
- [x] PEP8 compliant code
- [ ] tests added / passed
- [x] passes Travis


dbrakenhoff:
As discussed the following changes are implemented:

- Hantush changed so that parameter `A = gain` once more (so calculated uncertainty is the uncertainty of the gain)
- HantushWellModel can be used as an rfunc where `gain  = A * k0(sqrt(4*b))`. When `r` is passed in separately (i.e. WellModel) response function can be scaled with `r`. 
- The non-zero bounds for A in HantushWellmodel and Hantush are removed and set to +/- inf
- Update initial parameter estimates when using WellModel
- Add notebook explaining and showing the difference between the Hantush implementations
- Update docstrings, example notebooks with wells
- Some minor improvements in WellModel and `ps.plots.compare()`

Todo:

- [ ] calculate uncertainty for the gain based on uncertainty for parameter A and b for HantushWellModel.

raoulcollenteur:
Shall we move the new notebook to the concepts folder? It's not really an example I think.

pastas/stressmodels.py
```diff
@@ -709,7 +709,7 @@ def __init__(self, stress, rfunc, name, distances, up=False, cutoff=0.999,
                  settings="well", sort_wells=True):
         if not issubclass(rfunc, HantushWellModel):
             raise NotImplementedError("WellModel only supports the rfunc "
-                                      "HantushWellModel fow now!")
```
raoulcollenteur, line 712: 👍🏻

pastas/stressmodels.py
```diff
@@ -762,9 +763,14 @@ def set_init_parameters(self):
...
             np.max(self.distances) ** 2
-        # set initial value with mean distance
+        # set initial value b with mean distance,
+        # filling in mean distance as r will yield 1 as initial value
         self.parameters.loc[self.name + "_b", "initial"] /= \
             np.mean(self.distances) ** 2
+        # set initial value A
+        # divide by k0(2) to get same initial value as ps.Hantush
+        self.parameters.loc[self.name + "_A", "initial"] /= \
+            0.1138938727495334
```
raoulcollenteur, line 773: Maybe just use k0(2)? Probably clearer

raoulcollenteur (changes requested):

raoulcollenteur (approved):

dbrakenhoff:
@raoulcollenteur, would you mind taking a quick peek at the latest additions for the calculation of the uncertainty of the gain? 

Especially whether the addition of the methods to calculate the variance of the gain to the HantushWellModel rfunc and WellModel stressmodel make sense.

pastas/stressmodels.py
```diff
@@ -880,6 +889,52 @@ def get_parameters(self, model=None, istress=None):
             p_with_r = np.r_[p, distances]
         return p_with_r
 
+    def variance_gain(self, ml, istress=None):
```
raoulcollenteur, line 892: Can you remove this method here? I think it would be better to do something in ml.fit.get_ci_step_response(), or make a new function there, get_ci_gain or something, but apart from this PR. Would be good to think about a more general implementation first. Maybe add this code to the notebook?

pastas/stressmodels.py
```diff
@@ -762,9 +764,14 @@ def set_init_parameters(self):
             np.max(self.distances) ** 2
         self.parameters.loc[self.name + "_b", "pmin"] /= \
             np.max(self.distances) ** 2
-        # set initial value with mean distance
+        # set initial value b with mean distance,
+        # filling in mean distance as r will yield 1 as initial value
```
raoulcollenteur, line 768: I think this should be done in HantushWellModel.get_init_parameters(), just pass on r. That also makes it easier to add similar response functions that use a "distance" in the future no?

raoulcollenteur (changes requested):

pastas/stressmodels.py
dbrakenhoff: I'll remove it here and add it to the notebook for now. I'll open a separate issue for uncertainty and (Hantush)Wellmodel.

pastas/stressmodels.py
dbrakenhoff: Good idea, I'll change it with the next commit

dbrakenhoff:
Should be good to go now!

raoulcollenteur (approved):

PR #312, opened 2021-08-19
Warning when optimization hits parameter bounds

dbrakenhoff:
# Short Description
Add warning in fit_report when optimizaion hits parameter bounds.

- [x] ~closes issue~
- [x] is documented
- [x] PEP8 compliant code
- [x] tests passed
- [x] ~Example Notebook (for new features)~
- [x] ~API changes documented in Release Notes~


dbrakenhoff:
I'll merge dev into this PR after #313 is reviewed and merged so this one can be tested before merging.

raoulcollenteur:
There was already a method (`def _check_parameters_bounds`) in the Model class. Should we move the code there?

dbrakenhoff:
> There was already a method (`def _check_parameters_bounds`) in the Model class. Should we move the code there?
> 
> I think this would be nice because then we can also call that method separately and use the results in other places (e.g. perhaps to color the parameters in plots.results.)

Good idea!

dbrakenhoff:
This one should be ready after tests pass!

raoulcollenteur:
Yes I think so too. Can you add a short note in the release notes to finish it?

 https://github.com/pastas/pastas/blob/master/doc/releases/release_0.18.rst

dbrakenhoff:
I edited the release notes directly on dev.

PR #313, opened 2021-08-19
fix HantushWellModel

dbrakenhoff:
# Short Description
- Add support in `HantushWellModel.get_tmax()` for when k0rho --> 0.0 to return large response time (100 yrs), instead of np.inf.
- Relax upper bound for parameter b in HantushWellModel. Bounds used to be scaled with max distance between well and oseries, now the upper bound uses minimum distance instead.

This should allow wells at large distances to be added without constraining parameter b by those large distances.

**Question**: Is 100 years a reasonable large number, or too large?

- [x] closes issue 
- [x] ~is documented~ internal change
- [x] PEP8 compliant code
- [x] tests passed
- [x] ~Example Notebook (for new features)~
- [x] ~API changes documented in Release Notes~


dbrakenhoff:
Just had a thought we could also maybe solve this with `maxtmax` in `get_t()`. I'll take a look...

dbrakenhoff:
The thought was wrong. So no changes.

raoulcollenteur (approved):

PR #326, opened 2021-10-07
Kraijenhoff van de Leur Response

martinvonk:
# Short Description
Kraijenhoff van de Leur Response Function.
Impulse Response:
image
Step Response:
image

This response function has three parameters:
**S** - storativity [-]
**j** - reservoir coefficient [T]
**x/L** - location in a domain between two ditches seperated by distance L
 (**n** - is not actually a parameter but needs a number smaller than infinity to make the function faster. In practice >100 seems fine.)

Notes:
The gain of the response function can be calculated analytically for x/L = 0.5. The gain is then equal to  \pi * S / (8 * S)

I need help with Numba. Without Numba this response function takes forever to run. However, when importing @njit the step function reports `TypingError: non-precise type pyobject`. I tried converting every numpy array in the step function to `dtype=np.float64` but that does not seem to work. I feel like it has something to do with the array type retrieved from `get_t`

- [ ] closes issue #xxxx
- [ ] is documented
- [ ] PEP8 compliant code
- [ ] tests added / passed

### DIFF
--- a/pastas/rfunc.py
+++ b/pastas/rfunc.py
@@ -957,10 +957,10 @@ class Kleur(RfuncBase):
     @njit
     def step_kleur(p, t):
         s = np.zeros(len(t))
-        for i, value in enumerate(t):
+        for i, t_value in enumerate(t):
             s_part = np.array([0.0])
             for n in np.arange(1, p[3], 2):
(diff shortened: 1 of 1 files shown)


dbrakenhoff:
I don't get it either, but maybe has to do with the fact your PR is coming from a fork, and therefore does not have access to codacy?

I wouldn't bother figuring this out tbh, the tests pass, and any remaining style issues can be fixed by an autoformatter later.

PR #332, opened 2021-10-15
Add ChangeModel to dev

raoulcollenteur:
# Short Description
This PR add the ChangeModel to the stress models, implementing (on a basic level) the idea of Obergfjell et al (2019). It differs from that work, in that it can only take one stress (e.g., a river) as an input, not precipitation and evaporation. However, I still think it may be useful to some, and perhaps can be extended in the future to take more complicated functions into account.

- [ ] closes issue #xxxx
- [x] is documented
- [x] PEP8 compliant code
- [x] tests added / passed (through notebook)

### DIFF
--- a/doc/releases/release_0.19.rst
+++ b/doc/releases/release_0.19.rst
@@ -1,0 +1,22 @@
+Version 0.19 (Expected 1st of December 2021)
+--------------------------------------------
+Minor update introducing a few new features and fixes some smaller bugs.
+
+New Features / Enhancements
+***************************
+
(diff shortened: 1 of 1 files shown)


raoulcollenteur:
> Would it in the future also be possible to model the transition from one stressmodel to another stressmodel? In that way, it would also work for recharge-models.

Initially that is what I tried, but this proved very hard to implement due to how we set the parameters in that model (which depend on the response function and the recharge model). I think we can look at that in the future if someone needs that functionality though.

I'll add the API changes documented in Release Notes and merge the PR later today .

Issue #305, opened 2021-07-05
Gamma Impulse Response Function


martinvonk:
The gamma impulse response function in the notebook [response_functions.ipynb](https://github.com/pastas/pastas/blob/master/concepts/response_functions.ipynb) is not the one used by Pastas.

Pastas uses the scaled Gamma distribution ([Collenteur et al, 2019](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6899905/)):
<img src="https://render.githubusercontent.com/render/math?math=\theta(t) = A \frac{t^{n-1}}{a^n \Gamma (n)}e^\frac{-t}{a}">

but the notebook mentions:
<img src="https://render.githubusercontent.com/render/math?math=\theta(t) = A t^{n-1} e^\frac{-t}{a}">

PR #335, opened 2021-11-03
Kraijenhoff van de Leur / Bruggeman 133.15 Response

martinvonk:
# Short Description
Kraijenhoff van de Leur response is now calculated with the equation from Bruggeman 133.15. With the equation from Bruggeman the gain can be obtained without solving the entire equation. Additionally, the response is scaled such that the gain is always equal to A.

Three parameters:
A is the gain (scaled),
a is now the reservoir coefficient [d] (j = S * L^2 / (\pi^2 * K * D)) as described by Kraijenhoff van de Leur
b is the location in the domain.  Note that with Bruggeman the origin is now in the middle of the domain. This means that b=0 is in the middle and b=1/2 is at the drainage channel. At b=1/4 the response function is most similar to the exponential response function.

Additionally, in this PR some typos are fixed in two notebooks such that issue #305 is solved.

- [x] closes issue #305 
- [x] is documented
- [x] PEP8 compliant code
- [x] tests added / passed


martinvonk:
================================================== warnings summary ===================================================
..\..\..\miniconda3\envs\pastas_dev\lib\site-packages\win32\lib\pywintypes.py:2
  C:\Users\vonkm\miniconda3\envs\pastas_dev\lib\site-packages\win32\lib\pywintypes.py:2: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
    import imp, sys, os

-- Docs: https://docs.pytest.org/en/stable/warnings.html

----------- coverage: platform win32, python 3.9.7-final-0 -----------
Coverage XML written to file coverage.xml

pastas/__init__.py
```diff
@@ -10,7 +10,7 @@
 from .read import (read_dino, read_dino_level_gauge, read_knmi, read_meny,
                    read_waterbase)
 from .rfunc import (DoubleExponential, Exponential, FourParam, Gamma, Hantush,
-                    HantushWellModel, Kleur, One, Polder, Spline)
+                    HantushWellModel, Kraijenhoff, One, Polder, Spline)
```
raoulcollenteur, line 13: Mention here: https://github.com/pastas/pastas/blob/dev/doc/releases/release_0.19.rst

pastas/plots.py
```diff
@@ -1017,18 +1017,20 @@ def series(head=None, stresses=None, hist=True, kde=False, titles=True,
...
+                colour = 'C1'
             else:
                 colour = 'k'
             axes[0, 1].plot(gkde.evaluate(ind), ind, color=colour)
         if hist or kde:
             # stats table
             head_stats = [["Count", f"{head.count():0.0f}"],
                           ["Mean", f"{head.mean():0.2f}"],
+                          ["Max", f"{head.max():0.2f}"],
+                          ["Min", f"{head.min():0.2f}"],
```
raoulcollenteur, line 1033: Where f-strings faster than "{:0.2f}".format here?

pastas/rfunc.py
```diff
@@ -881,9 +880,18 @@ def step(self, p, dt=1, cutoff=None, maxtmax=None):
         return s
 
 
-class Kleur(RfuncBase):
-    """The function of Kraijenhoff van de Leur, describing the response of a
-    polder domain with length L between two ditches.
+class Kraijenhoff(RfuncBase):
+    """The function of Kraijenhoff van de Leur scaled. The function describes
```
raoulcollenteur, line 884: Class description should start with a one line description. The rest goes into "Notes" section.

pastas/rfunc.py
```diff
@@ -936,27 +946,17 @@ def get_tmax(self, p, cutoff=None):
...
-                )
-            s[i] = 4 / (np.pi * p[0]) * np.sum(s_part)
-        return s
+        return p[0]
 
     def step(self, p, dt=1, cutoff=None, maxtmax=None):
         t = self.get_t(p, dt, cutoff, maxtmax)
-        return self.step_kleur(np.asarray(p), t)
+        h = 0
+        for n in range(200):
```
raoulcollenteur, line 954: Why 200??

raoulcollenteur (changes requested):

raoulcollenteur:
@martinvonk I think Github CI should work from now on. But let's see once you made the suggested changes. Nice work.

pastas/plots.py
martinvonk: I don't know. I always use this way of formatting strings. 

Found this on https://blog.finxter.com/string-formatting-vs-format-vs-formatted-string-literal/:
> Summary: f-string is more readable and easier to implement than % and .format() string formatting styles. Furthermore, using f-strings is suggested for Python 3.6 and above while .format() is best suited for Python 2.6 and above. Versions prior to Python 2.6 only provide % option for string formatting. In terms of speed, % is the fastest string formatting style. 

I can change it if you want.

mbakker7:
200 is overkill. we should be smart in picking n, as a few terms will
commonly do. In fact, I think that 2 terms is probably sufficient, but I
have not proven that. This should be part of future research. Shall we make
n a parameter that is fixed by default, so we can change it if we want?

martinvonk:
@mbakker7 @raoulcollenteur 

> 200 is overkill. we should be smart in picking n, as a few terms will commonly do. In fact, I think that 2 terms is probably sufficient, but I have not proven that. This should be part of future research. Shall we make n a parameter that is fixed by default, so we can change it if we want?

Working on a short comparison for different values of `n` right now. 

In the equation the term is equal to infinity. However, we can be smarter than that and take a smaller number. The term is however dependent on the reservoir coefficient `a` (`j`). So the longer the response, the more terms of `n` you need. I agree 200 is overkill (for most values of `a`) but 2 seems a bit too short.

I'll share a notebook this morning.

martinvonk:
I think this image is also sufficient instead of a notebook. This is the RMSE of the Kraijenhoff van de Leur step response (using the amount of n terms on the x-axis) compared to the response using 10e5 n_terms.

image



It is actually the other way round of what I said previously. So the larger a, the less terms of n you need. Which is logical for the step response if the gain stays the same.

raoulcollenteur (approved):

pastas/rfunc.py
```diff
@@ -936,27 +948,17 @@ def get_tmax(self, p, cutoff=None):
...
-                )
-            s[i] = 4 / (np.pi * p[0]) * np.sum(s_part)
-        return s
+        return p[0]
 
     def step(self, p, dt=1, cutoff=None, maxtmax=None):
         t = self.get_t(p, dt, cutoff, maxtmax)
-        return self.step_kleur(np.asarray(p), t)
+        h = 0
+        for n in range(10):
```
raoulcollenteur, line 956: As @mbakker7 suggested, can you make `n` a parameter in the function `step_kleur`? So the response function gets an attribute self.n=10 which could be changed for research purposes.

raoulcollenteur:
> 200 is overkill. we should be smart in picking n, as a few terms will commonly do. In fact, I think that 2 terms is probably sufficient, but I have not proven that. This should be part of future research. Shall we make n a parameter that is fixed by default, so we can change it if we want?
> […](#)

@martinvonk can you make n a parameter? Then we can merge this PR.

raoulcollenteur:
I meant as attribute to the class, not as a calibration parameter. You don't want to optimise `n` I think?

martinvonk:
> I meant as attribute to the class, not as a calibration parameter. You don't want to optimise `n` I think?

Sorry my bad. _n_terms_ is now an attribute instead of an parameter.

raoulcollenteur:
> > I meant as attribute to the class, not as a calibration parameter. You don't want to optimise `n` I think?
> 
> Sorry my bad. _n_terms_ is now an attribute instead of an parameter.

The way you programmed it it waas still not possible to change n. I fixed it in my latest commit. 

However, I tried to run example.py with the new response function, and git very bad results. Can you look into that?

martinvonk:
> > > I meant as attribute to the class, not as a calibration parameter. You don't want to optimise `n` I think?
> > 
> > 
> > Sorry my bad. _n_terms_ is now an attribute instead of an parameter.
> 
> The way you programmed it it waas still not possible to change n. I fixed it in my latest commit.
> 
> However, I tried to run example.py with the new response function, and git very bad results. Can you look into that?

The Kraijenhoff van de Leur response is certainly specific and won't give good results for a lot of cases. When saturated (instantaneous) responses such as ps.Exponential gives good results, the Kraijenhoff vd Leur response can give good results too (where x/L = 1/4). An example of this is shown in #326.

However the Kraijenhoff van de Leur response can simulate different behaviour compared to the exponential response such as maintaining the water level a bit longer at a higher level, giving an reversed _s_ shape response. Given that it is a saturated response, my hope is that the Kraijenhoff van de Leur response works better in combination with nonlinear recharge models, where the delay of the unsaturated zone is integrated in the recharge model.

raoulcollenteur:
I see, but I didn't get the same results anymore because parameter bounds were hit. I changed the bounds in the last commit. @martinvonk Can you check if you agree and then we merge this.

martinvonk:
> I see, but I didn't get the same results anymore because parameter bounds were hit. I changed the bounds in the last commit. @martinvonk Can you check if you agree and then we merge this.

Yes. I forgot to adjust parameter A. The equation is now scaled such that parameter A is the gain and thus it should be larger than the original bounds.

Issue #336, opened 2021-11-08
[ENHANCEMENT] improve reindexing to equidistant timeseries in stoffer_toloi


dbrakenhoff:
In the `stoffer_toloi` diagnostic test, the series is reindexed to a equidistant time index using `series.asfreq(freq)`. The trouble with this method is that if the first timestamp from the original series (in this case the noise) is different from all other timestamps, an almost empty timeseries is returned which is then used in the tests, leading to vague errors in some cases. 

The following snippet shows the problem:
```python
freq = "D"
idx0 = pd.date_range("2000-01-01", freq=freq, periods=7).tolist()
idx0[0] = pd.Timestamp("2000-01-01 09:00:00")
series = pd.Series(index=idx0, data=np.arange(len(idx0), dtype=float))
series.asfreq(freq)
```
This results in the following timeseries:
```
2000-01-01 09:00:00    0.0
2000-01-02 09:00:00    NaN
2000-01-03 09:00:00    NaN
2000-01-04 09:00:00    NaN
2000-01-05 09:00:00    NaN
2000-01-06 09:00:00    NaN
Freq: D, dtype: float64
```
When I think we want something along these lines:
```
2000-01-01    0.0
2000-01-02    1.0
2000-01-03    2.0
2000-01-04    3.0
2000-01-05    4.0
2000-01-06    5.0
2000-01-07    6.0
Freq: D, dtype: float64
```

The following snippet is the fix I would propose:
```python
# build new equidistant index
idx = pd.date_range(series.index[0].floor(freq),
                    series.index[-1].ceil(freq),
                    freq=freq)
# get nearest index from original series
sample_idx = ps.utils.get_sample(series.index, idx)
# create a new equidistant series
s = pd.Series(index=idx, data=np.nan)
# fill new series with sampled data from original series
s.loc[idx[:sample_idx.size]] = series.loc[sample_idx].values
# trim any trailing nans
s = s.loc[s.first_valid_index():s.last_valid_index()]
```

dbrakenhoff:
Another simpler option would be to just round the index, drop the resulting duplicates (if there are any) and then use `asfreq`.
```python
series.index = series.index.floor(freq)
series = series.drop_duplicates(keep="first", inplace=False).asfreq(freq)
```

The code is simpler, but you might end up losing a few datapoints.

raoulcollenteur:
@dbrakenhoff can we close this issue now?

dbrakenhoff:
[...]

PR #338, opened 2021-11-11
Add `get_equidistant_timeseries()` method

dbrakenhoff:
# Short Description
Add method for getting equidistant timeseries. Could be useful sometimes, but especially necessary for the `stoffer_toloi` test (see #336). This test was using `pd.Series.asfreq` to get an equidistant timeseries of the noise but in certain situations this dropped a lot of data, resulting in a calculation being performed on a subset of the noise. This new method addresses that, but it does perform some nearest interpolation. The old method did not interpolate, and only sampled existing data points that occur at the user-specified frequency. 

I'm still looking for some input, as I'm not entirely sure whether we would prefer the rather complicated method I added, or something simpler using pandas methods with a few limitations as shown in the notebook. Also not sure if we want a notebook showcasing this method necessarily. So suggestions/feedback welcome on this PR! 

- [x] closes issue #336
- [x] is documented
- [x] PEP8 compliant code
- [x] tests added / passed


raoulcollenteur (approved):

raoulcollenteur:
Stoffer-toloi is for missing data, whereas many time series are irregular. One way to deal with the irregular time steps is to interpolate and get missing data time series. Another approach would be to conclude that the stoffer-toloi is not applicable. Yet another method would be is to select a smaller time step through the `freq` argument (perhaps this is the best). 

I would prefer to make the interpolation in the stoffer-toloi test optional through a keyword-argument in that method, and give users control over that. The current PR makes a fundamental change without users being aware which should IMO be prevented.

dbrakenhoff:
I'm okay with making the behavior optional, but if we use the `asfreq()` method as an alternative, I think we need to add a warning when the resulting equidistant noise series is missing more than some percentage of the number of original noise values?

And to be clear, the current implementation in pastas will silently drop a bunch of noise values if the original timeseries timesteps are not equidistant, especially if the timestamp of the first observation is different from others in the timeseries. The method `get_equidistant_series()` attempts to prevent this loss of data by creating an equidistant series that takes as much data as possible from the original noise series. It does shift observations in time (slightly) to create this equidistant timeseries. I think the latter is preferable in situations when `asfreq()` would drop a lot of values.

PR #354, opened 2021-12-06
Fix #353, precision of floats in JSON files

### LINKED ISSUE #353: [BUG] json (pas-file) float precision
**Describe the bug**
The float precision when calling `DataFrame.to_json()` is 10 by default (max 15) and stores numbers in decimal format when it would be much more logical to store most values in scientific format. This means that after loading a model from a pas-file, the float precision is affected which can lead to unexpected results. 

**To Reproduce**
Get some model with a very small parameter value somewhere in `ml.parameters` or `ml.fit.pcov` and then store model as pas file:
```python
ml.to_file(f"{ml.name}.pas")
ml2 = ps.io.load(f"{ml.name}.pas")
# try one of these, depending on which DataFrame contains the small float value
ml.fit.pcov.equals(ml2.fit.pcov)
ml.parameters.equals(ml2.parameters)
```

**Expected behavior**
Store values with higher precision and scientific notation in pas files.

dbrakenhoff:
# Short Description
Fix for float precision in pas-files. 
Also one other minor fix:
- add squeeze to WellModel.get_stress() to return series (which is expected by `ml.plots.stresses()`).

- [x] closes issue #353
- [x] is documented
- [x] PEP8 compliant code
- [x] tests added / passed
- [ ] ~API changes documented in Release Notes~


martinvonk (commented):
Looks good. Tested it for a model and the precision in the .pas file is now larger.

dbrakenhoff:
One review is fine, merging.

PR #359, opened 2021-12-23
Save and read options of FlexModel to/from pas-file

rubencalje:
# Short Description
This fix makes it possible to save the interception, snow and gw_uptake in a pas-file

- [ ] closes issue #xxxx
- [ ] is documented
- [ ] PEP8 compliant code
- [ ] tests added / passed

### DIFF
--- a/pastas/io/base.py
+++ b/pastas/io/base.py
@@ -90,7 +90,11 @@ def _load_model(data):
         if "rfunc" in ts.keys():
             ts["rfunc"] = getattr(ps.rfunc, ts["rfunc"])
         if "recharge" in ts.keys():
-            ts["recharge"] = getattr(ps.recharge, ts["recharge"])()
+            recharge_kwargs = {}
+            if 'recharge_kwargs' in ts:
+                recharge_kwargs = ts["recharge_kwargs"]
+            ts["recharge"] = getattr(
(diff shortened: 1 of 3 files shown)


martinvonk (approved):
Happy this works!

raoulcollenteur:
I see this is already closed, so this is just for future reference. Perhaps a to_dict method added to the recharge class would be cleaner approach with less redundancy (storing the same things twice). Then the RechargeModel could call the to_dict method, similar to the other objects in Pastas. But it's all internal methods so we could change that in the future.

rubencalje:
It is true that the methods are all internal. But the way it is programmed now, an extra property is added to the representation of a stressmodel in a pas-file, called 'recharge_kwargs'. This is the only part that might be difficult to change in the future.

I could not thick of a to_dict way of implementing the storing of extra arguments while keeping backwards-compatibity with older pas-files. But maybe we can sort that out later.

raoulcollenteur:
Ah yes, I meant only for writing the kwargs when storing a file (e.g., FlexModel.to_dict()). I think adding the 'recharge_kwargs' was necessary anyway so that makes sense. Anyway, happy it works now and we can adjust the rest easily in the future.

Issue #360, opened 2021-12-23
Provide instances instead of classes for response functions 


rubencalje:
Right now we pass a rfunc class to a stressmodel. In this way, we cannot add optional arguments to a response function. For some stressmodels, like Spline or Kraijenhoff, extra optional arguments are possible.

Therefore I propose we change the rfunc argument in a StressModel from an rfunc-class to a rfunc instance. To keep existing code working, an rfunc-class will be permitted for now as well. We will make an instance of the class  in the init of the stressmodel in that case.

Also these extra arguments need to be saved to a pas-file. We can use the same procedure as is implemented for recharge-models, see pull request #359.

rubencalje:
Implemented by pull request #361

raoulcollenteur:
Reopening this issue to complete the change over Pastas 0.23 and 1.0.

- [x] Add warnings when users use classes instead of errors for rfuncs in Pastas 0.22
- [ ] Throw Errors when users use classes instead of errors for rfuncs in Pastas 0.23
- [ ] Update all example materials to use rfunc instances in Pastas 0.23
- [ ] Remove error messages for Pastas 1.0.

 Please do not start making these changes in the Dev-branch before PR #455 is merged, otherwise this would be chaos.

raoulcollenteur:
After merging  #466 the only thing left to do is to remove the warnings for Pastas 1.0. Updated the milestone accordingly.

raoulcollenteur:
Closing after merging https://github.com/pastas/pastas/pull/504

PR #361, opened 2021-12-23
Make rfunc an instance

rubencalje:
# Short Description
This pull request makes it possible to use a response function instance instead of a class, when adding the response function to a StressModel. See issue #360.

- [x] closes issue #360
- [x] is documented
- [x] PEP8 compliant code
- [x] example edited
- [x] tests passed


rubencalje:
It turns out to be a rather large change, mostly internal, but if people use rfunc-instances directly without adding them to a model they may run into trouble. The  arguments `up`, `meanstress` and `cutoff` are removed from the rfunc's `__init__`, and are now set in a method called `_add_to_model`, which is called in a stressmodel.

Special changes needed to be made to HantushWellModel, where the distances need to be set. I now chose to do this in a method called `set_distances` (which is called by the WellModel), but maybe we should keep the distances in `__init__` of HantushWellModel...

Despite these doubts, overall I think this is a good change, to make instances of rfunc-argument in stressmodels, just like we do for stressmodels and other model-elements. And it is backward-compatible, and should perform the same in all in current models.

dbrakenhoff:
I'm all for this change, I've run into situations where I wanted to add additional options to my response function before and this makes that possible. 

Also I don't think moving the parameters `up`, `meanstress` and `cutoff` from the init impact the behavior that much. The first two are only used for in `get_init_parameters` and `cutoff` is parameter that can be passed separately in most rfunc methods. I don't think people use the init parameters from an rfunc separately.

In the case of `HantushWellModel` the distances are needed to rescale the default parameter bounds for parameter `b` and that's all they're used for. The neatest way would be to set them in init to ensure they're known before calculating the parameter bounds. With this proposal you need to know to call `set_distances()` first. But I can't really think of a case where it would go wrong with this proposed implementation with `set_distances()`. Perhaps we can set the default distance value to `None`, which will cause an error if they haven't been set and they're accessed for a calculation. We can trap that error in `HantushWellModel.set_init_parameter()` and return a more informative error. 

Maybe this new behavior should be explained in the response functions notebook? And maybe the other notebooks relating to response functions should be checked as well if the information is still up to date?

Finally, I'm not sure I'm a big fan of the the method name `_add_to_model()`. But I also don't have a better idea really... Maybe `_set_rfunc_init_parameter_settings()`? But that introduces another method with parameters in its name. So I'm neutral on that one. Other thoughts?

I'll approve the PR once we get `HantushWellModel` sorted out and the notebooks have been checked.

mbakker7:
This is an interesting change. And we do indeed need to make something like
this possible.

Correct me if I am wrong, but I thought we had discussed this in the last
PASTAS meeting and we wanted to add an option to pass arguments to an
rfunc, rather than adding the option to pass an instance. Do I mis-remember
that? Maybe we should discuss this (again) before we accept the PR.

rubencalje:
Thanks for the responses. 

@mbakker7 : I thought we did not really decided on a way to go. Just that there were several ways, of which I have implemented one now: where rfunc is (or can be) an instance of a response function class. The other way is to add an extra parameter to stressmodel called rfunc_kwargs, which contains the extra parameters for the response function, and rfunc remains a class. I think this is uglier, but I can see the advantage that once the rfunc-instance is initialized, all information is known (and we do not have to call a method like _add_to_model).

@dbrakenhoff : I agree with your comments about the HantushWellModel and the naming of _add_to_model.

I am not in a hurry to add these changes, so let's discuss them at a later time, so everybody agrees.

Have a good Christmas-holiday!

rubencalje:
@dbrakenhoff and @raoulcollenteur, can I merge this pull request?

pastas/stressmodels.py
```diff
@@ -210,7 +221,7 @@ class StressModel(StressModelBase):
     ----------
     stress: pandas.Series
         pandas Series object containing the stress.
-    rfunc: rfunc class
+    rfunc: rfunc class or instance
```
dbrakenhoff, line 224: Already change description to demand an instance? Add deprecation warning to docstring here as well?

dbrakenhoff (approved):

Issue #348, opened 2021-11-29
[BUG] simulation contains NaNs in period w missing data but residuals can be calculated for this period


dbrakenhoff:
**Describe the bug**
This is a bit of an edge case that should generally not occur if you provide your timeseries settings. 

When you add a StressModel with a timeseries that starts after the start of your calibration period, and do not provide any settings (containing the filling logic for missing data), the model simulation will contain NaNs. This makes sense, because the model cannot calculate the contribution of this stress in the period with no data. Pastas also prints a warning that this is the case. The funky thing is that the results image shows that the there are residuals in that period...

See the example below. The simulation is plotted starting in 2005, but the residuals can apparently be calculated before that time.

![bug](https://user-images.githubusercontent.com/6552225/143888595-88a5bd57-acd3-419b-895b-b0f84d8d4a4d.png)

The reason this is possible is because of the interpolation of the simulation to the observation timestamps. The period before 2005 is missing from the simulation series so it uses the first simulation value to backfill the missing period. That allows it to calculate residuals for this period. 

This bit of code that does this is in `models.py -> residuals()` (line 430):

```python
sim_interpolated = np.interp(oseries_calib.index.asi8, 
                             sim.index.asi8, sim.values)
```
A maybe logical fix would be to set extrapolation to np.nan (see below), but this causes some trouble when calculating fit statistics. That issue can probably be fixed as well, but wanted to check what the desired solution would be before going down that route. 

```python
sim_interpolated = np.interp(oseries_calib.index.asi8, 
                             sim.index.asi8, sim.values,
                             left=np.nan, right=np.nan)
```

**Expected behavior**
If you can't calculate the simulation in a period with missing data (because there is no filling logic for missing data), then you shouldn't be able to calculate the residuals either. 

**Python package version**
```
Python version: 3.9.7 | packaged by conda-forge | (default, Sep 29 2021, 19:20:46) 
[GCC 9.4.0]
Numpy version: 1.21.2
Scipy version: 1.7.1
Pandas version: 1.3.3
Pastas version: 0.19.0b
Matplotlib version: 3.4.3
```

raoulcollenteur:
~~@dbrakenhoff , is this Issue not related to the changes documented in #459 and solved in PR #459 ? If so, perhaps we can close this issue?~~

[...]

PR #379, opened 2022-03-11
fix(#348) model simulation contains Nans

dbrakenhoff:
# Long Description
Proposal for fixing/improving #348 

Currently:
- When a stress is not defined for a certain period, and no filling or extending logic is provided to the stress TimeSeries object, the model simulation will contain NaNs. These are dropped by Pastas (with a logger warning) prior to returning the result. 
- When calculating the residuals, the simulation is extrapolated to calculate the residuals. In the case of the above, the first simulation value (after dropping all the NaN values) is used to calculate the residuals for all observations that lie before the first simulation value (see image I posted in #348). 

All this only really occurs when you forget to provide settings to your stress TimeSeries objects. If you do forget though, Pastas gives some indications that this is happening in the logging output, but it's not very clear. Also the resulting model statistics are incorrect. 

I think this behavior is not desirable and should just lead to an error. 

My proposal:
- Pastas should raise an error when a stress is not defined within the simulation period. Not defined here means, either has no data, or no logic to fill/extend the time series. 
- The simulation should never be extrapolated to calculate residuals, only interpolated.
- One additional change was necessary to tmin/tmax in model settings to deal with the simulation interpolation. When the first index of the oseries is e.g. 01-01-2000 12:00:00, the first simulation date is 02-01-2000, by default. This would lead to the residuals containing a NaN value because the first observation lies before the first simulation value (which causes issues for calculating metrics etc.). To avoid this I now rounded the tmin down to the nearest freq (`tmin.floor(freq)`) and the tmax up to the nearest freq (`tmax.ceil(freq)`). 

This is implemented in this PR. Curious to hear your thoughts. Am I right in assuming there is no reason to allow simulations with NaNs?

- [x] closes issue #348 
- [ ] is documented
- [x] PEP8 compliant code
- [ ] tests added / passed


pastas/model.py
```diff
@@ -611,8 +613,9 @@ def initialize(self, tmin=None, tmax=None, freq=None, warmup=None,
             self._get_time_offset(self.settings["freq"])
 
         # Set tmin and tmax
-        self.settings["tmin"] = self.get_tmin(tmin)
-        self.settings["tmax"] = self.get_tmax(tmax)
+        self.settings["tmin"] = self.get_tmin(
+            tmin).floor(self.settings["freq"])
+        self.settings["tmax"] = self.get_tmax(tmax).ceil(self.settings["freq"])
```
martinvonk, line 618: This might be confusing because it can differ from the settings defined by the user. What do you think @raoulcollenteur ?

dbrakenhoff: Good point, perhaps the floor should then be moved into the `get_tmin()` method, and only applied when `tmin=None`.

And there is still a risk that when providing a tmin or tmax, the calculation will crash because the first observation lies before the first simulation timestep... Maybe we should allow extrapolation up to a timestep(=freq) away?

dbrakenhoff:
This is annoyingly complicated. Closing this PR. Trying to think up something new....

PR #400, opened 2022-06-29
Add SolveTimer()

dbrakenhoff:
# Short Description
Add a timer for model optimization. I was running into some _extremely_ slow running models somewhere within my 400 model dataset, and found this useful in diagnosing the issue, by printing the model name in my loop and then showing some record of elapsed time and no. of iterations. 

It does require `tqdm` to be installed, so I put it into a different file that is not imported by default. Due to the optional nature of the dependency I haven't added any examples or tests.

# Usage:
```python
from pastas.timer import SolveTimer

with SolveTimer() as t:
    ml.solve(callback=t.timer)
```
This prints the following to the console:
```
Optimization progress: 73it [00:01, 67.68it/s]
```

- [ ] ~closes issue #xxxx~
- [x] is documented
- [x] PEP8 compliant code
- [ ] ~tests added / passed~
- [ ] ~API changes documented in Release Notes~

### DIFF
--- a/pastas/modelplots.py
+++ b/pastas/modelplots.py
@@ -695,6 +695,8 @@ class Plotting:
         tmin : str or pandas.Timestamp, optional
         tmax : str or pandas.Timestamp, optional
         figsize : tuple, optional
+        stacklegend : bool, optional
+            Add legend to the stacked plot.
 
         Returns
         -------
--- a/pastas/modelplots.py
+++ b/pastas/modelplots.py
@@ -682,7 +682,8 @@ class Plotting:
         return ax
 
     @model_tmin_tmax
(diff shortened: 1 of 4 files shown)


pastas/modelplots.py
```diff
@@ -682,7 +682,8 @@ def contributions_pie(self, tmin=None, tmax=None, ax=None,
         return ax
 
     @model_tmin_tmax
-    def stacked_results(self, tmin=None, tmax=None, figsize=(10, 8), **kwargs):
+    def stacked_results(self, tmin=None, tmax=None, figsize=(10, 8),
+                        stacklegend=False, **kwargs):
```
raoulcollenteur, line 686: Not yet in docstring

raoulcollenteur (changes requested):

raoulcollenteur (approved):

PR #405, opened 2022-08-09
404 development export all time series from pastas model [closes #404]

### LINKED ISSUE #404: [Development] export all time series from Pastas model
**Describe the proposed feature**
Add a method to export all the time series from a Pastas Model to a single pandas DataFrame. This could then be used to export the results to a csv or excel, using regular pandas methods. Helps to share modeling results with others (and applications).

**Code to show the expected the behavior**

`df = ml.get_all_series(tmin, tmax)   # df is a pandas DataFrame`

raoulcollenteur:
# Short Description
As described in Issue #404 

- [x] closes issue #404
- [x] is documented
- [x] PEP8 compliant code
- [x] tests added / passed

### DIFF
--- a/pastas/model.py
+++ b/pastas/model.py
@@ -366,7 +366,7 @@ class Model:
             istart += 1
         if self.transform:
             sim = self.transform.simulate(sim, p[istart:istart +
-                                                 self.transform.nparam])
+                                                        self.transform.nparam])
 
         # Respect provided tmin/tmax at this point, since warmup matters for
         # simulation but should not be returned, unless return_warmup=True.
@@ -1164,7 +1164,7 @@ class Model:
         # use warmup
         if tmin:
             tmin_warm = (Timestamp(tmin) - warmup).floor(freq) + \
-                self.settings["time_offset"]
+                        self.settings["time_offset"]
         else:
             tmin_warm = None
 
@@ -1239,7 +1239,7 @@ class Model:
         sim_org = ml.simulate(tmin=tmin, tmax=tmax)
         return sim - sim_org
 
-    def get_all_series(self, tmin=None, tmax=None, split=True):
+    def get_output_series(self, tmin=None, tmax=None, split=True):
         """Method to get all the modeled time series from the pastas Model.
 
         Parameters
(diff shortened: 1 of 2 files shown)


dbrakenhoff (commented):
`tmin` is unused?

And "all_series" doesn't cover the stresses (input) time series, but since that is stated in the docstring under Notes, perhaps that's fine. It makes sense not to include these, since they are input series, but the observations also kind of fall under that category? Also, what if the stresses are resampled by the model, would it make sense to include them then?

raoulcollenteur:
Fixed the issue with tmin!

I thought of this method as the method to obtain all the time series to reconstruct a plot like plot_decomposition. Do you have a suggestion for a better name?

To get all the stresses we should probably add a new method ml.get_stresses next to ml.get_stress. similar to get_contribution and get_contributions.

dbrakenhoff:
Perhaps `get_input_series` and `get_output_series`, and we'll just accept that the observations series is one of the output timeseries (or we add it in both methods)?

p.s. the code isn't formatted according to autopep8 yet ;)

pastas/model.py
```diff
@@ -1239,6 +1239,54 @@ def get_transform_contribution(self, tmin=None, tmax=None):
...
+            Pandas DataFrame with the time series as columns and DatetimeIndex.
+
+        Notes
+        -----
+        Export the observed, simulated time series, the noise and residuals
+        series, and the contributions from the different stressmodels.
+
+        Examples
+        --------
+        >>> df = ml.get_all_series(tmin="2000", tmax="2010")  # return DataFrame
```
martinvonk, line 1270: still has old function name

pastas/model.py
```diff
@@ -1239,6 +1239,54 @@ def get_transform_contribution(self, tmin=None, tmax=None):
...
+            String with an end date for the simulation period (E.g. '2010').
+            If none is provided, the tmax from the oseries is used.
+        split: bool, optional
+            Passed on to ml.get_contributions. Split the contribution from
+            recharge into evaporation and precipitation. See also
+            ml.get_contributions.
+
+        Returns
+        -------
+        series: pandas.DataFrame
```
martinvonk, line 1260: Variable name `series` for a type `pandas.DataFrame` confused me when reviewing.. But I guess this is not the end of the world.

martinvonk (changes requested):

pastas/model.py
raoulcollenteur: Fixed

pastas/model.py
raoulcollenteur: changed to df

raoulcollenteur:
Should be good to go after my last commit I think, if you agree, please hit "squash and merge" and remove the branch afterwards :)

martinvonk (approved):

Issue #402, opened 2022-08-02
[ENHANCEMENT] Calculate all fit metrics with pastas.stats.metrics


martinvonk:
There are some places in the code (e.g. plots.py) where fit metrics are calculated without using the `pastas.stats.metrics` library. To make sure the metrics are consistent this should be enhanced by always calculating fit metrics using `pastas.stats.metrics`.

**Additional context**
This problem arose in PR #401.
@dbrakenhoff:  _This predates the changes to the metrics, they used to be calculated in the model methods, which could not be accessed with the callback in solve. But there is no reason not to change it now, so good idea to open a new issue._

raoulcollenteur:
Closing after last PR. Thanks!

PR #410, opened 2022-08-17
calculate all fit metrics with pastas.stats.metrics and fix pytest issues

martinvonk:
# Short Description
Solves issue where fit metrics were calculated without using the internal stats library. 

Also solves some issues that came up when testing locally using pytest.

- [x] closes issue #402 
- [x] PEP8 compliant code
- [x] tests added / passed


raoulcollenteur:
Is it necessary to change the notebook and the KNMI data file? Seems unrelated this this PR. Can you remove these files from the PR?

Otherwise fine, nice change!

martinvonk:
> Is it necessary to change the notebook and the KNMI data file? Seems unrelated this this PR. Can you remove these files from the PR?
> 
> Otherwise fine, nice change!

Yes I can remove them or maybe changing the name of this PR is enough?

The two issues popped up when testing locally with `pytest` and can be quickly fixed.

Issue #304, opened 2021-06-22
[Enhancement] Check data types in load_model method


raoulcollenteur:
**Describe the bug**
The data types are not well checked when reading a pastas model from a .pas file. See for example this issue in the Pastastore:

https://github.com/pastas/pastastore/issues/39

A name is a string, but load_model with return an integer of only numbers are provided.

**Expected behavior**
Data types should be the same, basically for everything..

dbrakenhoff:
Enforce dtypes, e.g. for names (=str).
Check for illegal characters in strings.

raoulcollenteur:
Closing after PR #415

PR #415, opened 2022-08-17
304 enhancement check data types in load model method

martinvonk:
# Short Description
This pull request updates the `validate_name` function in utils. The list of illegal characters are expanded with a list from [Michigan Technological University](https://www.mtu.edu/umc/services/websites/writing/characters-avoid/). However, the OS of the user is also checked such that Linux and MacOS users are not punished for Windows' sins (more illegal characters than need be). 

When a name contains an illegal character an warning is printed when creating a (stress)model with this name. When writing to  a dict of .pas file an Exception error is raised instead.
- [x] closes issue #304 
- [x] is documented
- [x] PEP8 compliant code

### DIFF
--- a/pastas/model.py
+++ b/pastas/model.py
@@ -1828,6 +1828,7 @@ class Model:
         --------
         :mod:`pastas.io.dump`
         """
+        self.name = validate_name(self.name, raise_error=True)
 
         # Get dicts for all data sources
         data = self.to_dict(series=series)
--- a/pastas/stressmodels.py
+++ b/pastas/stressmodels.py
@@ -142,6 +142,7 @@ class StressModelBase:
         data = []
 
         for stress in self.stress:
+            stress.name = validate_name(stress.name, raise_error=True)
             data.append(stress.to_dict(series=series))
 
         return data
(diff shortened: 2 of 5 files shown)


raoulcollenteur:
I like the use and extension of validate name for this. But for the solution to the problem, can we not better validate the name(s) when creating a Model or StressModel instance? Perhaps we can apply this in the __init__ methods, and tackle the issue early?

martinvonk:
> I like the use and extension of validate name for this. But for the solution to the problem, can we not better validate the name(s) when creating a Model or StressModel instance? Perhaps we can apply this in the **init** methods, and tackle the issue early?

Do you mean raising an error instead of a warning when providing a name with illegal characters?

martinvonk:
Discussed this in person with @raoulcollenteur: 
Now the names of the pastas.TimeSeries are validated there should be no problems anymore. The problem was sometimes that integers or floats would be provided as a TimeSeries name.

If either one of you, @raoulcollenteur or @dbrakenhoff, can review then this PR?

dbrakenhoff (approved):

raoulcollenteur (approved):

Issue #407, opened 2022-08-16
[ENHANCEMENT] Remove pastas.read module entirely


dbrakenhoff:
**Describe the proposed enhancement**
Add deprecation warnings

**Additional context**
Link to hydropandas for read methods.

raoulcollenteur:
Closing Issue as per last PR #417 from @OnnoEbbens . Thanks!!

raoulcollenteur:
Re-opening this issue to document the removal of this module better.. 

- [ ] Read module gives warnings in pastas 0.22.
- [ ] Read module gives errors in pastas 0.23 (all code within the method can be removed, only raising errors code remains)
- [ ] Read module is removed in pastas 1.0 (entered folder is removed)

For future reference, @OnnoEbbens where can we find the new functions, can you post a link to Hydropandas here?

raoulcollenteur:
After merging  #470 the only thing left to do is to remove the the module entirely for Pastas 1.0. Updated the milestone accordingly.

raoulcollenteur:
Closing issue as completed after merging PR #494.

PR #417, opened 2022-08-17
407 enhancement deprecation warnings for pastasread methods

OnnoEbbens:
# Short Description
Add a deprecation warning for all functions in pastas.read. Modify the notebook of Dutch datasets to use Hydropandas.

- [x] closes issue #407
- [x] PEP8 compliant code


OnnoEbbens:
I have added an example of how you can use Hydropandas to create a Pastas model using Dutch datasets. I think pastas tests all notebooks so maybe this means we have to install hydropandas for the tests. If that is something we don't want I can move the notebook to hydropandas.

raoulcollenteur:
I think we can remove the read_dutch_data notebook, and as we don't want to add a dependency to HydroPandas, so all the example data needs to be replaced or read using pandas.read_csv. Quite a bit of work actually...

OnnoEbbens:
You are right. I did the following:
- remove the Dutch dataset notebook. I added all the information in the notebook to a hydropandas notebook -> https://github.com/ArtesiaWater/hydropandas/blob/724efed5aa6fbd0780788611a0b35ea86f3fc4ee/examples/03_hydropandas_and_pastas.ipynb
- I removed all Dino, KNMI, Waterbase and Menyanthes files and replaced them with csv files. This also involved some small changes to the notebooks. I visually checked if the results remained the same.
- remove the tests specifically testing the old read functions. I think it is not necessary to replace these tests with .csv files because the Pastas functions used in these tests are also tested elsewhere.

I still want to add the example of reading a Menyanthes file to hydropandas. This is not related to this pull request

OnnoEbbens:
I have no idea what is happening to the tests. It seems like this pull_requests is not tested again when I commit to it but it does show the failure of the last test.

requirements.ci.txt
```diff
@@ -10,4 +10,6 @@ jupyter>=1.0.0
 
 numba>=0.51
 
-lmfit>=1.0.0
\ No newline at end of file
+lmfit>=1.0.0
+
+hydropandas>=0.6.0
```
raoulcollenteur, line 15: This is not necessary anymore right now I think?

raoulcollenteur (changes requested):
Can you remove hydropandas from requirements.ci.txt ? Then it should all be good.

raoulcollenteur:
That looks like it was quite some work, but great PR! Just one minor comment before "squash and merge"!

OnnoEbbens:
Thanks! I removed the hydropandas requirement and yes the PR was more work than expected. Also it is remarkable how often we use the same time series in the example notebooks. Maybe we could try to use more different datasets in the notebooks so we automatically test more edge cases. However that's kind of a boring job to do so only for another lockdown I would say :)

raoulcollenteur (approved):

Issue #389, opened 2022-04-07
[ENHANCEMENT] Implement Type Hints into Pastas


raoulcollenteur:
**Describe the proposed enhancement**
Following Pastastore, implement Type Hinting into Pastas to improve Error feedback. @dbrakenhoff, good idea?

dbrakenhoff:
I think it's a "nice-to-have" addition for users. As far as I know, they're just hints that can generally also be picked up by users from the docstring. So not sure how error feedback would improve? But it does allow you to check the code with mypy (which I haven't really looked at much) and I think the hints can be pretty helpful.

OnnoEbbens:
I thought you could also enforce the types but now that I look into it I see that's not true.

This issue came up when someone tried to use the `ps.plots.series` function with a DataFrame instead of a Series. From the error it was hard to know what went wrong. We could manually add type checks like this :
```
if not isinstance(head, [pd.Series, ps.Timeseries]):
    raise TypeError('expected pandas Series or pastas Timeseries object')
```
But this is maybe a bit to rigid for pastas. Besides it is a lot of work to implement this for the relevant pastas functions.

martinvonk:
It would be nice to have one big file with the correct types of each arguments. This way these types can be implemented correctly in functions. Additionaly, checks can be implemented to force the correct types.

PR #441, opened 2022-12-27
Implement Type Hints

martinvonk:
# Short Description
This pull request implements Type Hints into Pastas. Python does not enforce function and variable type annotations. However they can be used by third party tools such as type checkers, IDEs, linters, etc. This helps the Pastas users and developers to write better code.

- [x] closes issue #389 
- [x] is documented
- [x] PEP8 compliant code
- [x] tests added / passed


martinvonk:
I spent a lot of time in public transport this Christmas so this is the end result 🎁. I did not implement the type annotation in the read functions since they will be deprecated in the future.

martinvonk:
The read the docs test fails because it uses a `numpy` version > 1.23 and `numba` does not support it yet.
From https://github.com/numba/numba/issues/8615 :

> For devs who are trying to fix their environments
> The problem occurs in the latest version of numpy. Installing a lower version of numpy (<1.24) fixed the problem for us.

It is being worked on:
https://github.com/numba/numba/pull/8620

For now I'll try and fix it in the requirements by forcing `numpy` < 1.24

martinvonk:
This can be merged into the dev branch after a new version of Pastas is released (at the latest next Tuesday) by @raoulcollenteur. 

Maybe @dbrakenhoff can check this PR?

raoulcollenteur:
> The read the docs test fails because it uses a `numpy` version > 1.23 and `numba` does not support it yet. From [numba/numba#8615](https://github.com/numba/numba/issues/8615) :
> 
> > For devs who are trying to fix their environments
> > The problem occurs in the latest version of numpy. Installing a lower version of numpy (<1.24) fixed the problem for us.
> 
> It is being worked on: [numba/numba#8620](https://github.com/numba/numba/pull/8620)
> 
> For now I'll try and fix it in the requirements by forcing `numpy` < 1.24

Nice catch. Should we not directly change this in the dev-version, rather than in every single PR?

martinvonk:
> Nice catch. Should we not directly change this in the dev-version, rather than in every single PR?

I don't have permission to write directly to the dev branch :), so I'd just merge this PR to the dev and everything will be solved. I'll remove the change from the #441 PR afterwards.

pastas/utils.py
```diff
@@ -446,15 +449,15 @@ def datenum_to_datetime(datenum):
         + timedelta(days=days) - timedelta(days=366)
 
 
-def datetime2matlab(tindex):
+def datetime2matlab(tindex: Type[DatetimeIndex]) -> pstAL:
```
dbrakenhoff, line 452: This can also be a datetime array theoretically

pastas/utils.py
```diff
@@ -542,11 +545,11 @@ def remove_console_handler(logger=None):
             logger.removeHandler(handler)
 
 
-def add_file_handlers(logger=None, filenames=('info.log', 'errors.log'),
-                      levels=(logging.INFO, logging.ERROR), maxBytes=10485760,
-                      backupCount=20, encoding='utf8',
-                      fmt='%(asctime)s - %(name)s - %(levelname)s - %(message)s',
-                      datefmt='%y-%m-%d %H:%M'):
+def add_file_handlers(logger: Optional[Type[Any]] = None, filenames: Optional[Tuple[str, str]] = ('info.log', 'errors.log'),
+                      levels: Optional[Tuple[Any, Any]] = (logging.INFO, logging.ERROR), maxBytes: Optional[int] = 10485760,
```
dbrakenhoff, line 549: I think this can just be Optional[Tuple[Any]], it doesn't necessarily contain only two items?

dbrakenhoff (changes requested):
Nice work! There are some things to look into though:
- Remove unnecessary Type[] calls
- Check if Series should be replaced by Series|DataFrame Union
- More explicit Type names in typeh.py, remove pst prefix where appropriate.
- Check for other solutions for type hinting while avoiding circular imports? Use strings in annotations or `from __future__ import annotations`? Use Protocol classes that define structure of e.g. the Model class without containing any code? 

Have you tried running mypy yet?

pastas/utils.py
```diff
@@ -128,7 +131,7 @@ def _get_dt(freq):
     return dt
 
 
-def _get_time_offset(t, freq):
+def _get_time_offset(t: Type[Timestamp], freq: str) -> Type[Timedelta]:
```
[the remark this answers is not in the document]
martinvonk: Unnecessary indeed, removed

pastas/recharge.py
```diff
@@ -98,13 +99,13 @@ def __init__(self):
...
 
-    def get_init_parameters(self, name="recharge"):
+    def get_init_parameters(self, name: Optional[str] = "recharge") -> Type[DataFrame]:
         parameters = DataFrame(
             columns=["initial", "pmin", "pmax", "vary", "name"])
         parameters.loc[name + "_f"] = (-1.0, -2.0, 0.0, True, name)
         return parameters
 
-    def simulate(self, prec, evap, p, **kwargs):
+    def simulate(self, prec: pstAL, evap: pstAL, p: pstAL, **kwargs) -> pstAL:
```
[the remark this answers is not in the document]
martinvonk: I now copy the NumPy type `ArrayLike` to define our own array_like (pstAL) type. Since we use a lot of NumPy arrays in Pastas and they were documented either as `np.array`, `array`, `np.ndarray`, `array_like` etc. I thought it would be best to define one type for all (and for now copy it from NumPy's definition).

pastas/utils.py
martinvonk: agreed, removed

pastas/utils.py
```diff
@@ -179,7 +182,7 @@ def get_sample(tindex, ref_tindex):
         return tindex[ind]
 
 
-def timestep_weighted_resample(series0, tindex):
+def timestep_weighted_resample(series0: Type[Series], tindex: Type[Index]) -> Type[Series]:
```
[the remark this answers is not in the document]
martinvonk: I disagree with the notion that if a method or function could work with another type, we should add it to the type annotation. Type hints are for what type we intend to be parsed.

martinvonk:
In hindsight, I should have explained my choices when opening this pull request a bit more. However, I thought this would be just an administrative task but you have to put a lot of thought and time into type hinting if you want to implement it. 

So the choices I made for this PR:
- Follow the current docstring as closely as possible. If no docstring is available, I try to infer the type. 
- Annotate the type of what is intended for the function/method by Pastas, don't annotate the type if something else could also work.
- Implementing type checking (by running MyPy) is a nice feature. However, now the type annotations are hints, not "truths". When running type checking we will find "lies" in the docstring and type hints and that'll result in a lot of extra work in fixing those. So for now just implement the type annotation as a hint and in the future, we can make Pastas more high quality by implementing type checking.
- I want the types to be defined in a single place, so not in every .py file and I want it to be clear that something is a "Pastas" type. That's why I prepend "pst" (short for "Pastas type") such that the user knows that it is a type we invented. For the case of the Axes and Figure, which are Matplotlib types, I agree that we could change it to make it more clear because these aren't Pastas' own classes. However, you could argue that these are types that we use so it can be defined as a Pastas type.
- I want to be as consistent as possible for similar types across different .py files. So for the NumPy arrays, I use the NumPy type ArrayLike and convert it to our own type pstAL. This way we remove the issue of documenting NumPy arrays as `np.array`, `array`, `np.ndarray`, `array_like` etc. I thought it would be best to define one type for all (and for now copy it from NumPy's definition).
- Since I want to define the types in a single place, I can't import Pastas' classes into the pastas.typing.types.py file due to circular imports. We could solve the circular import error by moving the abstract base classes to a separate file. And if we are changing things on the ABC's, we could think about implementing Protocols instead/also as @dbrakenhoff mentions. Protocols are supported since Python version 3.8 but I believe we intend to drop support for Python v3.7 since Python will drop security support for v3.7 this year. However, I think it is best to not change this in this pull request, and if we want to change something about the base classes and implement protocols this is something for a separate pull request.

martinvonk:
I discussed with @dbrakenhoff, I removed the "pst" prefix. If this leads to duplications in name I import the type using the `import as` statement.

martinvonk:
Can someone (I guess @raoulcollenteur) have a look at the readthedocs failure? 

Now I get:
> Exception occurred:
>   File "/home/docs/checkouts/readthedocs.org/user_builds/pastas/envs/441/lib/python3.10/site-packages/sphinx/util/nodes.py", line 271, in get_node_source
>     raise ValueError("node source not found")
> ValueError: node source not found

One version before I got: 

> Theme error:
> An error happened in rendering the page api/generated/generated/generated/pastas.noisemodels.ArmaModel.__init__.
> Reason: UndefinedError("'logo' is undefined")

All seems a bit random since it should not depend on my changes.

pastas/io/base.py
```diff
@@ -6,11 +6,13 @@
...
 import pastas as ps
 from pandas import to_numeric
+# Type Hinting
+from pastas.typing import pstMl
 
 logger = getLogger(__name__)
 
 
-def load(fname, **kwargs):
+def load(fname: str, **kwargs) -> pstMl:
```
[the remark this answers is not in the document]
martinvonk: See comments below. I like the idea of implementing a Protocol but that's something for another PR/day :)

pastas/typing/typeh.py
```diff
@@ -0,0 +1,43 @@
...
+from pandas import Timestamp  # Series, DataFrame
+# Matplotlib
+from matplotlib.axes._base import _AxesBase
+from matplotlib.figure import FigureBase
+# Numpy
+from numpy.typing import ArrayLike
+# Typing
+from typing import Type, Union, Optional, Tuple, List, Any, TypeVar
+
+pstAx = TypeVar("pstAx", bound=_AxesBase)  # Matplotlib Axes
```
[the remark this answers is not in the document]
martinvonk: Fixed in latest commits

martinvonk:
How about formatting this pull request with [Black](https://black.readthedocs.io/en/stable/)? I feel like methods and functions with type hinting become much more readable with Black.

For instance this (with Black formatting) is much more clear:
```
    def get_stress(
        self,
        p: Optional[Array_Like] = None,
        tmin: Optional[Tminmax] = None,
        tmax: Optional[Tminmax] = None,
        freq: Optional[str] = None,
        istress: Optional[int] = None,
        squeeze: bool = True,
        **kwargs,
    ) -> DataFrame:
```

than this (without Black formatting):

```
def get_stress(self, p: Optional[Array_Like] = None, tmin: Optional[Tminmax] = None, tmax: Optional[Tminmax] = None, freq: Optional[str] = None,
                   istress: Optional[int] = None, **kwargs) -> DataFrame:
```

And there aren't any significant pull request at the moment to create nasty merge conflicts with.
Thoughts @raoulcollenteur, @dbrakenhoff, @rubencalje?

doc/examples/example_uncertainty.py
```diff
@@ -27,14 +27,14 @@
 print('percentage inside:', np.count_nonzero(inside) / len(inside) * 100)
 
 # # Plot some results
-axes = ml.plots.results(tmin="2010", tmax="2015", figsize=(10, 6))
-axes[0].fill_between(df.index, df.iloc[:, 0], df.iloc[:, 1], color="gray",
-                     zorder=-1, alpha=0.5, label="95% Prediction interval")
-axes[0].legend(ncol=3)
+axs = ml.plots.results(tmin="2010", tmax="2015", figsize=(10, 6))
```
dbrakenhoff, line 30: what did the 'e' do wrong ... :P?

pastas/model.py
```diff
@@ -1,25 +1,36 @@
...
+from pastas.stressmodels import Constant
+from pastas.timeseries import TimeSeries
+from pastas.transform import ThresholdTransform
+from pastas.utils import (_get_dt, _get_time_offset, frequency_is_supported,
+                          get_sample, validate_name)
+from pastas.version import __version__
+
+# Type Hinting
+from typing import Union, Optional, Tuple, List
+from pastas.typing import StressModel, Solver, Tminmax, Array_Like
```
dbrakenhoff, line 30: Should we just label Tmintmax as a Timestamp?

pastas/typing/__init__.py
```diff
@@ -0,0 +1 @@
+from .types import Axes, Figure, Tminmax, Model, TimeSeries, StressModel, NoiseModel, Solver, Recharge, Reservoir, CallBack, Function, RFunc, Array_Like
```
dbrakenhoff, line 1: this line seems long?

pastas/stats/tests.py
```diff
@@ -411,8 +415,8 @@ def stoffer_toloi(series, lags=15, nparam=0, freq="D",
     return qm, pval
 
 
-def diagnostics(series, alpha=0.05, nparam=0, lags=15, stats=(),
-                float_fmt="{0:.2f}"):
+def diagnostics(series: Series, alpha: float = 0.05, nparam: int = 0, lags: int = 15, stats: tuple = (),
```
dbrakenhoff, line 418: Tuple with capital T?

pastas/stats/tests.py
```diff
@@ -7,17 +7,21 @@
...
+from typing import Tuple
+
+
 logger = getLogger(__name__)
 __all__ = ["durbin_watson", "ljung_box", "runs_test", "stoffer_toloi",
            "diagnostics", "plot_acf", "plot_diagnostics"]
 
 
-def durbin_watson(series=None):
+def durbin_watson(series: Series = None) -> float:
```
dbrakenhoff, line 24: Optional[Series]?

dbrakenhoff: Wait, why is series optional to begin with?

pastas/stats/tests.py
```diff
@@ -86,7 +90,7 @@ def durbin_watson(series=None):
     return dw_stat, p
 
 
-def ljung_box(series=None, lags=15, nparam=0, full_output=False):
+def ljung_box(series: Series = None, lags: int = 15, nparam: int = 0, full_output: bool = False) -> Tuple[float, float]:
```
dbrakenhoff, line 93: Why series=None?

pastas/stressmodels.py
```diff
@@ -47,8 +51,7 @@ class StressModelBase:
     """
     _name = "StressModelBase"
 
-    def __init__(self, name, tmin, tmax, rfunc=None, up=True, meanstress=1,
-                 cutoff=0.999):
+    def __init__(self, name: str, tmin: Tminmax, tmax: Tminmax, rfunc: Optional[RFunc] = None, up: bool = True, meanstress: float = 1.0, cutoff: float = 0.999) -> None:
```
dbrakenhoff, line 54: line too long

pastas/stats/metrics.py
```diff
@@ -417,7 +423,7 @@ def bic(obs=None, sim=None, res=None, missing="drop", nparam=1):
             nparam * log(res.index.size))
 
 
-def aic(obs=None, sim=None, res=None, missing="drop", nparam=1):
+def aic(obs: Series = None, sim: Series = None, res: Series = None, missing: str = "drop", nparam: int = 1) -> float:
```
dbrakenhoff, line 426: should this be Optional[Series]?

doc/examples/example_uncertainty.py
martinvonk: Nothing, this is me trying to clean up the names since the type name is now `Axes` but we in a lot of places in the code we assign the variable name `axes` to a list or dict of `Axes`. However, I gave up renaming because we use the name `axes` too much, also as an attribute name e.g. in the CompareModels class.

pastas/typing/types.py
```diff
@@ -0,0 +1,34 @@
...
+Axes = TypeVar("Axes", bound=_AxesBase)  # Matplotlib Axes
+Figure = TypeVar("Figure", bound=FigureBase)  # Matplotlib Figure
+Array_Like = TypeVar("Array_Like", bound=ArrayLike)  # Array Like (NumPy based)
+
+# Internal library
+if TYPE_CHECKING: # https://mypy.readthedocs.io/en/latest/runtime_troubles.html
+    import pastas as ps
+
+# Internal Types
+Tminmax = TypeVar("Tminmax", bound=Union[str, Timestamp])  # Tmin or Tmax
```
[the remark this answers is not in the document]
martinvonk: I don't think that makes it clear that users can also parse a string instead of a pd.Timestamp. Also that will give issues when importing.

pastas/plots.py
```diff
@@ -7,19 +7,25 @@
...
+
+# Type Hinting
+from typing import Optional, List, Tuple
+from pastas.typing import Array_Like, Axes, Figure, Tminmax, Model
+
 
 logger = logging.getLogger(__name__)
 
-__all__ = ["compare", "series", "acf", "diagnostics", "cum_frequency",
+__all__ = ["compare", "Series", "acf", "diagnostics", "cum_frequency",
```
dbrakenhoff, line 24: Series shouldn't be capitalized

pastas/modelstats.py
```diff
@@ -321,8 +325,8 @@ def summary(self, tmin=None, tmax=None, stats=None):
         return stats
 
     @model_tmin_tmax
-    def diagnostics(self, tmin=None, tmax=None, alpha=0.05, stats=(),
-                    float_fmt="{0:.2f}"):
+    def diagnostics(self, tmin: Optional[Tminmax] = None, tmax: Optional[Tminmax] = None, alpha: float = 0.05, stats: tuple = (),
```
dbrakenhoff, line 328: Small tuple instead of big Tuple?

pastas/model.py
```diff
@@ -1314,7 +1325,7 @@ def _get_response(self, block_or_step, name, p=None, dt=None, add_0=False,
             timestep for the response function.
         add_0: bool, optional
             Add a zero at t=0.
-        kwargs
+        kwargs: dict
```
dbrakenhoff, line 1328: mention to which method kwargs are passed

dbrakenhoff (changes requested):
Looking much better already. Nice move with the quoted annotations in typing.py and the mypy trick.

I think we should replace Tminmax with a normal Timestamp (or Union[str, Timestamp]), because those are functionally the same types. 

Also nitpicking, but perhaps rename Array_Like to ArrayLike (more in line with Python class naming conventions)?

I agree on the black formatting style for functions with lots of input parameters. Whether we should apply it to the whole code base is a separate discussion maybe.

pastas/stats/tests.py
martinvonk: Not needed here because the type is `tuple`. `Tuple` with capital T is only needed for Python version 3.8 and lower if a tuple would be called like: `tuple[...]`. Same goes for `list` and `List`.

pastas/plots.py
martinvonk: Agreed, resolved.

pastas/modelstats.py
martinvonk: See comment above for test.py

pastas/model.py
martinvonk: done.

pastas/stats/metrics.py
martinvonk: Yes, resolved.

pastas/model.py
martinvonk: See comment below on types.py

pastas/typing/__init__.py
martinvonk: resolved

pastas/stats/tests.py
martinvonk: Agreed that None should not be optional to begin with.

pastas/stressmodels.py
martinvonk: formatted all functions and methods :)))

pastas/stats/tests.py
martinvonk: See comment above for tests.durbin_watson()

martinvonk:
> How about formatting this pull request with [Black](https://black.readthedocs.io/en/stable/)? I feel like methods and functions with type hinting become much more readable with Black.

I was crazy enough to format the functions and methods by hand. Leaving the rest of the code (and thus git blame) intact which is one the arguments @dbrakenhoff raises in #448

raoulcollenteur (approved):

raoulcollenteur:
Don't know how to rerun all GH action jobs without 3.11.

Issue #442, opened 2022-12-27
[Development] Add pyproject.toml


martinvonk:
We need to add an pyproject.toml

Useful reference: https://ianhopkinson.org.uk/2022/02/understanding-setup-py-setup-cfg-and-pyproject-toml-in-python/

Warning I encountered:
DEPRECATION: ... is being installed using the legacy 'setup.py install' method, because it does not have a 'pyproject.toml' and the 'wheel' package is not installed. pip 23.1 will enforce this behaviour change. A possible replacement is to enable the '--use-pep517' option. Discussion can be found at https://github.com/pypa/pip/issues/8559

I believe @dbrakenhoff already encountered some issues with this in [@pastastore](https://github.com/pastas/pastastore/commit/08d9f3017fb313ff5989604586f5e43130b1f1e2)

raoulcollenteur:
OK, not a fan of moving everything into a src/pastas folder but if that is what Pypa wants I think this should be done.

Let make a 0.22.0 release first, and right after that merge the TypeHinting PR and move the whole thing into src folder.

I kinda feel like we should release a 0.23.0b to master to check if the GH workflow for packaging still works...

martinvonk:
Closing this since #443 is merged.

martinvonk:
I foresee a problem here when creating the next release :)

https://github.com/pastas/pastas/blob/17c6fd4960d5adf7fa0bf645e03a98691c0dcb6f/.github/workflows/python-publish.yml#L30

raoulcollenteur:
Yes, can you fix this? Also, do we still really need the version.py with the new format? And if so, we could move `show_version`, `check_numba` and `check_numbascipy` to version.py?

PR #443, opened 2022-12-28
Transition to pyproject.toml

martinvonk:
# Short Description
See #442 

- [x] closes issue #442 
- [x] is documented
- [x] PEP8 compliant code

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,46 @@
+[project]
+name = "pastas"
+version = "0.22.0b"
+description = "Python package to perform time series analysis of hydrological time series."
+license = { file="LICENSE" }
+readme = "README.rst"
+authors = [ { name = "Collenteur et al. 2019" } ]
+maintainers = [
+{name="R.A. Collenteur", email="raoulcollenteur@gmail.com"},
+{name="M. Bakker", email="markbak@gmail.com"},
+{name="R. Calje", email="r.calje@artesia-water.nl"},
+{name="F. Schaars", email="f.schaars@artesia-water.nl"},
+{name="D.A. Brakenhoff", email="d.brakenhoff@artesia-water.nl"},
+{name="O. Ebbens", email="o.ebbens@artesia-water.nl"},
+{name="M.A. Vonk", email="m.vonk@artesia-water.nl"}
+]
+
+requires-python = ">=3.7"
+dependencies = [
+"numpy>=1.17",
+"matplotlib>=3.1",
+"pandas>=1.1",
+"scipy>=1.3"
+]
+
+keywords = ["hydrology", "groundwater", "timeseries", "analysis"]
+classifiers = [
+    'Development Status :: 4 - Beta',
+    'Intended Audience :: Science/Research',
+    'Intended Audience :: Other Audience',
+    'License :: OSI Approved :: MIT License',
+    'Programming Language :: Python :: 3 :: Only',
+    'Programming Language :: Python :: 3.7',
+    'Programming Language :: Python :: 3.8',
+    'Programming Language :: Python :: 3.9',
+    'Programming Language :: Python :: 3.10',
+    'Topic :: Scientific/Engineering :: Hydrology',
+]
+
+[project.urls]
+homepage = "https://pastas.dev"
+repository = "https://github.com/pastas/pastas"
+documentation = "https://pastas.readthedocs.io/en/latest/"
+
+[project.optional-dependencies]
+full = ["numba >= 0.51", "lmfit >= 1.0.0"]
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,10 @@
+[metadata]
+name = pastas
+
+[options]
+package_dir=
+    =src
+packages=find:
+
+[options.packages.find]
+where=src
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,0 @@
-from setuptools import setup
-
-if __name__ == "__main__":
-    setup()
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ with open("README.rst", "r") as fh:
 
 # Get the version.
 version = {}
-with open("pastas/version.py") as fp:
+with open("src/pastas/version.py") as fp:
     exec(fp.read(), version)
 
 setup(
--- a/setup.py
+++ b/setup.py
@@ -1,45 +1,4 @@
-from setuptools import find_packages, setup
+from setuptools import setup
 
-with open("README.rst", "r") as fh:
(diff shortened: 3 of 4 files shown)


martinvonk:
Code still has to be moved to .src folder.

martinvonk:
If there is a new Pastas version released and all pull requests are merged in the dev branch we can update this branch. Then we can move the pastas scripts to the src/ folder and finish this PR.

dbrakenhoff:
Is it really necessary to move the code to the `src/` folder? This post suggests both (our current structure and the src-structure) are valid: https://godatadriven.com/blog/a-practical-guide-to-setuptools-and-pyproject-toml/?

setup.cfg
```diff
@@ -0,0 +1,10 @@
+[metadata]
+name = pastas
+
+[options]
+package_dir=
+    =src
```
dbrakenhoff, line 6: do not use the src structure?

pyproject.toml
```diff
@@ -0,0 +1,46 @@
...
+    'Topic :: Scientific/Engineering :: Hydrology',
+]
+
+[project.urls]
+homepage = "https://pastas.dev"
+repository = "https://github.com/pastas/pastas"
+documentation = "https://pastas.readthedocs.io/en/latest/"
+
+[project.optional-dependencies]
+full = ["numba >= 0.51", "lmfit >= 1.0.0"]
```
dbrakenhoff, line 46: add custom installation options here (see flopy for example):
- docs (sphinx stuff)
- testing (pytest, pytest-cov, etc.)
- develop (everything?)

That way we can move all requirements for specific tasks to this file and get rid of all the .txt files.

dbrakenhoff (changes requested):
See comments.

martinvonk:
> Hi @martinvonk, now would be a good time to work on this PR ;-) I agree with @dbrakenhoff that if we can stick to the current folder structure and not use a src folder I would prefer that.

Will continue working on this after #441 is merged. 

On the `/src` layout this is worth a read: https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/

setup.cfg
martinvonk: agreed

pyproject.toml
martinvonk: I added the packages to the pyproject.toml. However, I have not yet updated the installation of these packages for the read the docs from pyproject.toml.

martinvonk:
At this point the method works. I'm now try and replace the requirements.txt files.

dbrakenhoff (approved):

dbrakenhoff:
I was hoping to merge dev into this PR without all the changes showing up in this PR, but for some reason I didn't do that correctly...? Anyway, hopefully black linting passes now. 

Last thing is that we should maybe document somewhere what the procedure now looks like to create a new Release.
- where to update the pastas version
- ... (any other steps?)

martinvonk:
Can we change this line as well?
https://github.com/pastas/pastas/blob/17c6fd4960d5adf7fa0bf645e03a98691c0dcb6f/.github/workflows/ci.yml#L32

pyproject.toml
```diff
@@ -0,0 +1,77 @@
...
+    "nbsphinx",
+    "Ipython",
+    "ipykernel",
+    "pydata-sphinx-theme",
+    "sphinx-gallery",
+    "sphinx>=3.1, <6.0",
+    "sphinxcontrib-bibtex",
+    "requests",
+]
+develop = ["pastas[full]", "pastas[ci]", "numba-scipy >= 0.3.1"]
```
martinvonk, line 71: Shouldn't we do something here with numby-scipy support of scipy<=1.7.3 while scipy is already at v1.10.0?

martinvonk (commented):
I think this is change going to work :)

So to list the changes:
- setup.py is removed in favor of pyproject.toml
- update the pastas version is now only done in the pyproject.toml file
- the version.py file now uses importlib to determine the used pastas version
- setup.cfg is added but empty (which makes maintenance easy). However we need an empty setup.cfg file, otherwise we can't do `pip install -e .` 
- remove requirements.txt files and get dependencies for readthedocs and ci from pyproject.toml
- add black formatting checks

pyproject.toml
dbrakenhoff: Good point, but I'm guessing numba-scipy will cause scipy to be downgraded in this case?

Numba-scipy is also very specific to modfelling multiple wells far so maybe we should just remove mention of it altogether, or move it back into its own little category, like you had before.

dbrakenhoff: I'll move it back into its own special group.

pyproject.toml
```diff
@@ -0,0 +1,77 @@
...
+    { name = "O. Ebbens", email = "o.ebbens@artesia-water.nl" },
+    { name = "M.A. Vonk", email = "m.vonk@artesia-water.nl" },
+]
+
+requires-python = ">= 3.7"
+dependencies = [
+    "numpy >= 1.17, < 1.24",
+    "matplotlib >= 3.1",
+    "pandas >= 1.1",
+    "scipy >= 1.3",
```
raoulcollenteur, line 28: Please change to 1.8

raoulcollenteur (changes requested):
Couple of small changes as per the review comments. Particularly the Scipy version change. Then it should be good to go!

.github/workflows/ci.yml
```diff
@@ -29,20 +29,14 @@ jobs:
...
-          pip install codacy-coverage
-          pip install -e .
-      
-      - name: Lint with flake8
-        run: |
-          # stop the build if there are Python syntax errors or undefined names
-          flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
-          # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
-          flake8 . --count --exit-zero --max-complexity=10 --max-line-length=80 --statistics
+          pip install -e .[ci]
```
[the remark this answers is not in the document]
dbrakenhoff: It should be ... :crossed_fingers:

pyproject.toml
dbrakenhoff: Yes, will do.

pastas/io/base.py
```diff
@@ -3,11 +3,12 @@
 from importlib import import_module
```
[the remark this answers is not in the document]
dbrakenhoff: In theory this file hasn't been touched in this PR. I merged in the black formatting that was already done on dev, but somehow the changes are showing up here.

I might have messed up, but I was hoping to just try merging and hope it works out ok... ?

raoulcollenteur (approved):

Issue #448, opened 2023-01-09
[ENHANCEMENT] black formatting


dbrakenhoff:
As proposed by @martinvonk in #389, use black formatting, especially to make the type hinting prettier in function definitions.

One argument I had against doing this for the entire code base was that the git blame would be affected, but turns out they've thought about this: https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html

So +1 from me for black formatting for pastas!

raoulcollenteur:
+1 for black and automating this through GH actions if possible.

raoulcollenteur:
Before this issue can be closed:

- [ ]   The line-wrap seems to be 88 now, instead of the old 79, that and the fact that we use black formatting should be written in the Docs. https://pastas.readthedocs.io/en/latest/developers/index.html
- [ ]   If we apply black, I think it would be good to automate that with GH actions: https://black.readthedocs.io/en/stable/integrations/github_actions.html

dbrakenhoff:
I can add linting with Black step (replacing the old flake8 step) which checks whether code is black formatted prior to running tests.

We can automate the black formatting and commit it directly to a PR (https://peterevans.dev/posts/github-actions-how-to-automate-code-formatting-in-pull-requests/) but this does not work on forks, so not sure how useful that is. Running black is very easy, so I'm ok with keeping that manual step... but let me know what you think.

martinvonk:
https://github.com/pastas/pastas/blob/59e85aba98ef989d913b000c075fd0e544c2a100/.github/pull_request_template.md?plain=1#L7
Maybe we can change this to 
- [ ] formatt code with black

and close this issue?

raoulcollenteur:
I agree with running black formatting locally now. PR's will fail anyway when it is not black formatted. 

@martinvonk I changed the template, so following your suggestions I am closing this Issue :)

PR #457, opened 2023-01-10
448 black formatting

dbrakenhoff:
# Short Description
- run black formatting on all python files and notebooks
- run isort on all python files
- run docformatter on all python files

- [x] closes issue #448
- [ ] ~is documented~
- [x] PEP8 compliant code
- [ ] ~tests added / passed~


doc/examples/example.py
```diff
@@ -10,21 +10,21 @@
 ps.set_log_level("ERROR")
 
 # read observations and create the time series model
-obs = pd.read_csv("data/head_nb1.csv", index_col=0,
-                  parse_dates=True).squeeze("columns")
+obs = pd.read_csv("data/head_nb1.csv", index_col=0, parse_dates=True).squeeze("columns")
```
raoulcollenteur, line 13: So the line-wrap is not set to 79 anymore in black?

raoulcollenteur (approved):

raoulcollenteur:
As I understand this PR this introduces black formatting to Pastas, without affecting the git blame. Am I right?

If so, I propose to merge this PR ASAP, such that we can work on other things, and then in a follow-up PR do the following:

- [ ] The line-wrap seems to be 88 now, instead of the old 79, that and the fact that we use black formatting should be written in the Docs. https://pastas.readthedocs.io/en/latest/developers/index.html
- [ ] If we apply black, I think it would be good to automate that with GH actions: https://black.readthedocs.io/en/stable/integrations/github_actions.html

Issue #455, opened 2023-01-10
[ENHANCEMENT] Only allow Pandas.Series as user-provided input to stress models


raoulcollenteur:
# Proposed enhancement

After discussions in the last year, we agreed to change the only allowed time series input of the Stress models and for the oseries to be pandas.Series with a pandas.DatetimeIndex. This means the pastas.TimeSeries object will not be allowed as input anymore and will only be used internally. The settings can exclusively be provided through the Stress models objects (see also #408), creating one place to set this. The purpose of these changes is to make it easier for users to understand what is happening and make the modelling process more transparent. The complexity of the code will also substantially decrease, making it easier to maintain the code in the future.

## To-do-list

- [x] Change all docstrings to document that only pandas.Series with pandas.DatetimeIndex is allowed.
- [x] Create a function called internally but also available to users to validate the time series.
- [x] Update the TimeSeries object to include all the changes and remove old functionality (see list below).
- [x] Raise errors when users use old TimeSeries functionalities for Pastas 0.23 (`check_input `function).
- [x] Update the load functions for pas-files to deal with the new TimeSeries format.
- [x] Update the load function to somehow deal with the old TimeSeries format (works, but it is a hack).
- [x] Add `get_settings` method to stressmodels, also through ml.get_stressmodel_settings.
- [x] Document the use of these and existing/new pre-processing functions in an example notebook 

## List of changes made to TimeSeries object:

- `freq_original` keyword is removed, because equidistant time series are now required. The `freq_original` attribute is maintained, but directly inferred using `pd.infer_freq`.
- `validated_series` attribute is removed, as method will error when input time series do not pass all checks and thus there is no need for this attribute anymore.
- method `multiply` is deprecated, as users should not access the whole TimeSeries object anymore. Instead, multiplication can be done as a pre-processing step.
- internal method `_normalize` is removed. as this can (and probably already is) done when pre-processing the time series. In this case the user is in control of how the time series are normalized (e.g., on what period). The process was not transparent in the old implementation. 
- The following settings arguments are removed:
  - `"fill_nan"` is removed from settings, as nan-values are not allowed anymore and all stress time series are equidistant.
  - `norm` is removed from settings, as _normalize is also removed.

## Expected behavior

ts = pd.read_csv("prec.csv", parse_dates=True).squeeze()
sm = ps.StressModel(ts, name="press", settings=settings_dict)

**List of related Issues:**

- #397
- #408 
- #68

## Disclaimer

[...]

PR #459, opened 2023-01-11
455 only allow pandas series as user input

raoulcollenteur:
# Short Description

Updating the TimeSeries object. This PR is still work in progress.

- [ ] closes issue #455 
- [ ] is documented
- [ ] PEP8 compliant code
- [ ] tests added / passed


mbakker7:
What is the reason to only allow Pandas Series? Can users still supply a Pandas DataFrame with one column or is that not allowed anymore? Just wondering.

raoulcollenteur:
> What is the reason to only allow Pandas Series? Can users still supply a Pandas DataFrame with one column or is that not allowed anymore? Just wondering.

This would not be allowed anymore and result in an error. This will keep it simple: only pandas.Series allowed. Hope that is fine with everyone.

mbakker7:
I am not sure what the benefit is. Can we not accept a DataFrame and then convert it to a Series? It is easy to catch, right? Why require the user to use `squeeze()` or a similar function?

rubencalje:
I agree that a 1-column DataFrame should still be possible, as it does not create a lot of code we need to maintain. Also, I am not sure about removing the fillnan-functionality. Can't we just keep it, and raise a warning when filling of nans take place (instead of just writing info to the log)?

Do not forget to remove the output of the notebooks before merging.

raoulcollenteur:
Just changed the code to allow the 1D DataFrames again. For the fill_nan, I am also fine with keeping it as long as equidistant stresses must be provided. Then Pastas could fill the nans without making any assumptions. In that case for only for oseries "drop" (=.dropna()) would be allowed, and for stresses the other options that main equidistance. What do others think about this?

Forgot that we need to clear the notebooks from output. Will do so and also add this to the to-do-list for new PRs.

pastas/io/base.py
```diff
@@ -184,6 +127,128 @@ def _load_model(data: dict) -> Model:
...
+    if metadata:
+        ts["metadata"] = metadata if len(metadata) > 1 else metadata[0]
+    if settings:
+        ts["settings"] = settings if len(settings) > 1 else settings[0]
+
+    sm = stressmodel(**ts)
+    return sm
+
+
+def _unpack_series(**kwargs):
```
dbrakenhoff, line 211: why not just accept dictionary? instead of unpacking and repacking using ** and  **kwargs?

pastas/__init__.py
```diff
@@ -36,7 +36,7 @@
     TarsoModel,
     WellModel,
 )
-from .timeseries import TimeSeries
+from .timeseries import TimeSeries, validate_stress, validate_oseries
```
dbrakenhoff, line 39: Do we want to expose TimeSeries directly?

pastas/timeseries.py
```diff
@@ -763,3 +499,186 @@ def plot(self, original: bool = False, **kwargs) -> Axes:
...
+    Notes
+    -----
+    If any of these checks are not passed the method will throw an error that needs
+    to be fixed by the user.
+
+    """
+    # Because we are friendly and allow 1D DataFrames
+    if isinstance(series, pd.DataFrame):
+        if len(series.columns) == 1:
+            series = series.iloc[:, 0]
```
dbrakenhoff, line 619: Do add a message here when we convert?

dbrakenhoff: nvm, I think it's fine like this

pastas/__init__.py
raoulcollenteur: Good point, this should not be necessary anymore. I will remove this import.

pastas/io/base.py
raoulcollenteur: I will change this to a dict. I was initially thinking to change the export for stressmodels to include the settings and metadata there that is way I used **kwargs. But now I refrained from that so it can be a dict.

raoulcollenteur:
Added support for the fill_nan option again. PR is ready for review I think.

rubencalje:
> Added support for the fill_nan option again. PR is ready for review I think.

Thanks! Looks good, but I cannot judge if this will lead to problems with other Pastastore or other scripts.

In some notebooks 'dropna' is added. Can those changes be reverted again?

doc/examples/06_adding_trends.ipynb
```diff
@@ -40,7 +40,9 @@
     "# Load the input data\n",
     "evap = pd.read_csv(\"data/evap_260.csv\", index_col=0, parse_dates=[0]).squeeze()\n",
     "rain = pd.read_csv(\"data/rain_260.csv\", index_col=0, parse_dates=[0]).squeeze()\n",
-    "head = pd.read_csv(\"data/B32C0609001.csv\", index_col=0).squeeze()\n",
+    "head = (\n",
+    "    pd.read_csv(\"data/B32C0609001.csv\", index_col=0, parse_dates=[0]).squeeze().dropna()\n",
```
dbrakenhoff, line 44: Remove dropna()?

doc/examples/11_sgi_example.ipynb
```diff
@@ -244,7 +246,32 @@
...
+   "outputs": [],
+   "source": [
+    "well = well.asfreq(\"D\").bfill()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Now we are ready to built the model."
```
dbrakenhoff, line 274: typo

pastas/timeseries.py
```diff
@@ -740,26 +523,192 @@ def to_dict(self, series: Optional[bool] = True) -> dict:
...
+    if not series.index.is_unique:
+        msg = (
+            f"duplicate time-indexes were found in the time series {name}. Make sure "
+            f"there are no duplicate indices. For example by "
+            f"`grouped = series.groupby(level=0); series = grouped.mean()`"
+        )
+        logger.error(msg)
+        raise ValueError(msg)
+
+    # 6. Make sure the time series has no nan-values
```
dbrakenhoff, line 698: For oseries NaNs are dropped automatically, do we add a flag checking for NaNs in oseries? Or do we prefer people dropping nans from the oseries prior to passing it to Model?

tests/test_comparison.py
```diff
@@ -6,12 +6,12 @@
 evap = read_csv("tests/data/evap.csv", index_col=0, parse_dates=True).squeeze("columns")
 obs = read_csv("tests/data/obs.csv", index_col=0, parse_dates=True).squeeze("columns")
 
-ml1 = ps.Model(obs, name="Test_Model")
+ml1 = ps.Model(obs.dropna(), name="Test_Model")
```
dbrakenhoff, line 9: remove dropnas?

dbrakenhoff:
Also, this will certainly be somewhat bothersome for pastastore, but we'll start figuring that out after this is merged.

doc/examples/06_adding_trends.ipynb
raoulcollenteur: done.

pastas/timeseries.py
raoulcollenteur: Different philosophies exist here I think. I am in favour of requiring dropping NaNs by users, but it an easy operation like 1D DataFrames so I included it :)

tests/test_comparison.py
raoulcollenteur: I will keep it like this, as the purpose of the test is not to test if Model works for oseries with nan-values.

raoulcollenteur:
@dbrakenhoff can you do the honours of squash and merge on this one?

Cheers,
Raoul

dbrakenhoff:
Happily!

PS. you didn't like the try/except additions to show the validate_stress() error messages without erroring in notebook 00_prepare_timeseries?

Issue #421, opened 2022-08-22
Features that will be changed/deprecated for version 1.0 [Development]


mbakker7:
1. All stressmodels will only accept equidistant Pandas time series only (either a series or dataframe with one column). 
2. Stressmodels will only accept instance of response function (#360)
3. solve method will only accept instance of solver class (#479)
4. (Dutch) weather data and head data can not be read from inside Pastas. Functionality is available in HydroPandas.
5. StressModel2 will be removed.
6. argument `meanstress` to `gain_scale_factor` (#419)

To assist with #3, there will be helper functions to validate series
StressModel will still have functionality to fill NaN values and to extend series backward (warmup) or forward depending on the settings. The settings will be clearly documented.
Current TimeSeries metadata will be stored elsewhere.

raoulcollenteur:
Check-list for things to remove in Pastas 1.0 (please change if you find something)

- [ ] warn bounds in ml.fit_report

PR #466, opened 2023-01-16
Require user-provided response function to be instances 

raoulcollenteur:
# Short Description

Continues the implementation of only allowing rfuncs to be provided as Instances (ps.One()) instead of classes, as requested and started per #360. If users provide a rfunc class instead of an instance, an Error will be thrown asking the user to provide an instance. 
 
EDIT: still work I progress, but should be done 17.01.2023 9:00, ready for review.

- [x] part of issue #360
- [x] is documented
- [x] PEP8 compliant code
- [x] tests added / passed


raoulcollenteur:
I think this PR should be good to go. Change also to be documenten as per #421.

dbrakenhoff (approved):

martinvonk (approved):
I searched for any leftover rfunc classes but didn't find any so seems good to me.

pastas/stressmodels.py
```diff
@@ -694,7 +693,7 @@ def __init__(
             isinstance(rfunc, HantushWellModel) or issubclass(rfunc, HantushWellModel)
```
martinvonk, line 693: Same comment here as for the issubclass check on line 1424 @davidbrakenhoff mentioned. I believe only the isinstance check is enough. `issubclass(class, classinfo)` returns True if a class is a subclass of classinfo. The subclass here would be RfuncBase for classinfo HantushWellModel (and Exponential). And we don't need to check for RfuncBase right?

raoulcollenteur: Agreed, removed.

pastas/stressmodels.py
```diff
@@ -1422,7 +1421,7 @@ def __init__(
         if rfunc is None:
             rfunc = Exponential()
         if not (isinstance(rfunc, Exponential) or issubclass(rfunc, Exponential)):
```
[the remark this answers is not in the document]
raoulcollenteur: Agreed, removed.

raoulcollenteur:
Thanks for the quick review, made the suggested changes and I will merge as soon as GH actions finishes (positively). @rubencalje no need for the review anymore :) 

Cheers,
Raoul

Issue #452, opened 2023-01-10
[ENHANCEMENT] Add Python 3.11 support


raoulcollenteur:
**Describe the proposed enhancement**

It would be great if Pastas 1.0 would support Python 3.11. As noted by @martinvonk Numba is not ready yet, but should be soon.

dbrakenhoff:
Perhaps set 3.11 to allow failure for now?
- continue on error: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

raoulcollenteur:
I'll just check up on Numba every now and then, this Issue: https://github.com/numba/numba/issues/8304

martinvonk:
We'll remove Numba from the optional dependencies and add it as a full on dependency, since we follow Numba's dependencies one on one.

martinvonk:
Since we now use Numba with a global setting, we could maybe add Python 3.11 support by setting the global to `False`?

martinvonk:
> Since we now use Numba with a global setting, we could maybe add Python 3.11 support by setting the global to `False`?

Nevermind this comment, it won't work.

martinvonk:
Release candidate for numba with Python v3.11 support: https://github.com/numba/numba/issues/8841

martinvonk:
New Numba version released : https://github.com/numba/numba/releases/tag/0.57.0

So we should be good to go now. I already created a branch py311 for testing purposes https://github.com/pastas/pastas/tree/py311

Issue #451, opened 2023-01-09
[BUG] Python 3.11 support, Numba support


martinvonk:
Python 3.11 is currently not supported by Numba so maybe we shouldn't either? They are working on it though: https://github.com/numba/numba/issues/8304
Either way all tests on the dev branch are failing (see https://github.com/pastas/pastas/pull/441) since we added python 3.11 to setup.py :)

However, if we always want/need to follow Numba , should we enforce it as a dependency and not make it optional anymore? We also don't support NumPy 1.24 because of Numba. Or maybe we should do different tests on the dev branch that don't need Numba (and thus are slower).

martinvonk:
https://github.com/pastas/pastas/commit/7acc5046afaf9c32625d8d75ff2acf0e39ec7f87

martinvonk:
Closing because #452 exists

PR #468, opened 2023-01-18
452 Change Numba from optional to full dependency

martinvonk:
# Short Description
This pull request makes Numba a full on dependency of Pastas. This is because Numba basically always has to be used to make Pastas faster. Additionally, we follow the Numba dependencies for Python and NumPy one on one.

Numba decorators can still be turned off using the `USE_NUMBA` global variable set with `ps.set_use_numba(bool)`.

- [x] relates to issue #452 and #451 
- [x] is documented
- [x] black formatted code
- [ ] tests added / passed

### DIFF
--- a/pastas/decorators.py
+++ b/pastas/decorators.py
@@ -3,6 +3,13 @@ from logging import getLogger
 
 logger = getLogger(__name__)
 
+USE_NUMBA = True
+
+
+def set_use_numba(b: bool) -> None:
+    global USE_NUMBA
+    USE_NUMBA = b
+
 
 def set_parameter(function):
     @wraps(function)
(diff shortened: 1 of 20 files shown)


pastas/decorators.py
```diff
@@ -3,6 +3,13 @@
 
 logger = getLogger(__name__)
 
+USE_NUMBA = True
+
+
+def set_use_numba(b: bool) -> None:
+    global USE_NUMBA
+    USE_NUMBA = b
+
```
raoulcollenteur, line 12: Should this not be in utils.py ? It is not really a decorators. But I can see why you put it here..

raoulcollenteur (approved):

raoulcollenteur:
Is this not yet ready for review? I reviewed it. Looks good.

pastas/decorators.py
martinvonk: I agree that it's in a bit of an awkward place but let's keep it here. I think that it is most logical to keep it close to the njit decorator.

Issue #419, opened 2022-08-18
meanstress in StressModel contains confusion function


mbakker7:
In StressModel.init, the meanstress is computed as:

        if meanstress is None:
            meanstress = stress.series.std()

Why is that? The it should either be called stdstress rather than meanstress. Or the name remains `meanstress` and the `mean` function is called rather than the `std`.

rubencalje:
This was changed by me a few years ago, after an offline discussion. A time series model is influenced by the changes of a stress, and less by the absolute value of a stress. For example, when the stress is the air pressure or a river level, the absolute value of the stress-series is not so important, but the variation is. Therefore I think the standard deviation of the series is a better quantification of the influence of the stress on the head than the mean.

I agree that the name meanstress can be confusing, but one could argue that the real stress is the change in values of the stress-series, and the mean of this stress can be represented by the standard deviation of this series.

We could change the name to stdstress, but we need to make sure this is backwards compatible, as meanstress is stored in pas-files I think.

mbakker7:
I suggest to call it `scalefac` for scalefactor. Then when it is set for each stress clearly document what is used for the scalefactor (as is probably obvious from the computation of the scale factor).

raoulcollenteur:
what about `gain_scale_factor` to be more explicit? It's a bit long but given how not often users will provide it, that won't hurt most users. 

To-do-list
- [ ] Change self.meanstress attribute in rfuncs
- [ ] Change keyword in stressmodels and rfunc. 
- [ ] Deal with pas- files

mbakker7:
is fine. as long as the code is clear about the way it is scaled (mean, std, ...)

raoulcollenteur:
Fixed this issue in PR #483. No problems with backward compatibility as the use is really limited, and pas-files did not even store it.

raoulcollenteur:
`meanstress` was changed to `gain_scale_factor`. Providing `meanstress` produces a warning. This should be entirely removed for Pastas 1.0. Updating the milestone accordingly.

raoulcollenteur:
Closing after merging https://github.com/pastas/pastas/pull/504

PR #483, opened 2023-01-25
change meanstress to gain_scale_factor

raoulcollenteur:
# Short Description

In this PR the meanstress argument in rfuncs and stressmodels is changes to gain_scale_factor. Actually, only the StressModel has this argument.

Funny enough, there is no need to change anything in the loading of pas-files, because the argument was never exported to begin with (I fixed that in this PR as well, exporting through to_dict of StressModel). That made this change very easy to implement. 

- [x] closes issue #419 
- [x] is documented
- [x] PEP8 compliant code
- [ ] tests added / passed

### DIFF
--- a/doc/concepts/response_functions.ipynb
+++ b/doc/concepts/response_functions.ipynb
@@ -60,7 +60,7 @@
    "source": [
     "# Default Settings\n",
     "cutoff = 0.999\n",
-    "meanstress = 1\n",
+    "gain_scale_factor = 1\n",
     "up = True\n",
     "\n",
     "responses = {}\n",
@@ -150,6 +150,7 @@
    ]
   },
   {
+   "attachments": {},
    "cell_type": "markdown",
    "metadata": {
     "pycharm": {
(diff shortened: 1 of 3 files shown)


martinvonk:
I don't like this last minute pr. I'm not sure how if this gives any issues with Pastastore support for Pastas v0.23 if we merge this last minute. Maybe leave it for v1.0.0?

raoulcollenteur:
This should definitely be part of 1.0 and since there is a deprecation 0.23. I guess we'll have to wait with releasing 0.23 then. There are a couple of more changes around the rfunc that are necessary actually so perhaps better to wait.

martinvonk:
Can't we just only rename `meanstress` to `gain_scale_factor` and not adjust the to_dict method of StressModel as you mention below?

> Funny enough, there is no need to change anything in the loading of pas-files, because the argument was never exported to begin with (I fixed that in this PR as well, exporting through to_dict of StressModel).

Then I think this pr is fine.

raoulcollenteur:
We could do that. Then I'll open a issue with the bug that gain_scale_factor is not stored in pas-files. makes sense. Change coming :)

martinvonk (approved):

Issue #479, opened 2023-01-25
[ENHANCEMENT] Solvers as instances instead of classes


raoulcollenteur:
**Describe the proposed enhancement**

With the latest changes in 0.23.0b, users have to provide instances of classes virtually everywhere, except for the solver (e.g., `ps.Gamma()`, `ps.StressModel()`, `ps.rch.FlexModel()`, `ps.ThresholdTransform()`, `ps.NoiseModel()`). 

For consistency and ease of use, I propose to also change the way to how solvers are provided, to be instances. This has two advantages I think:

1. Pastas becomes entirely consistent in its expected behaviour, users provide instance. 
2. Users can provide additional arguments to the solver, which might be useful in the future.

A disadvantage is that the solver instance needs the Model instance, so the solver should be provided as below. 

**Expected behavior**

`solver = ps.LeastSquares(model=ml)
ml.solve(solver=solver)`

Anyone has any thoughts on this?

Raoul

martinvonk:
I'm in favor of making the the solver an instance to make Pastas consistent. I can imagine that in the future arguments such as _maximum number of function evaluations_ can be parsed to the solver.

mbakker7:
I agree that we should pass an instance, but the question is how to do that. 
I don't really like your suggestion.
First creating a solver with a model as input and then passing the solver to the model seems like a confusing circular operation. 

A solution is that the solver does not take the model as an input on initiation anymore. Instead, every solver has a `set_model` function that sets the model. Then the `solve` method of a model can simply set the model for the provided solver. In the code something like:
```
def solve(self, tmin=None, tmax=None, freq=None, warmup=None, noise=True,
        solver=None, report=True, initial=True, weights=None,
        fit_constant=True, **kwargs):

                # Store the solve instance
                if solver is None:
                    if self.fit is None:
                        self.fit = LeastSquares() # modify this line
                elif not issubclass(solver, self.fit.__class__): # not sure what this checks. what happens if this is False?
                    self.fit = solver() # modify this line
                self.fit.set_model(self) # add this line
```
BTW, I think `self.fit` to store the solver is a pretty cryptic name. Why not `self.solver`?

martinvonk:
Agree with @mbakker7. I didn't notice that you have to parse the Model to the solver class the original suggestion of @raoulcollenteur. I'd only like it if 
`ml.solve(solver=ps.LeastSquares)` becomes `ml.solve(solver=ps.LeastSquares())`.

[...]

Issue #180, opened 2020-03-19
Improve docs for Pastas v1.0


raoulcollenteur:
This is just an issue to keep track of some other changes that could be made to the docs:

- [x] Create an example notebook gallery https://sphinx-nbexamples.readthedocs.io/en/latest/examples/index.html
- [x] Improve "benchmarks" section
- [x] Better example Notebooks
    - [X] River levels
    - [x] StepModel
    - [x] LinearTrend
    - [x] Model verification / Diagnostics
    - [x] Solve Strategies
    - [x] Uncertainty Quantification
- [x] Add "recommended reading" section in user guide / or references list using sphinx
- [x] Automate adding Notebooks (seems to fail now) on RTD
- [ ] Give the notebooks from Groundwater paper a place on the Docs.
- [x] Add "Publications" section for peer-reviewed accepted articles using Pastas
- [ ] Make list of terminology used in Pastas
- [x] Name examples
- [ ] Check if sphinx-gallery >0.11. can be used (https://github.com/spatialaudio/nbsphinx/issues/655) or drop sphinx-gallery if this PR gets through for sphinxnb (https://github.com/spatialaudio/nbsphinx/pull/706/files)
- [ ] Start using sphinx >6.0 if pydata-sphinx-theme 0.13 is released (https://github.com/pydata/pydata-sphinx-theme)
- [ ] add example of setting solver settings

raoulcollenteur:
Let's add "better examples" to the list. Going through them made me realise they are not always that good, and we lack proper examples for most stress models. More specifically we need:

- new groundwater level data that is more clearly impacted by a rivers
- data showing a clear step trend
- data showing a linear trend (would be nice to test al well..)
- groundwater data where a gamma function is needed (no its often exponential)
- .. more?

Someone has these data that we can add?

dbrakenhoff:
I can find you a timeseries that is heavily influenced by a river.
For step and linear trend I would just create fictitious examples by adding manually.

dbrakenhoff:
[oseries_stresses_w_riv.zip](https://github.com/pastas/pastas/files/4360924/oseries_stresses_w_riv.zip)

This is a zipfile with:
- oseries 
- prec
- evap
- river (interpolated between stations to point nearest on river to observation point)

This oseries should have significant influence from the river.

raoulcollenteur:
Cool, I'll make a new example with river levels with that! Thanks!

martinvonk:
To add to the todo list:

[...]

PR #487, opened 2023-01-25
Solver as instance instead of classes

raoulcollenteur:
# Short Description

This PR addresses #479, and requires the users to provide instances of solvers rather than classes. If an instance is provided, Pastas will throw an DeprecationError now. 

No backward compatibility issues for this PR, only scripts will need to be changes. Existing pas-files will show no change.

- [x] closes issue #479 
- [x] is documented
- [x] PEP8 compliant code
- [x] tests added / passed


raoulcollenteur:
This PR is ready for review. I think everything should be good now after the last important changes. I build in a check to check if a model is already assigned (using set_model), because otherwise strange things may happen when a solver instance is used for different pastas Models.

raoulcollenteur:
I added the example request to this list, so I will add it when working on the documentation this week. https://github.com/pastas/pastas/issues/180 . Can you approve the PR if you think it is ready @mbakker7 ?

mbakker7 (approved):

Issue #486, opened 2023-01-25
[BUG] input arguments of stressmodels to export in to_dict


raoulcollenteur:
**Describe the bug**

For some stress models, user-provided input arguments are not exported in the `to_dict` method. For example, mean stress/gain_scale_factor may be provided to StressModel, but is not exported. This should be checked for all stressmodels. Basically every argument that is in the __init__ should be in to_dict to ensure the exact same Stress model is created when loading from pas-files.

raoulcollenteur:
Going to check this now, as this should really be done before releasing 0.23

raoulcollenteur:
Fixed after merging the last PR. Closing issue.

PR #497, opened 2023-01-26
Fix to_dict methods for stressmodels

raoulcollenteur:
# Short Description

Many of the to_dict methods for the different stressmodels were missing or wrong. I added the method everywhere and ensured that the exported keywords and values match those of the _init__ of each method. This ensures the stressmodels are the same when doing ml.copy or exporting/loading pas-files.

- [x] closes issue #486
- [x] is documented
- [x] PEP8 compliant code
- [x] tests added / passed

### DIFF
--- a/pastas/stressmodels.py
+++ b/pastas/stressmodels.py
@@ -1731,7 +1731,7 @@ class ChangeModel(StressModelBase):
 
         return h
 
-    def to_dict(self, series: Optional[bool] = True):
+    def to_dict(self, series: bool = True):
         """Method to export the ChangeModel object.
 
         Returns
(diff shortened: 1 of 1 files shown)


raoulcollenteur:
This PR is ready for review.

pastas/stressmodels.py
```diff
@@ -1678,6 +1731,29 @@ def simulate(
 
         return h
 
+    def to_dict(self, series: Optional[bool] = True):
```
martinvonk, line 1734: No `Optional[bool]` but `bool`.

`Optional` = `Union[type, None]` = `type | None`

martinvonk (approved):

pastas/stressmodels.py
raoulcollenteur: Thanks for the type hint. I'll change it. I thought this is that now this argument is optional but I guess not.

raoulcollenteur:
Thanks for the review @martinvonk ! I'll merge this as soon as GH Actions is done and green.

PR #507, opened 2023-01-30
Dynamic versioning

martinvonk:
# Short Description
Allows for dynamic version control as `__version__` is now retrieved from `__init__.py` and parsed to pyproject.toml.

Note that `pastas.version.show_version()` gets the version from the installed Pastas version, not from `__version__`. However, `pastas.__version__` gives us the one in `__init__.py`.

This means the version is now maintained in `__init__.py` instead of `pyproject.toml`.


mbakker7:
I installed the lastest version and it keeps telling me the version is '0.21.0b0'. 
While I can run all the latest updates. Does this have to do with the dynamic versioning?

dbrakenhoff:
Yep, welcome to hell :p

dbrakenhoff:
For a little more context:

With editable installs using `pip install -e .` the versioning of the package becomes a bit annoying. The new method of installing packages using `pyproject.toml` allows for dynamic versioning, but it always places a hard-coded version number in a metadata file in `site-packages`. So updating your local git repository will not modify this hard-coded version number and leave you very confused why `version("pastas")` keeps showing the wrong number.  

There are 2 options as far as I can tell so far, neither ideal, but option 2 is preferable in my opinion.

1. Version is set in pyproject.toml
   - A hard-coded copy of the version at the time of running `pip install -e .` is stored in `site-packages` in the metadata file. 
   - `version("pastas")` will always return this hard-coded version number.
   - This means the version number never changes, regardless of the status of the local git repository, until you force reinstall the package. 
2. `__version__` is set in `__init__.py`
   - Note that it is not possible to set the version in `version.py`, as the current version of setuptools (or whatever manages the installs) ignores this version number, and thinks it is installing pastas 0.0.0. Placing it in `__init__.py` does work for some reason. 
   - `version("pastas")` will still reproduce the hard-coded version number that is placed in `site-packages`, so for packages that need to check the pastas version, use `packaging.version.parse(pastas.__version__)`
   - This means `pastas.__version__` does update when the local git repository changes, which is good. But as stated the `version("pastas")` number will not change untill you force reinstall the package.

martinvonk:
Let me add on @dbrakenhoff comment and explain what this PR is for:

With the `transition` from `setup.py` to `pyproject.toml` #443 we now store the version in the `pyproject.toml` since Pastas 0.23.0. With setup.py the version was stored as a variable `__version__` in `pastas/version.py` and called from `setup.py`. This meant that calling the version (with `__version__`) would return you the version string in `pastas/version.py`.

We aren't the only Python package to make this transition and if you want to be able to call `__version__`, `importlib` is recommended which is a standard Python library since v3.8. So what we did in Pastas 0.23.0 is use `__version__ = importlib.metadata.version("pastas")`. This however does not read pyproject.toml but the pip installed metadata and that does not have to be the code that's currently imported. For most use cases, use without editable installations this is fine (and recommended). 

With editable installations you run in the problems of @mbakker7 and @dbrakenhoff mention above. So what we try to do with this PR is do dynamic version control using a (.py) file. This means that the __version__ is retrieved from a file in `pastas/` by both `pyproject.toml` and __version__. Since we use `setuptools` as `[build-system]` in `pyproject.toml` this should work since it is supported since `setuptools>46.0`*.

_*since we use `pyproject.toml` we can choose which `[build-system]` our package requires. E.g. `flint` or `poetry` could also be used in favor of `setuptools`. However, `poetry` does not (yet) support dynamic version control. `flint` does though but it works basically the same as `setuptools`._

martinvonk:
No clue why black fails on 01_basic_model.ipynb. Locally it is fine so...

dbrakenhoff (commented):
How come this does work, and putting __version__ into version.py does not...? 

I'm a bit annoyed by the need for 2 version.py files, can't we get rid of one of them somehow? Either by moving everything that remains in version.py elsewhere (utils? or __init__?) or by renaming it?

martinvonk:
> How come this does work, and putting version into version.py does not...?

Putting it in `version.py` also works as long as `__version__ = "x.x.x"` is defined. In `pyproject.toml` the path then needs to be:
`version = { attr = "pastas.version.__version__" }`

However, I thought it was a bit of an awkward spot to maintain the version because of all the functions that are also present in `version.py`. It makes it look a bit messy so I went looking for alternatives. Other conventions are to:
- create a file `_version.py` with `__version__ = "x.x.x"` . The underscore before the version.py indicates the immutability of the file so I like that.
- create a file `__about__.py` with __version__ = "x.x.x". Sometimes also `__name__` or other metadata are defined here.
- create a file `__version__.py` with __version__ = "x.x.x." but this also feels awkward to write `from __version__ import __version__`

`pyproject.toml` then of course gets `version = { attr = "pastas.file.__version__" }`

dbrakenhoff:
Ah, that makes sense.

I'm pro 1 file, I think. I don't mind the clutter. Or renaming version.py to something else, and having version.py contain only the version number. So I'll let you/the rest choose.

martinvonk:
> I'm pro 1 file, I think. I don't mind the clutter.

I don't mind. I discussed with @mbakker7 also and he is in favor of moving `__version__` to `version.py` so that's what I did :)

raoulcollenteur (approved):

raoulcollenteur:
As long as there is a single place for the version fine with me. Merging.

ofek:
> I also saw another package, `hatchling` which easy but capable: [hatch](https://hatch.pypa.io/latest/)

Hey maintainer of Hatch/Hatchling here; let me know if you have any questions should there be a desire to switch!

Issue #509, opened 2023-01-31
[ENHANCEMENT] Update (unit) testing and GitHub continuous integration with tox


martinvonk:
**Describe the proposed enhancement**
This enhancement proposes two a few things:
- improve testing by making tests more specific by separate testing for python versions, formatting etc.
- allow ALL tests to run locally in different (Python) environments

I want to do a few things:

1.  Improve the `pytest` implementation using fixtures for models.
Currently our tests have a lot of repetition test with a new model being created for every test function:
```
def test_add_stressmodel():
    ml = ps.Model(obs, name="Test_Model")
    sm = ps.RechargeModel(prec=rain, evap=evap, rfunc=ps.Gamma(), name="rch")
    ml.add_stressmodel(sm)
    return None

def test_add_stressmodels():
    ml = ps.Model(obs, name="Test_Model")
    sm1 = ps.StressModel(rain, rfunc=ps.Exponential(), name="prec")
    sm2 = ps.StressModel(evap, rfunc=ps.Exponential(), name="evap")
    ml.add_stressmodel([sm1, sm2])
    return None
```

However, this could be simplified (and sped up?) by using pytest fixtures.

```
@pytest.fixture
def ml():
    return ps.Model(obs, name="Test_Model")

def test_add_stressmodel(ml):
    sm = ps.RechargeModel(prec=rain, evap=evap, rfunc=ps.Gamma(), name="rch")
    ml.add_stressmodel(sm)
    return None

```

2. Use [tox](https://tox.wiki/en/latest/) to automate and standardize the testing. tox is an environment orchestrator. Use it to define how to setup and execute various tools on your projects. This can all be done locally so we can test it before committing to GitHub. Additionally we can define different routines for all the tests we want to run, e.g.:

- test runners (such as pytest and hopefully mypy in the future #480)
- linters (flake8)
- formatters (black and isort)
- documentation generators (sphinx)

This also makes it more clear which tests fail on GitHub ci, so we won't just see python 3.11 failed because it failed on black. But you see the formatting test failed.

martinvonk:
3. Test notebooks from documentation when releasing a new version on master.

dbrakenhoff:
Forgot to post my comment here :innocent: :

I address a small part of this in #515. Separating linting and testing.

raoulcollenteur:
Looks like all this is done in PR #515 now?

martinvonk:
> Looks like all this is done in PR #515 now?

[...]

PR #515, opened 2023-02-01
Improve testing routine and GitHub CI

dbrakenhoff:
# Short Description
Move linting out of testing step

Linting step checks:
- black formatting -> pass/fail
- isort formatting -> pass/fail
- flake8 -> always pass (but does show code style issues + statistics)

Testing step now only runs tests and notebooks.


martinvonk:
I like this change, can we relate this (partially) to #509 ?

martinvonk:
Really like this change. Thanks for the hard work on the ci.yml @dbrakenhoff :)

So with this change we can use _tox_ for testing all every change locally. Tox sets up virtual test environments locally (takes a bit of time the first time).
Run the formatting tests (black + isort) with: `tox -e format`
Run the linting tests (flake8) with: `tox -e lint`
Run pytest (without notebooks): `tox -e py38` (or `py39`, `py310 `for other python versions)
Run pytest (with notebooks): `tox -e notebooks`
The settings of these tests can be changed in `pyproject.toml` under `[tool.tox]`

GitHub actions uses tox to test the continuous integration also, giving a nice overview of the passing/failing of each specific test. In the future, building the docs (readthedocs) can also be done this way if we want. Same goes for static type checking with mypy.

This PR also adds the pytest.fixtures for different model types in the unit testing routine which is a nice improvement in terms of speed and test readability.

raoulcollenteur (approved):

raoulcollenteur:
Oh this is fun I need to change to settings for this PR. Which checks should be passed for a PR to be merged? Until now RTD, Py38-310 were required. I need to adjust this in the settings.

raoulcollenteur:
Just made everything "required", except for Coverage. Okay?

Issue #539, opened 2023-02-08
[ENHANCEMENT] Optional dependencies for workshops


martinvonk:
**Describe the proposed enhancement**
Create an optional dependency in the pyproject.toml for courses/workshops. We release a new version of pastas when we give a workshop anyways so I think this makes installing a lot easier. 

So attendees of workshops only have to install pastas using:
`pip install pastas[workshop]` where the optional dependencies are: full, hydropandas and jupyterlab?

PR #540, opened 2023-02-08
Update pyproject.toml

martinvonk:
Create an optional dependency in the pyproject.toml for courses/workshops. We release a new version of pastas when we give a workshop anyway so I think this makes installing easier. So attendees of workshops only have to install pastas using:
pip install pastas[workshop] where the optional dependencies are (currently): pastas[full], hydropandas and jupyterlab

- [x] closes issue #539

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,9 +69,9 @@ rtd = [
     "numpydoc",
     "sphinx-design",
 ]
-dev = ["tox", "pastas[formatting,linting,ci,rtd]"]
+workshop = ["pastas[full]", "hydropandas>=0.7.0", "jupyterlab"]
+dev = ["tox", "pastas[formatting,linting,ci,rtd,full]"]
 numbascipy = ["numba-scipy >= 0.3.1"]
-
 
 [tool.setuptools.dynamic]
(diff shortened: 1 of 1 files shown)


raoulcollenteur:
Should be add lmfit? And while we're at it, spotpy? I'll keep this PR open until 1.1.0, to see what new features are in that version and should be in the workshop installation.

martinvonk:
> Should be add lmfit? And while we're at it, spotpy? I'll keep this PR open until 1.1.0, to see what new features are in that version and should be in the workshop installation.

Does not really matter what it contain's. We can adjust it for every workshop since there is basically a release every workshop. No, lmfit is not required for the workshop notebooks the moment but I can imagine it could be in the future. However, I can imagine that latexify is useful for workshops. This pr is more about the concept of creating the possibility of `pip install pastas[workshop]` for workshop/course purposes. If SpotPy is implemented in #555 the pyproject.toml can be adjusted in that PR right?

martinvonk:
If SpotPy #555 is not implemented we can merge this right?

martinvonk:
Won't merge this #539, workshops have to find another way of installing the correct Python packages.

Issue #480, opened 2023-01-25
[Development] Static type checking and PEP 561 compliance


martinvonk:
**Describe the proposed feature**
Since v0.23.0 Pastas has implemented type hints. It would be nice to implement static type checking using `mypy` and make Pastas PEP [561 ](https://peps.python.org/pep-0561/) compliant. This improves the code quality and robustness. 

The downside I think is that this will be a lot of work (2days at least) because a lot of errors and issues will be raised by `mypy`. Also, Matplotlib, SciPy, Numba don't fully support type hinting, as that they are PEP [561](https://peps.python.org/pep-0561/) compliant. NumPy and Pandas are though, via `numpy.typing` and `pandas-stubs` respectively.

martinvonk:
We probably should remove the types in Pastas for packages that are not PEP 561 compliant. So for instance, NumPy and Pandas are PEP 561 compliant, but Matplotlib and SciPy aren't. Currently in Pastas we use we use `FigureBase` and `_AxesBase` from Matplotlib to define the `Figure` and `Axes` types in Pastas. However, these types were only implemented from Matplotlib>v3.3. 

So to:
```
Axes = TypeVar("Axes", bound=_AxesBase)  # Matplotlib Axes
Figure = TypeVar("Figure", bound=FigureBase)  # Matplotlib Figure

```
should be redefined in pastas.typing.types to: 
```
Axes = TypeVar("Axes", bound=Any)  # Matplotlib Axes
Figure = TypeVar("Figure", bound=Any)  # Matplotlib Figure
```

or we should require Matplotlib>v3.3

raoulcollenteur:
Should we drop this issue @martinvonk ? I don't think it is worth our time now..

martinvonk:
It depends, we can also leave it open on low priority. 

I think static type checking would improve our code quality. However, it is a lot of work... Running mypy on our current code yields: `Found 2051 errors in 56 files (checked 69 source files)`

mbakker7:
Won't fix for now. This doesn't cause any problems at the moment.

PR #553, opened 2023-02-15
Update types.py to support Matplotlib <v3.4

martinvonk:
# Short Description
Matplotlib only has `FigureBase` since >v3.3 so this gives this would give an `ImportError` if the Matploblib version of the user was <=v3.3. If we want to keep supporting: https://github.com/pastas/pastas/blob/730251140dab2392b88dc15050f6c2ed006bf4b9/pyproject.toml#L25
this small change has to be made.

- [x] closes part of issue #480


martinvonk:
Not sure why the tests fail. Locally things seem fine.

dbrakenhoff:
I think a new version of matplotlib or pandas is causing an issue between the two.

dbrakenhoff (approved):

Issue #136, opened 2019-08-06
[DEVELOPMENT] add tests for freq != "D"


dbrakenhoff:
I talked with @rubencalje and we think it would be a good idea to add some tests for frequencies other than daily to our test suite. Also the relationship between the obtained parameters when solving on different frequencies might be interesting to explore in a notebook.

(We discussed this because I was working on models with freq="14D" but I was not entirely sure this always worked as well as I'd hoped. I've now decided to solve on a daily basis and getting a sample every 14 days from my oseries.)

dbrakenhoff:
Pushing to release after next. I'll be working on models with different frequencies soon, but not before next release.

raoulcollenteur:
Perhaps also add some tests  and close this issue as well?

dbrakenhoff:
Add notebook

PR #556, opened 2023-02-17
Improve modeling with non-daily timesteps

dbrakenhoff:
# Short Description
@mbakker7 this should fix the issue you ran into recently

Currently a work in progress:
- [x] get freq, avoiding anchored freqs like W-MON
- [x] this should allow easier modeling with freq > "D"
- [x] check warmup calculation (see comment below)
- [x] remove rounding of index in get_equidistant_series_nearest()
- [x] add tests for non-daily models
- [x] add example for non-daily model?

- [x] closes issue #136 
- [x] is documented
- [x] tests added / passed


martinvonk:
relates to : https://github.com/pandas-dev/pandas/issues/15303

dbrakenhoff:
Alright, this is ready for review I think. 

What's been added/changed:
- modeling with freq > "D" is improved, by dealing pandas' annoying habit of converting a 7D freq to W-MON. The method `_infer_fixed_freq()` has been added to deal with this.
- a notebook is added with synthetic examples modeling with non-daily timesteps
- tests are added:
   - for modeling on hourly timestep (synthetic tidal simulation)
   - modeling D, 7D, 14D, and 28D (synthetic linear recharge simulation)

The remaining question I have (@raoulcollenteur, @mbakker7, @rubencalje) is what the warmup should be by default?
- Option 1 (current implementation):
   - 3650 * the freq supplied to ps.Model() or 3650 days:
      - `"T"` -> 3650 minutes
      - `"H"` -> 3650 hours
      - `"1H"` -> 3650 days  -> note the weirdness here :confused: 
      - `"D"` -> 3650 days
      - `"7D"` -> 3650 days  -> note it remains 10 years here
      - `"W"` -> 25550 days -> but not here :sweat:  
- Option 2:
    - Always 3650 days
- Option 3:
   - Always 3650 * freq
- Option 4:
   - Something smart to select a somewhat appropriate warmup period?

I don't think the logic behind Option 1 is desirable, so let me know what your preference would be!

mbakker7:
Regarding the warmup, I think 10 years is a good default value. 
But that means we need to convert that to the correct number depending on the frequency. 
When the warmup is provided by the user, we should also know what time units are meant. Not quite sure yet how to do that. Maybe the best option is to require a `pd.Timedelta`, for example, `pd.Timedelta(days=3650)`.

rubencalje:
The way it is now (in ml.simulate(), ml.residuals() and ml.solve()) is that the supplied warmup is interpreted as the number of days:
```
if not isinstance(warmup, Timedelta):
    warmup = Timedelta(warmup, "D")
```

This is also what is in the docs. The frequency should never have been part of the default value of warmup I guess.

So I would also think the default warmup period should always be 10 years. We can change the value from 3650 to 3652, but that is another question.

dbrakenhoff:
This is ready for review. 

Fixed the warmup to always be 10 years. The notebook contains some discussion on the warmup and shows how to modify the warmup.

raoulcollenteur (approved):

raoulcollenteur:
The new notebook does not show up automatically anymore after changes from @martinvonk. New notebooks need to be added manually. Perhaps add this to the PR before merging.

raoulcollenteur:
Thanks @martinvonk ! @dbrakenhoff, please check the docs: https://pastas--556.org.readthedocs.build/en/556/examples/modeling_timestep.html . Couple of headers are formatted wrongly.

Love the NB though, very nice example, or is it a benchmark? Not really sure how the categorise this one.

Cheers,
Raoul

raoulcollenteur:
Looks like pydata theme just updated to 0.13, which is great but apparently causes some issue with older sphinx versions. I'll fix that in another PR. Merging this one!

dbrakenhoff:
Oops, just committed an attempted fix :innocent: .

raoulcollenteur (approved):

Issue #542, opened 2023-02-08
[ENHANCEMENT] improve documentation of WellModel


dbrakenhoff:
be more explicit about what the distances mean, and check the rest of the docstring for good measure.

dbrakenhoff:
Fixed in #570

Issue #543, opened 2023-02-08
[ENHANCEMENT] better error message in WellModel if stress is not list


dbrakenhoff:
See title.

dbrakenhoff:
Fixed in #570

PR #570, opened 2023-03-21
improve WellModel

dbrakenhoff:
# Short Description
Improve WellModel:
- rfunc=None by default which creates HantushWellModel() instance.
- #542 improve docstring
- #543 improve error msg if stress is not list (by moving stress handling up in init)

Question: setting rfunc to None by default, which instantiates HantushWellModel() modifies the call signature of this function. How do we deal with that (with deprecations and such)?

- [x] closes issue #542, #543
- [x] is documented
- [x] tests added / passed

### DIFF
--- a/doc/examples/multiple_wells.ipynb
+++ b/doc/examples/multiple_wells.ipynb
@@ -321,7 +321,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "w = ps.WellModel(extraction_ts.values(), \"Wells\", distances)\n",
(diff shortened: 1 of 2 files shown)


raoulcollenteur (approved):

raoulcollenteur:
About the question, code that you used before still works right? And providing None never worked? So I think we can change this without any warnings?

dbrakenhoff:
I'm not entirely sure why I asked that question anymore, as far as I can tell you're right and the new code shouldn't cause any problems.

PR #581, opened 2023-03-24
Add (Pandas) resample wrapper function to timeseries_utils

martinvonk:
# Short Description
A lot of users have difficulty with resampling time series and choosing the right settings when interpolating a time series. This pull request creates a function `pastas.timeseries_utils.resample()` which uses pandas.resample but sets the default arguments for `closed` and `label` to `"right"`. This way users don't have to think that hard about it anymore when resampling and/or interpolating time series. This function can then also be used internally which is nicer I think.

- [x] is documented
- [x] tests added / passed

### DIFF
--- a/pastas/timeseries_utils.py
+++ b/pastas/timeseries_utils.py
@@ -467,9 +467,9 @@ def pandas_equidistant_nearest(
     Parameters
     ----------
     series : str
-        _description_
+        time series.
     freq : str
-        _description_
+        frequency string.
     tolerance : str, optional
         frequency type string (e.g. '7D') specifying maximum distance between original
         and new labels for inexact matches.
--- a/pastas/timeseries_utils.py
+++ b/pastas/timeseries_utils.py
@@ -524,21 +524,25 @@ def pandas_equidistant_asfreq(series: Series, freq: str) -> Series:
 def resample(
     series: Series, freq: str, closed: str = "right", label: str = "right", **kwargs
 ) -> Resampler:
-    """Wrapper around the pandas resample function with logical some Pastas defaults.
-       In Pastas, we assume the timestamp is at the end of the period that
-       belongs to each measurement. For more information on this read the
-       example notebook on preprocessing time series.
+    """Resample time-series data.
+
+    Convenience method for frequency conversion and resampling of time series.
+    This function is a wrapper around Pandas' resample function with some
+    logical Pastas defaults. In Pastas, we assume the timestamp is at the end
+    of the period that belongs to each measurement. For more information on
+    this read the example notebook on preprocessing time series.
 
     Parameters
     ----------
-    series : Series
-        time series
+    series : pandas Series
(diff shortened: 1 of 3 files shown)


pastas/timeseries_utils.py
```diff
@@ -518,3 +519,35 @@ def pandas_equidistant_asfreq(series: Series, freq: str) -> Series:
...
+    Returns
+    -------
+    Resampler
+        pandas Resampler object which can be manipulated using methods such as:
+        '.interpolate()', '.mean()', '.max()' etc. For all options see:
+        https://pandas.pydata.org/docs/reference/resampling.html
+
+    """
+
+    return series.copy().resample(freq, closed=closed, label=label, **kwargs)
```
dbrakenhoff, line 553: Is this copy necessary?

pastas/timeseries_utils.py
```diff
@@ -518,3 +519,35 @@ def pandas_equidistant_asfreq(series: Series, freq: str) -> Series:
         .squeeze()
     )
     return spandas
+
+
+def resample(
+    series: Series, freq: str, closed: str = "right", label: str = "right", **kwargs
+) -> Resampler:
+    """Wrapper around the pandas resample function with logical some Pastas defaults.
```
dbrakenhoff, line 527: word order

dbrakenhoff: + blank line after first descriptive sentence

pastas/timeseries_utils.py
```diff
@@ -518,3 +519,35 @@ def pandas_equidistant_asfreq(series: Series, freq: str) -> Series:
...
+    Parameters
+    ----------
+    series : Series
+        time series
+    freq : str
+        frequency string
+    closed: str, default 'right'
+        which side/end of bin interval is closed
+    label: str, default 'right'
+        which bin edge label to label bucket with
```
dbrakenhoff, line 541: bucket = bins?

pastas/timeseries_utils.py
martinvonk: Don't know. I just do it as a habbit. We do it for the other `timeseries_utils` functions as well, e.g.:
https://github.com/pastas/pastas/blob/730251140dab2392b88dc15050f6c2ed006bf4b9/pastas/timeseries_utils.py#L453

dbrakenhoff: The one you link is probably not necessary either, but in other cases we do modify something about the series, which makes the copy necessary to ensure we leave the original data untouched. 

We can remove this one though.

pastas/timeseries_utils.py
dbrakenhoff: Fair enough. I was just very slightly confused by bin/bucket in the same sentence :P. Fine to keep it this way.

dbrakenhoff (commented):
Good to go after removing the .copy()

PR #588, opened 2023-05-02
Support Python 3.11

martinvonk:
# Short Description
Since Numba supports Python 3.11 (numba >= 0.57) we can add support for Python v3.11 as well.

The optional dependency latexify-py does not support Python 3.11 yet (officially). I now install that from the GitHub page cause that should work but is not very robust. 

- [x] closes issue #452 
- [x] is documented
- [x] tests added / passed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ dependencies = [
     "matplotlib >= 3.1",
     "pandas >= 1.1",
     "scipy >= 1.8",
-    "numba >= 0.51",
+    "numba >= 0.57.0rc1",
 ]
 keywords = ["hydrology", "groundwater", "timeseries", "analysis"]
 classifiers = [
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -90,13 +93,13 @@ markers = ["notebooks: run notebooks", "bnotebooks: run benchmark notebooks"]
...
     requires = tox>=4
-    env_list = format, lint, notebooks, py{38,39,310}
+    env_list = format, lint, notebooks, py{38,39,310,311}
 
     [testenv]
     description = run unit tests
     extras = ci
     commands =
-            pytest tests -m "not notebooks"
+            pytest tests -m "not notebooks" -m "not bnotebooks"
```
raoulcollenteur, line 102: What does this line do? benchmark notebooks?

raoulcollenteur (approved):

pyproject.toml
martinvonk: Yes, so with #583 we now have seperate pytest markers for the benchmark notebooks.

PR #595, opened 2023-05-11
Update master to Pastas 1.1.0

raoulcollenteur:
# Short Description


pastas/modelcompare.py
```diff
@@ -365,10 +379,10 @@ def get_diagnostics(
         if models is None:
             models = self.models
 
-        diags = DataFrame(index=[x.name for x in models])
-        for ml in models:
+        diags = DataFrame(index=self.modelnames)
+        for i, ml in enumerate(models):
             mldiag = ml.stats.diagnostics()
-            diags.loc[f"{ml.name}", mldiag.index] = mldiag[diag_col].values
+            diags.loc[self.modelnames[i], mldiag.index] = mldiag[diag_col].values
```
dbrakenhoff, line 385: This doesn't look right? If models is passed in separately, then self.modelnames doesn't necessarily match that list/order

dbrakenhoff (approved):
Made some minor changes in a dbrakenhoff-patch-1.

After merging those, good to go as far as I'm concerned.

pastas/modelcompare.py
dbrakenhoff: Fixed in patch

Issue #602, opened 2023-06-06
Issues with WellModel [BUG]


rubencalje:
**Describe the bug**
There are three issues with the wellModel:

1 strange info message
When running the notebook http://localhost:8888/notebooks/pastas/doc/examples/multiple_wells.ipynb
During `ml_wm.solve()` I get the following info message:
`INFO: No distance passed to HantushWellModel, assuming r=1.0.`

2 sorting error, when distances are not sorted yet, this results in a key-errror:
`KeyError: 'Extraction_2'`

3 stacklegend in stacked_results gives the following error: 
`UnboundLocalError: local variable 'loc' referenced before assignment`

**To Reproduce**
1 Run the notebook in http://localhost:8888/notebooks/pastas/doc/examples/multiple_wells.ipynb
2 In the above notebook, change line
`w = ps.WellModel(list(extraction_ts.values()), "Wells", distances)` in:
`w = ps.WellModel(list(extraction_ts.values()), "Wells", distances[::-1])`
3 In the above notebook, run:
`ml_wm.plots.stacked_results(figsize=(10, 8), stacklegend=True);`

**Expected behavior**
1 No info message
2 No error
3 No error

**Python package version**
Python version: 3.9.7
NumPy version: 1.23.5
Pandas version: 1.5.3
SciPy version: 1.9.1
Matplotlib version: 3.7.1
Numba version: 0.56.4
LMfit version: 1.2.1
Latexify version: Not Installed
Pastas version: 1.1.0

dbrakenhoff:
I'll take a look.

1. The first issue is quite challenging to deal with, as discussed in #437 and #601. Whenever a call is made to `ml.get_parameters(name)`, no distances are provided to WellModel, and it uses `r=1.0`. I don't think there is a clean way to distinguish between an internal call to `rfunc.step` and an external one without adding loads of if/else's internally. Implementing an argument to silence the warning then results in lots of unnecessary `warn=False` being passed around just to get rid of this message. Buy maybe someone has a better idea? 

2 and 3 are simple fixes I will commit soon.

raoulcollenteur:
@dbrakenhoff and @rubencalje is this issue fixed after merging #603 and #604 ?

dbrakenhoff:
Yep, all fixed!

PR #604, opened 2023-06-06
silence HantushWellModel info msg about r=1.0

dbrakenhoff:
This PR silences the info message internally by adding a warn keyword argument and checking whether the current rfunc is HantushWellModel in `ps.Model` methods.

In my opinion this is a bit ugly, but perhaps worth silencing the confusing message? So far 3 issues have been posted about it, so it is confusing people.

- [x] closes issue #602
- [x] is documented
- [x] tests added / passed

### DIFF
--- a/pastas/model.py
+++ b/pastas/model.py
@@ -1547,7 +1547,11 @@ class Model:
 
     @get_stressmodel
     def get_response_tmax(
-        self, name: str, p: ArrayLike = None, cutoff: float = 0.999
+        self,
+        name: str,
+        p: ArrayLike = None,
+        cutoff: float = 0.999,
+        warn: bool = True,
     ) -> Union[float, None]:
         """Method to get the tmax used for the response function.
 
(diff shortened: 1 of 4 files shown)


raoulcollenteur:
@dbrakenhoff , thanks for the PR. I see that fixing this would be nice. 

I thought of another (easyier?) fix. Can we not deal with the distance=None issue when creating a WellModel instance? An assign r=1 if no distances are provided? That would fix it as well write? It seems to me that r should not be None in all cases anyway.

Cheers,
Raoul

dbrakenhoff:
That could work, and we could set the default value to 1 m or the distance to the first well. But that would mean users would have to be aware that when using `ml.<method>` some assumption is being made about which distance is used. I'm not sure if a silent assumption is the way to go here.

I think my preference is this PR, which provides a warning when users call `ml.<method>` on WellModels directly, but otherwise stays silent. But I'm open to others opinions/votes on this issue :).

raoulcollenteur (approved):

raoulcollenteur:
I agree that it's better to make there users aware and raise a warning. Still, could we not do that when creating the stress model without a distance and setting r to 1, and raising the warning once upon creation? Would prevent a lot of code and it would not be a silent assumption right?

dbrakenhoff:
> I agree that it's better to make there users aware and raise a warning. Still, could we not do that when creating the stress model without a distance and setting r to 1, and raising the warning once upon creation? Would prevent a lot of code and it would not be a silent assumption right?

The trouble is that the warning isn't likely to be generated at creation, but when users want to use the Model class convenience methods for getting information from the stressmodels (i.e. `ml.get_step_response`). So for now, I think this is the best solution, even though it does add some extra code.

Issue #633, opened 2023-09-05
[ENHANCEMENT] Add map with Pastas models around the world to docs


raoulcollenteur:
**Describe the proposed enhancement**

I propose to add a page to the docs showcasing Pastas models around the world. A start is here: 

- https://pastas.github.io/worldtour/

From the GitHub repo:

- https://github.com/pastas/worldtour

raoulcollenteur:
Closing this issue here. Further development of the actual page will be done here: https://github.com/pastas/worldtour

PR #637, opened 2023-09-06
Add map with Pastas models to docs

raoulcollenteur:
# Short Description

- [x] closes issue #633 
- [x] tests added / passed

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -108,6 +108,10 @@ html_context = {
     "doc_path": "doc",
 }
 
+html_sidebars = {
+    "map": [],  # Test what page looks like with no sidebar items
+}
+
 # -- Napoleon settings ----------------------------------------------------------------
 
 napoleon_include_init_with_doc = False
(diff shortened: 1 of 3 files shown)


raoulcollenteur:
I think we can actually merge this, as this PR only adds one page to the docs. The actual page is made in another repo and can be updated independently. 

The page can be viewed here: https://pastas--637.org.readthedocs.build/en/637/map.html

What do you all think?

martinvonk:
Awesome! Maybe we can make the vertical (and maybe also horizontal) space of the map on the web page a bit larger. I'm having a bit of trouble navigating.

dbrakenhoff:
Cool! Agree with Martin that it could use some styling but this is a good start!

FYI: some models are currently located in the Baltic Sea...

raoulcollenteur:
Thanks! I am trying to remove the right sidebar now to increase the width, hopefully I used the right code now.  Let's see after RTD build finished. 

Over the next couple of weeks I will update the actual website, using the HTML plots from poorly @dbrakenhoff ;-)

raoulcollenteur:
Worked! In addition I will make sure the sidebar will be toggled off by default so you only see the map :) Merging!

Issue #642, opened 2023-09-08
[DEVELOPMENT]Modify code for stressmodels without an rfunc


mbakker7:
In preparation of pastas plugins, we need to do some code refactoring so that stressmodels also work well without an rfunc. 
For example, when looking through the code, the rfunc._name is used sometimes, but if the stressmodel doesn't have an rfunc (i.e., `rfunc=None`) this doesn't work. Suggestion is to make sure to check whether an rfunc is not None first, although we don't want to do that all the time.

mbakker7:
Right now, it is checked whether the name of the rfunc is a certain string. It would be better to check whether rfunc `isinstance` of the required rfunc class rather than check the name. In that case rfunc can be `None` and the code still runs. @dbrakenhoff  would that work for you? I think your code is the primary user of this syntax. Any objection to replacing 
```if rfunc._name == 'HantushWellModel'```
to
```if isinstance(rfunc, ps.HantushWellModel)```

dbrakenhoff:
Good idea, I'll submit a PR with this change :+1:

dbrakenhoff:
Fixed in #643

PR #643, opened 2023-09-10
642 developmentmodify code for stressmodels without an rfunc

dbrakenhoff:
# Short Description
Ensure pastas continues working if StressModel has no rfunc.

- [x] closes issue #642
- [x] is documented
- [x] tests added / passed

### DIFF
--- a/pastas/model.py
+++ b/pastas/model.py
@@ -27,6 +27,7 @@ from pastas.io.base import _load_model, dump
 from pastas.modelplots import Plotting, _table_formatter_stderr
 from pastas.modelstats import Statistics
 from pastas.noisemodels import NoiseModel
+from pastas.rfunc import HantushWellModel
 from pastas.solver import LeastSquares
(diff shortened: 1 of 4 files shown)


mbakker7 (approved):

mbakker7:
Looks good. We may want to reconsider these checks for HantushWellModel in the future. A good modular design should not have if statements for certain stressmodels in these functions. In that case the function should be redefined in the stressmodel. But that may require more changes in the code. A topic for future discussion. For now this works fine and is an improvement over the current code with `._name`.

dbrakenhoff:
>  A good modular design should not have if statements for certain stressmodels in these functions.

Agreed, and the current design actually works fairly well. The only reason I put these ugly if-statements in is to avoid confusing warnings being triggered by internal Pastas calls, but yielding warnings when users use a method that is not aware of the distance parameter(s) in WellModel.

Issue #645, opened 2023-09-14
[BUG] Solver instance does not parse kwargs to solve method.


martinvonk:
**Describe the bug**
kwargs parsed to the solver are not parsed to the solve method.

**To Reproduce**

kwargs not parsed:
```python
solver = ps.LeastSquares(verbose=2)
ml.solve(solver=solver)
```

kwargs parsed:
```python
solver = ps.LeastSquares()
ml.solve(solver=solver, verbose=2)
```

raoulcollenteur:
I am not sure this is a bug but more how it is designed to work. Only arguments that will change some kind of attribute to the Solvers are passed to the solver, the rest is passed through ml.solve. There is no information moving from the solver to ml.solve, only the other way around. Seems more a question on how-to use Pastas to me.

Why would you need the first option?

martinvonk:
> the rest is passed through ml.solve.
This is not the case currently so that's why I made an issue. One would expect that the kwargs of the solver instance are parsed to the solve method.

I am fine with not making it an option to use the kwargs of the solver object, but then we should report that the kwargs of the solver are ignored / or remove it.

raoulcollenteur:
> I am fine with not making it an option to use the kwargs of the solver object, but then we should report that the kwargs of the solver are ignored / or remove it.

I think this would be the best option, otherwise there are two options to pass the kwargs, which might confuse other users. **kwargs should probably not be removed though, as it might affect loading pas-files with old arguments (that are now ignored but do not cause errors). Looking at the Solvers, we might actually want to add doctoring for BaseSolver, LeastSquares , and LmFitSolve to prevent people from doing what you tried to do ;-)

mbakker7:
I got lost in what the appropriate syntax is. 
I think it should be, as in the BaseSolver doc:
```
ml.solve(solver=ps.LeastSquares(solver settings))
```
Is that correct? What other ways are there that don't work?

mbakker7:
Reading back the bug report, it sounds like this is what @martinvonk tried to do and it doesn't work. 
I think this syntax should work and the other syntax shouldn't. Similar to the syntax for `rfunc`-s
So it should be:
```
ml.solve(solver=ps.LeastSquares(solver settings))
```
Any other option should get a deprecation warning.

raoulcollenteur:
I do not think that this is what we want, except for anything that should be an attribute of an instance. With how it is currently implemented all the arguments from least_squares in scipy can be used dynamically. e.g., you can use:

[...]

PR #646, opened 2023-09-14
Raise warning if kwargs parsed to solver instance

martinvonk:
- [x] closes issue #645 
- [x] is documented
- [x] tests added / passed


raoulcollenteur:
If we do it like this we could also store the kwargs in the pas-files I think. But I'm not convinced yet why this is needed.

raoulcollenteur:
This one should be good to go after passing CI, right, @martinvonk ?

Issue #667, opened 2023-11-29
Do not add a noisemodel to a Model by default anymore


raoulcollenteur:
A discussed in the last Pastas meeting, we plan to make `noise=False` the new default for ml.solve for Pastas 2.0 release. This Issue is to document the work related to this big change. Here are some of the reasons why we plan to make this change:

- This way we don't force users to use a noise model when they don't need it. 
- This will improve the calibration results in terms of goodness-of-fit for many beginning users.
- Some solve methods don't use a noise model anymore. This is for example already the case when using MCMC solvers and a likelihood function that accounts for autocorrelated errors. 
- If users need a noise model for any reason this will be a more conscious choice and still be possible.
- Other methods for UQ with autocorrelated errors are explored.
- .... ?

To  do list related to this Issue:

[ ] Investigate alternative methods for uncertainty quantification with autocorrelated errors.
[ ] Document this change throughout pastas
[ ] Provide FutureWarnings in the last version before 2.0 (whenever that is).
[ ] ...?

raoulcollenteur:
A similar warning as to Pandas could be given: 

`FutureWarning: The default of observed=False is deprecated and will be changed to True in a future version of pandas. Pass observed=False to retain current behavior or observed=True to adopt the future default and silence this warning.`

rubencalje:
One question we need to answer: what do we do with the parameter uncertainty, for example the standard error in the fit report, when noise is set to False? The uncertainty will likely be underestimated, as there will be much autocorrelation in the residuals. There are several options:

- just display the standard error (like we do now)
- do not display the standard error
- display the standard error in another way (Italic type or in orange/red)


I will start a branch to make some of the needed changes, so we can use this for testing.

mbakker7:
Good question. I think I am in favor of not displaying the standard error anymore whether the noise model is used or not. After all, just using it doesn't necessarily mean the autocorrelation is gone and so the standard error may still not be representative. Then we should probably create a separate function to compute the standard error.

rubencalje:
Comment from @mbakker7 in PR https://github.com/pastas/pastas/pull/678#issuecomment-2007939213:

> For version 2.0, I think the default should be to add a NoiseModel like adding a stress model. That way different noise models can be added. So my suggestion for version 2.0 is that ml.solve can not add a noise model. Come to think of it, this may have been the original implementation of a noise model in the first versions of pastas.

rubencalje:
[...]

PR #678, opened 2024-01-19
Do not add NoiseModel to Model-class by default anymore

rubencalje:
# Short Description
This Pull Request does not add a noisemodel to the Model class by default anymore.

See issue #735 for more information.

- [x] one of the needed changes to fix issue #667
- [x] is documented
- [x] tests added / passed


raoulcollenteur (approved):

raoulcollenteur:
Should we merge this for 1.4 or wait one version? Giving a warning early about this future change might be good?

rubencalje:
> Should we merge this for 1.4 or wait one version? Giving a warning early about this future change might be good?

I agree, an early warning is good. I am in favor of merging before the next version.

mbakker7:
This is a big change (although I am all for it). As this will give different results with the default values for many models, I think it warrants a major version number change. Should we go to 2.0 for this? 
Regarding a warning, what warning would you like to give. Every time ml.solve is called warn the user that in the future (2.0) the default will be noise=False? That would surely get annoying and we shouldn't do that for too many versions.

dbrakenhoff:
Perhaps a global "I got it, don't keep warning me" option would be nice to include, to avoid repeating the warning so often people stop reading it?

mbakker7:
> Perhaps a global "I got it, don't keep warning me" option would be nice to include, to avoid repeating the warning so often people stop reading it?

Any thought on how to implement this? You want to give this warning once after the first solve after a Pastas import?
I am mostly worried that people will turn warnings off. Which is what I would do in their case as well.

dbrakenhoff:
> Any thought on how to implement this? You want to give this warning once after the first solve after a Pastas import?
> I am mostly worried that people will turn warnings off. Which is what I would do in their case as well.

Maybe a global boolean setting that is flipped after one solve call? And can be turned off manually as well? And turning it off would then mean your code includes something like `ps.turn_off_default_noise_false_warning()` which indicates some awareness of the future change in default behavior? But I haven't completely thought it through yet, it was just an idea that popped up as I read the previous comments.

rubencalje:
People can just set `noise=True` in their code, which is what we want them to do. Implementing a global boolean or something else to bypass this warning will make them not set `noise=True` in their code, which will then give different results in the future. So I am not in favor of hiding the warning.

martinvonk:
I found this on [stackoverflow](https://stackoverflow.com/questions/9595009/warnings-warn-vs-logging-warning) so that you can raise warnings and log them without duplicating the message: [logging.captureWarnings(True)](https://docs.python.org/library/logging.html#logging.captureWarnings)

martinvonk:
Related to the comment above: this works for me in #700 to raise a FutureWarning in the logger:

```python
from logging import captureWarnings, getLogger
from warnings import warn

captureWarnings(True)
logger = getLogger(__name__)
warnings_logger = getLogger("py.warnings")
warnings_logger.addHandler(logger)

warn(
    "This function `kge_2012` will be deprecated in Pastas version 2.0. Please use"
    "`pastas.stats.kge(modified=True)` to get the same outcome.",
    category=FutureWarning,
)
```

pastas/model.py
```diff
@@ -685,7 +685,14 @@ def initialize(
         manually. See the solve-method for a description of the arguments.
         """
         if noise is None and self.noisemodel:
-            noise = True
+            msg = (
+                "The default of noise=True is deprecated and will be changed to "
+                "False in a future version of pastas. Pass noise=True to retain "
+                "current behavior or noise=False to adopt the future default and "
+                "silence this warning."
```
raoulcollenteur, line 692: Should we add a reference to issue here: https://github.com/pastas/pastas/issues/667 ? Then I think users are informed and we can merge this..

raoulcollenteur, line 692: Should we add a reference to issue here: https://github.com/pastas/pastas/issues/667 ? Then I think users are informed and we can merge this..

mbakker7:
For version 2.0, I think the default should be to add a NoiseModel like adding a stress model. That way different noise models can be added. So my suggestion for version 2.0 is that ml.solve can not add a noise model. Come to think of it, this may have been the original implementation of a noise model in the first versions of pastas.

raoulcollenteur:
@rubencalje, can you implement the same for ml = Model(noisemodel=True) and set the default to False? Then no noisemodel is silently added by default anymore and users have to be explicit there as well.

rubencalje:
> @rubencalje, can you implement the same for ml = Model(noisemodel=True) and set the default to False? Then no noisemodel is silently added by default anymore and users have to be explicit there as well.

I can do that, but I think we then need to change the default behavior of the noise-parameter in ml.solve() to: if you have added a noisemodel, solve with noise = True, and if you have not added a noisemodel, solve with noise = False.

We then remove the warning implemented now, and add a similar warning to the `__init__` of the Model-class.

rubencalje:
The default of the `noisemodel` parameter in `Model` is now changed to `None`, which will raise a warning that the behavior will change in the future. I removed the warning for the `noise` parameter in `ml.solve`, which now defaults to use the noisemodel only if it is added to a model (which is still the case by default now, the user just gets to see a warning).

Furthermore, I changed all relevant notebooks, so the documentation will not be full of warnings, while the resulting graphs stay the same. I hope I made all required changes. This PR is ready for review again.

raoulcollenteur:
Yes I think this makes more sense now, thanks! I think we just have one issue left to discuss (originally from @mbakker7), whether or not to remove the keyword `noise` in ml.solve at all or not. Then, if people want to solve with a noise model, they add one, and if not, they don't. That way it is very similar to adding a stressmodel or not, which I think makes sense. Let's discuss on Tuesday at the Pastas meeting and then merge this PR :)

rubencalje:
PR is ready to be merged again.

mbakker7 (approved):
This looks good. Bit step forward.
Should be quickly followed by https://github.com/pastas/pastas/pull/729 and https://github.com/pastas/pastas/pull/732

pastas/model.py
```diff
@@ -59,8 +59,9 @@ class Model:
     constant: bool, optional
         Add a constant to the model (Default=True).
     noisemodel: bool, optional
-        Add the default noisemodel to the model. A custom noisemodel can be added
-        later in the modelling process as well.
+        When true, add the default noisemodel to the model. A custom noisemodel can be
```
raoulcollenteur, line 62: Should we not set the default to None, and then if True or False is provided, raise an Error that this argument is deprecated and 1) no noisemodel is added by default anymore (no need for False), and 2) a noisemodel should be added by using Model.add_noisemodel() ? That way people can use 1.5 to clean up their code (like we did in the examples) to match the new pattern to adding a noise model.

pastas/model.py
```diff
@@ -747,7 +751,7 @@ def solve(
         tmax: Optional[TimestampType] = None,
         freq: Optional[str] = None,
         warmup: Optional[float] = None,
-        noise: bool = True,
+        noise: bool = True,  # will be removed in version 2.0.0
```
raoulcollenteur, line 754: Same here, I think we should put this argument to None, and raise a deprecationwarning when used (either true or False), to inform users to no longer use this code and that to use a noise model, one needs to be added.

pastas/model.py
```diff
@@ -692,8 +694,10 @@ def initialize(
             noise = True
         elif noise is True and self.noisemodel is None:
             logger.warning(
-                "Warning, solving with noise=True while no noisemodel is present. "
-                "noise set to False."
+                "Cannot solve with noise=True while no noisemodel is present. "
+                "Add a noisemodel with Model.add_noisemodel(ps.NoiseModel()) "
+                "to replicate results from pastas < 1.5. "
```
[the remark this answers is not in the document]
raoulcollenteur: plus ml.solve should not use pass this argument anymore so we don't raise the warning twice.

raoulcollenteur (changes requested):

raoulcollenteur:
Unless we want to bring out a version 1.6 before 2.0 this is not how we should make this change. Now users cannot use 1.5 to adapt their scripts to be able to run with 2.0 (i.e. when they get no warnings anymore the script should run with 2.0). If you want I can make the suggested changes.

rubencalje:
I thought we would just add these deprecated messages in version 2.0, and now focus on the change to not add a noise model.

You are free to edit this pull request.

raoulcollenteur:
I thought we wanted to go cold turkey and force users from 1.5 to adapt their scripts to the new behavior, and make the next version 2.0 where the noise and noisemodel arguments are removed at all. I made the necessary changes for that and committed these just now. I think if we give users very clear error messages (I tried to write it like that), and change the examples (which is already done), the transition is not that hard. 

Perhaps everyone can have a quick look and we have a short meeting tomorrow about this.. ?

rubencalje:
Now the user never gets to see a warning that no noisemodel is used anymore. That was the main goal of the PR...

And should we not warn the user the use of `noisemodel` and `noise` are deprecated, and not raise an error yet?

martinvonk:
> This looks good. Big step forward. Should be quickly followed by #729 and #732

I wanted to merge #732 into the `prepare_noise_is_false` branch. But we can also merge #732 quickly after this one.

dbrakenhoff:
> I thought we wanted to go cold turkey and force users from 1.5 to adapt their scripts to the new behavior, and make the next version 2.0 where the noise and noisemodel arguments are removed at all. I made the necessary changes for that and committed these just now. I think if we give users very clear error messages (I tried to write it like that), and change the examples (which is already done), the transition is not that hard.
> 
> Perhaps everyone can have a quick look and we have a short meeting tomorrow about this.. ?

Pastas 1.5 should communicate the new defaults when working in the old way (adding noisemodels in the model class) and also work the new way (explicitly adding noisemodels) without generating any warnings.

Pastas 2.0 should error on the old way, but I'm tempted to vote for still giving clear feedback to users that they're doing it wrong. That means leaving some arguments in, but raising exceptions when they are used. I know this isn't very pretty, but our previous shift to instances instead of classes for response functions went by too quickly for many users, who never noticed any warnings/errors. So my vote is for going lukewarm turkey :thinking:.

raoulcollenteur:
I agree it would be better to add a warning first in 1.5, but that also means we need to bring out 1.6 (which is fine with me). I think the challenge is we want to 1) change the default to False (no noisemodel) and 2) remove the keyword argument. This is basically impossible to do in one version smoothly unless we go cold turkey. So then we need at least two versions before 2.0 (1.5 and 1.6).

So I think ideally, we would:
1. in 1.5, set noisemodel to None, and make a warning that the default behavior changed and that the new method is to use ml.add_noisemodel. And then set noisemodel=True and add a noisemodel to keep the current behavior. To silence this, users would need to set noisemodel=False, and then use ml.add_noisemodel. This will be unpleasant for users and teaching but I don't think there is another way to inform users (other than the release notes and docs) about the new behavior. For ml.solve, the behavior will depend on the existence of a noisemodel, and not on noise=True/False anymore. I think we should set noise=None, and raise a warning that this argument is deprecated if used and determined by the existence of a noisemodel or not (e.g., for noise=False, use del_noisemodel). 
2. in 1.6, change the warnings to an error for when noisemodel is provided (like I implemented now), such that users can clean their scripts before 2.0. 
3. In 2.0, the entire argument noisemodel and noise is removed. We could still raise an error for these arguments if they are passed to inform the users.

This is how we implemented the changes for 1.0 as well (most notable, rfuncs as instances instead of classes).

rubencalje:
To silence the warning, people should then use a keyword that we are going to remove in the next version. I do not think this is useful.

Therefore, I made the changes so that the noisemodel is not added anymore from this version already, and the user would receive a warning when solving their model. So they know the behavior has changed. They can silence this warning by adding a noisemodel with ml.add_noisemodel().

raoulcollenteur:
In that case noisemodel should default to None, otherwise people who set it to False don't get an appropriate warning. But then people need to add noise=False to silence the warning from ml.solve, an argument that will also be removed. So to me raising errors on both the use of noisemodel and noise seems unavoidable... I think that would also be clearest to users to be honest.

raoulcollenteur:
This PR is ready for review again.

rubencalje:
I made some changes to the messages that I think will inform the user a bit better.

rubencalje:
Made a change so that loading of pas-files from versions before 1.5 that were solved with ml.solve(noise=False) can be loaded and will give the same results, by removing the noisemodel from these models. PR can be merged.

mbakker7:
Before putting this in dev, I think @martinvonk and @OnnoEbbens should merge their PRs in this PR first.

mbakker7 (approved):
I think this is ready to go. 
Big but important step forward.

raoulcollenteur (approved):

Issue #702, opened 2024-03-13
[BUG] When using WellModel ml.plots.results(block_or_step="block") raises error


rubencalje:
**Describe the bug**
When using WellModel the line

`ml.plots.results(block_or_step="block") ` 

raises

`TypeError: block() got an unexpected keyword argument 'warn'`

**Python package version**
Python version: 3.9.7
NumPy version: 1.23.5
Pandas version: 1.5.3
SciPy version: 1.12.0
Matplotlib version: 3.8.3
Numba version: 0.57.1
LMfit version: 1.2.1
Latexify version: Not Installed
Pastas version: 1.5.0b

raoulcollenteur:
I actually can't replicate the bug with the current development or master version. Are you sure this still happens?

rubencalje:
I only just now see your comment, sorry. We merged the fix for this issue into the dev-branch this morning. However, if you run the following code (copied from https://pastas.readthedocs.io/en/master/examples/multiple_wells.html, and added the last line) in the master branch from the doc/examples directory you will see the error:

```
import os
import numpy as np
import pandas as pd
import pastas as ps

# Specify coordinates observations
xo = 85850
yo = 383362

# Specify coordinates extractions
relevant_extractions = {
    "Extraction_2": (83588, 383664),
    "Extraction_3": (88439, 382339),
}

# calculate distances
distances = []
for extr, xy in relevant_extractions.items():
    xw = xy[0]
    yw = xy[1]
    distances.append(np.sqrt((xo - xw) ** 2 + (yo - yw) ** 2))

df = pd.DataFrame(
    distances,
    index=relevant_extractions.keys(),
    columns=["Distance to observation well"],
)

# read oseries
oseries = pd.read_csv(
    "data_notebook_10/Observation_well.csv", index_col=0, parse_dates=[0]
).squeeze()
oseries.name = oseries.name.replace(" ", "_")
# read stresses
stresses = {}
for fname in os.listdir("data_notebook_10"):
    series = pd.read_csv(
        os.path.join("data_notebook_10", fname), index_col=0, parse_dates=[0]
    ).squeeze()
    stresses[fname.strip(".csv").replace(" ", "_")] = series


prec = stresses["Precipitation"]
prec.index = prec.index.round("D")
prec.name = "prec"
evap = stresses["Evaporation"]
evap.index = evap.index.round("D")
evap.name = "evap"

extraction_ts = {}

for name in relevant_extractions.keys():
    # get extraction timeseries
[...]

PR #703, opened 2024-03-13
Fix issue 702

rubencalje:
# Short Description
Fix issue #702

This PR also fixes a bug where the use of `freq_obs` would result in the following error, when saving and reading a model from file again:
`ValueError: Weights and observations time series have different lengths! Check observation and simulation time series.`

- [x] closes issue #702
- [x] is documented
- [x] tests added / passed
- [ ] Remove output for all notebooks with changes

### DIFF
--- a/pastas/model.py
+++ b/pastas/model.py
@@ -480,12 +480,16 @@ class Model:
             tmax = self.settings["tmax"]
         if freq is None:
             freq = self.settings["freq"]
+        if self.settings["freq_obs"] is None:
+            freq_obs = freq
+        else:
+            freq_obs = self.settings["freq_obs"]
 
         # simulate model
         sim = self.simulate(p, tmin, tmax, freq, warmup, return_warmup=False)
 
         # Get the oseries calibration series
-        oseries_calib = self.observations(tmin, tmax, freq)
+        oseries_calib = self.observations(tmin, tmax, freq_obs)
 
         # Get simulation at the correct indices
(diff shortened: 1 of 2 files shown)


pastas/model.py
```diff
@@ -480,12 +480,16 @@ def residuals(
             tmax = self.settings["tmax"]
         if freq is None:
             freq = self.settings["freq"]
+        if self.settings["freq_obs"] is None:
+            freq_obs = freq
+        else:
+            freq_obs = self.settings["freq_obs"]
```
VascoSch92, line 486:
```suggestion
        freq_obs = self.settings.get("freq_obs", freq)
```

raoulcollenteur: Clearly won't work. settings does not have a get. Author appears to be pushing his GH profile.

rubencalje: Well, it does work: settings is a dictionary, and a dictionary has a get-method. However, when `settings["freq_obs"]` is `None`, it will not revert to `freq`. It only does when `"freq_obs"` is not in `self.settings`. So this change will not perform as desired. We can shorten the code by:
`freq_obs  = self.settings["freq_obs"] if self.settings["freq_obs"] is not None else freq`

VascoSch92: Ah ok because the key `freq_obs` is always present but with None as value right?

i see i see :-) my bad ;-)

dbrakenhoff (approved):

Issue #722, opened 2024-04-09
Show estimated parameter uncertainties or not


raoulcollenteur:
The estimated parameter uncertainties are now shown by default (in ml.plots.results and ml.fit_report). However, these may be reliable or not, depending on the properties of the residuals or noise ([as we clearly showed in this article](https://www.sciencedirect.com/science/article/pii/S0022169423010624#fig6)). Right now it is the responsibility of the modeler to check this, which I think is the case/default in most softwares. I am in favor of showing these estimates, but we should probably point out to users that these are only reliable under certain conditions. This could for example be in the docstrings of these methods.

Let's discuss. And decide before Pastas 2.0. :)

raoulcollenteur:
As of Pastas 1.5, standard errors will not be shown be default anymore, but only after setting stderr=True in ml.fit_report and in the plots. Closing issue after #722 was merged.

PR #729, opened 2024-04-09
722 show estimated parameter uncertainties or not

OnnoEbbens:
# Short Description
Remove the estimated parameter uncertainties from the fit report and the model results plot by default. Add an optional argument to include the parameter uncertainties in both.

The estimated parameter uncertainties are only accurate estimates if some conditions are met. This is described in this notebook: https://pastas.readthedocs.io/en/master/examples/diagnostic_checking.html. To avoid people using the uncertainties while they are not accurate we remove them by default from some visuals. 

- [x] closes issue #722
- [x] is documented
- [x] tests added / passed


rubencalje:
Maybe it is better to rename "par_uncertainty" to "stderr" or "par_stderr"?

dbrakenhoff:
or just write it out entirely :)? `parameter_stderr`

OnnoEbbens:
I renamed it to `stderr`

mbakker7:
If I am not mistaken, none of the examples in the docs now show the stderr. But there are a few notebooks where we check the diagnostics. Those would be good notebooks to modify and show that the stderr can added (and the good workflow is then to do a diagnostics check first).

OnnoEbbens:
You are correct, I've added the standard errors to the diagnostics and uncertainty notebooks. The standard error is also mentioned in the 'adding_trends' notebook (https://pastas.readthedocs.io/en/master/examples/adding_trends.html) for drawing a confidence interval. I think we can leave it like this. The only thing is that I don't see the confidence interval in the plot but that is another issue

raoulcollenteur:
I think there are a couple more notebook where we can show the stderr. I know that for recharge estimation this is possible. I'll have a look at that after merging PR #678 . Otherwise I think this looks good!

PR #752, opened 2024-04-30
Caching load_dataset and adding read_csv_kwargs

### LINKED ISSUE #721: Add cache=True to ps.load_dataset
A caching option should be added to the load-dataset method, because GH doesn't like it this way. GH Actions sometime fails now, so the tests should also be removed and only be executes when the master branch is updated.

martinvonk:
# Short Description

Add @lru_chache decorator to load_dataset and list_dataset to prevent #721 
Add `settings.json` with read_csv kwargs for each file in the dataset.

- [x] closes #721

### DIFF
--- a/pastas/dataset.py
+++ b/pastas/dataset.py
@@ -5,6 +5,7 @@ a subfolder in the pastas-data repository.
 
 """
 
+from functools import lru_cache
 from typing import Dict, List, Union
 
 from pandas import DataFrame, read_csv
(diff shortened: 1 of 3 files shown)


raoulcollenteur:
Looks good! lru_cache seems like a really neat function that we might want to explore more to use in pastas (i.e. ml.simulate?).

Merging!

raoulcollenteur (approved):

Issue #687, opened 2024-02-19
[DEVELOPMENT] Make Pastas compatible for Pandas 3.0


raoulcollenteur:
**Describe the proposed feature**
Pandas 3.0 will be released this summer. Some pattern won't work anymore after 2.2, so once we accept 3.0 we need to bumo minimum version to 2.2.

- [ ] replace "A", "A-MAR", and "M" by "YE", "YE-MAR", and "M"
- [ ]

martinvonk:
Pandas 2.0.3 gives a KeyError "YE" etc. Additionally, Pandas has no support for Python 3.8 since version >=2.1. So that's why the Python 3.8 tests failed with the "YE". So if we drop Python 3.8 only Pandas >=2.1 can be installed.

raoulcollenteur:
I added `pd_options.future.infer_string = True` to version.py in the dev branch to keep checking this. Only one issue arose so far so I think were good, but let's keep it until the next release.

martinvonk:
Done by @rubencalje, @raoulcollenteur, and @martinvonk in #1108

PR #754, opened 2024-05-14
Change frequency offset based on Pandas version

martinvonk:
# Short Description
Change frequency offset based on Pandas version
"YE" instead of "A" if the pandas version is lower than 2.2.0
"ME" instead of "M" if the pandas version is lower than 2.2.0

- [x] partially fixes issue #687

### DIFF
--- a/pastas/stats/dutch.py
+++ b/pastas/stats/dutch.py
@@ -8,6 +8,7 @@ groundwater time series in the Netherlands.
 from typing import Optional, Union
 
 from numpy import nan
+from packaging.version import parse as parse_version
 from pandas import Series, Timedelta
 from pandas import __version__ as pd_version
 from pandas import concat, date_range
@@ -15,7 +16,9 @@ from pandas import concat, date_range
 from pastas.timeseries_utils import get_sample
 from pastas.typing import Function, TimestampType
 
-year_offset = "YE" if pd_version >= "2.2.0" else "A"
+pandas_version = parse_version(pd_version)
+
+year_offset = "YE" if pandas_version >= parse_version("2.2.0") else "A"
 
 
 def q_ghg(
--- a/pastas/stats/dutch.py
+++ b/pastas/stats/dutch.py
@@ -8,10 +8,14 @@ groundwater time series in the Netherlands.
 from typing import Optional, Union
 
 from numpy import nan
-from pandas import Series, Timedelta, concat, date_range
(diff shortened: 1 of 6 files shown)


pastas/stats/dutch.py
```diff
@@ -8,11 +8,15 @@
...
 from numpy import nan
-from pandas import Series, Timedelta, concat, date_range
+from pandas import Series, Timedelta
+from pandas import __version__ as pd_version
+from pandas import concat, date_range
 
 from pastas.timeseries_utils import get_sample
 from pastas.typing import Function, TimestampType
 
+year_offset = "YE" if pd_version >= "2.2.0" else "A"
```
dbrakenhoff, line 18: use the following for more robust comparison of version numbers, e.g.:
```python
from packaging.version import parse as parse_version

PASTAS_VERSION = parse_version(ps.__version__)
```

pastas/stats/signatures.py
```diff
@@ -23,13 +23,18 @@
...
+from pandas import DataFrame, DatetimeIndex, Series, Timedelta
+from pandas import __version__ as pd_version
+from pandas import concat, cut, to_datetime
 from scipy.optimize import curve_fit
 from scipy.stats import linregress
 
 import pastas as ps
 from pastas.stats.core import acf
 
+year_offset = "YE" if pd_version >= "2.2.0" else "A"
```
dbrakenhoff, line 35: see comment above about comparing versions

pastas/stressmodels.py
```diff
@@ -1321,7 +1323,8 @@ def __init__(
             # precipitation in the world, then the precipitation is very likely in m/d
             # and not in mm/d. In this case a warning is given for nonlinear models.
 
-            if self.prec.series.resample("A").sum().max() < 12:
+            freq_offset = "YE" if pd_version >= "2.2.0" else "A"
```
dbrakenhoff, line 1326: see above.

dbrakenhoff (changes requested):
Looks good, but i would use the more robust version comparison to check the pandas version.

pastas/stats/dutch.py
dbrakenhoff: So it would be something like this I think:
```python
"YE" if parse_version(ps.__version__) >= parse_version("2.2.0") else "A"
```

dbrakenhoff (approved):

Issue #733, opened 2024-04-10
[ENHANCEMENT] Remove logger-initialization from Pastas


rubencalje:
**Describe the proposed enhancement**
Right now when importing pastas a specific logger for Pastas is initialized, by the following two lines in `__init__.py`:
```
logger = logging.getLogger(__name__)
initialize_logger(logger)
```
This adds a logger with a default level of "INFO". This is useful, as the user then sees all messages in his console. For example, when running example.py (with PR 678), you see:

```
WARNING: Cannot solve with noise=True while no noisemodel is present. Add a noisemodel with Model.add_noisemodel(ps.NoiseModel()) to replicate results from pastas < 1.5. Solving without a noisemodel.
INFO: Time Series 'rain' was extended in the past to 1975-11-17 00:00:00 with the mean value (0.0021) of the time series.
INFO: Time Series 'evap' was extended in the past to 1975-11-17 00:00:00 with the mean value (0.0016) of the time series.
```

If the user wants to change the log-level to WARNING, he can use the code `ps.set_log_level("WARNING")`, and you can remove the logger with the line `ps.utils.remove_console_handler()`. This works, but is made up by us (me), and I think is not the general way to do it Python. Integrating Pastas in other packages is therefore not so straightforward, and may result in double log-messages: one of the pastas-logger, and one of the logger initiated by the user. 

Therefore, it might be better to not initialize a logger in `__init__.py` anymore, by removing the two lines at the top of this message.  Then, by default of the logging-package, only warnings will be displayed. When running example.py (with PR 678) you then see:

```
Cannot solve with noise=True while no noisemodel is present. Add a noisemodel with Model.add_noisemodel(ps.NoiseModel()) to replicate results from pastas < 1.5. Solving without a noisemodel.
```

If as user wants to see log-messages of other log-levels, he can set this up with the logging-package (https://docs.python.org/3/library/logging.html):

```
import logging
logging.basicConfig(level=logging.INFO)
```

When running example.py (with PR 678) you then see:
```
WARNING:pastas.model:Cannot solve with noise=True while no noisemodel is present. Add a noisemodel with Model.add_noisemodel(ps.NoiseModel()) to replicate results from pastas < 1.5. Solving without a noisemodel.
INFO:pastas.timeseries:Time Series 'rain' was extended in the past to 1975-11-17 00:00:00 with the mean value (0.0021) of the time series.
[...]

PR #772, opened 2024-06-12
Remove logger-initialization in __init__.py

rubencalje:
# Short Description
This PR removes the logger initialization when importing Pastas. See issue #733 for more details. This will change the messages people receive in their console, as info-messages will not be displayed by default anymore. People can use the following lines in their script to show info-messages:

```
import logging
logging.basicConfig(level=logging.INFO)
```

This PR removes the setting of any logger handlers from Pastas, because of the follwing note in https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library: 

> It is strongly advised that you do not add any handlers other than NullHandler to your library’s loggers. This is because the configuration of handlers is the prerogative of the application developer who uses your library. The application developer knows their target audience and what handlers are most appropriate for their application: if you add handlers ‘under the hood’, you might well interfere with their ability to carry out unit tests and deliver logs which suit their requirements.

- [x] closes issue #733
- [x] is documented
- [ ] type hints for functions and methods
- [ ] tests added / passed
- [ ] Remove output for all notebooks with changes


rubencalje:
I do not understand the ruff-errors. This PR is ready.

doc/examples/example_logger.py
```diff
@@ -4,19 +4,38 @@
...
-ps.set_log_level("WARNING")
-# which is just a wrapper around
-ps.utils.set_console_handler(ps.logger, level="ERROR")
+# when we set the level to ERROR we do not see the warning anymore
+ps.set_log_level("ERROR")
+
+ts = ps.Model(obs)
+
+# when we set the log-level to "INFO", info-messages are still not printed
+ps.set_log_level("INFO")
```
raoulcollenteur, line 24: Should we not deprecate this method then?

raoulcollenteur (changes requested):

raoulcollenteur:
PR looks good. One thing before merging. 

ps.set_log_level method is basically rendered useless, no? I perhaps I misunderstand. But if it is not useful anymore we could remove it from the examples and deprecate it?

doc/examples/example_logger.py
rubencalje: The set_log_level-method is useful, as it sets the log-level of Pastas. The messages are still not printed, though, as we have not initialized a logger yet (which is what we do in the next line).

rubencalje:
> PR looks good. One thing before merging.
> 
> ps.set_log_level method is basically rendered useless, no? I perhaps I misunderstand. But if it is not useful anymore we could remove it from the examples and deprecate it?

ps.set_log_level is still usefull, so I am in favor of keeping it.

raoulcollenteur:
RUFF was reporting a missing line between imports :p Oh well... I'll merge after the checks are done.

raoulcollenteur (approved):

Issue #800, opened 2024-08-06
[ENHANCEMENT] One decorator to deprecate them all!


dbrakenhoff:
**Describe the proposed enhancement**
One decorator for all Pastas deprecations of classes/methods and functions. 

(Note, this is not intended for deprecated function arguments).

**Expected behavior**
- Shows a `logger.warning` if the object will be deprecated in the future. 
- Raises a `DeprecationWarning` if the current pastas version is greater than or equal to the version in which the deprecation was scheduled to occur. 

This allows us to deprecate classes/methods or functions in some future target version. We can leave that code untouched after the deprecation version has passed as the decorator will then raise an error. 

```python

# deprecate class
@PastasDeprecationWarning(remove_version="1.7.0", reason="Switch to the NotDeprecated class.")
class Deprecated:
    def __init__(self, a)
         self.a = a

# deprecate method
class Deprecated:
    def __init__(self, a)
         self.a = a
    
    @PastasDeprecationWarning(remove_version="1.7.0", reason="Use bar().")
    def foo(self, b):
         return self.a + b

# deprecate function
@PastasDeprecationWarning(remove_version="1.7.0", reason="Use bar().")
def foo(a):
     print(a)
```

As long as users are on a version <1.7.0, a warning is shown. From 1.7.0+ a DeprecationWarning is raised, which prevents users from using the class, but does provide feedback about this deprecation. It is then up to us to remove the code entirely, one or two versions after the deprecation was scheduled. 

**Code to reproduce the behavior**
```python
def PastasDeprecationWarning(remove_version: str, reason: str = "") -> Any:
    """Provide a warning or error when a Pastas class, method or function is deprecated.

    Logs a warning if the current Pastas version is lower than the version in which the
    class, function or method is removed. Raises a DeprecationWarning if the current
    Pastas version is higher than the version in which the class, function or method was
    removed.

    Parameters
    ----------
    remove_version: str
        The version in which the function or class will be removed.
    reason: str, optional
        The reason why the function or class is deprecated. Or provide a message
        that tells the user which alternative should be used.
    """

    def wrapper(obj: Any):
        name = obj.__name__

        def _function(*args, **kwargs):
            if CURRENT_PASTAS_VERSION < parse_version(remove_version):
                msg = (
                    "%s is deprecated and will be removed in Pastas version %s. "
                    % (name, remove_version)
                ) + reason
                logger.warning(msg)
            else:
                msg = (
[...]

PR #801, opened 2024-08-06
Add flexible PastasDeprecationWarning

dbrakenhoff:
# Short Description
The Deprecation decorator to rule them all!

- [x] closes issue #800 
- [x] is documented
- [x] tests added / passed

Developed with @martinvonk.

### DIFF
--- a/pastas/__init__.py
+++ b/pastas/__init__.py
@@ -1,4 +1,4 @@
-# flake8: noqa
+# ruff: noqa: D104 F401
 import logging
 import warnings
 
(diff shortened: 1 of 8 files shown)


dbrakenhoff:
I added a function to deprecate arguments or kwargs, similar to the logic contained in PastaDeprecationWarning.

If the deprecation is in the future, logger.warn, if the deprecation is in the past, raise DeprecationWarning.

To maintain the current logic w.r.t. the `noise` and `noisemodel` kwargs, I added a `force_raise` option. These two kwargs are deprecated from version 2.0.0 according to the message, but already lead to errors. This probably should not happen in the future as the message is contradictory.  Once we've properly deprecated these kwargs, we can remove the force_raise option. 

Another option would be to adjust the message and say noise/noisemodel were deprecated in version 1.5.0. Which is more a more accurate description of what we did probably :innocent: .

Let me know what you guys prefer.

raoulcollenteur (approved):

Issue #806, opened 2024-08-12
Python 3.13 support and Python 3.9 support drop


martinvonk:
Python 3.13 will be released on: 2024-10-01

**NumPy** already released version 2.1.0 that officially supports python 3.13. However they dropped support for Python 3.9: [numpy/releases/tag/v2.1.0rc1](https://github.com/numpy/numpy/releases/tag/v2.1.0rc1)
**Pandas**: https://github.com/pandas-dev/pandas/issues/58734 ([python 3.9](https://github.com/pandas-dev/pandas/blob/a3f2d487387aae8989b363c596810e57574e6ae8/pyproject.toml#L28) support already dropped)
**Numba**: https://github.com/numba/numba/issues/9413
**Matploblib**: [status](https://github.com/search?q=repo%3Amatplotlib%2Fmatplotlib+python+3.13&type=pullrequests) (but [python 3.9](https://github.com/matplotlib/matplotlib/blob/1e983776e9210fbc7426f28ce6e7eeadea18f9e9/pyproject.toml#L43) also already dropped)
**SciPy**: 1.14.0 supports python 3.13 (support for [python 3.9](https://github.com/scipy/scipy/blob/8d79d11773747ac9f643fe503e76b6242a966226/pyproject.toml#L47) dropped)

martinvonk:
Numba added NumPy 2.1 support: https://github.com/numba/numba/issues/9708

martinvonk:
Numba milestone: https://github.com/numba/numba/milestone/81

martinvonk:
Numba 0.61.0 rc: https://github.com/numba/numba/issues/9777

martinvonk:
Most of our dependencies have dropped python 3.9 support so I think we should too.

martinvonk:
Numba has a release candidate (0.61.0rc) with support for Numpy 2.1 and Python 3.13 so after the official release we can merge #807 .

PR #807, opened 2024-08-12
Add Python 3.13 and drop Python 3.9 support

martinvonk:
Adds tests that are allowed to fail. 

- [x] closes issue #806


martinvonk:
I think we can just leave this open 'till Numba adds python 3.13 support.

dbrakenhoff (approved):
Squash and Merge away!

Issue #625, opened 2023-09-04
Add check_model method


raoulcollenteur:
**Describe the proposed feature**

I propose to add a method that takes in a Pastas model, and checks against possible issues. This could help future users I think.

**Expected usage**

`ps.check_model(ml)`

e.g., Returns

- Response function longer than calibration time
- Parameter on bound
- ...

martinvonk:
Discussion on this topic: https://github.com/orgs/pastas/discussions/658

raoulcollenteur:
Introduced as per #829

PR #829, opened 2024-09-25
Add model check functionality

dbrakenhoff:
# Short Description
This is a first shot at introducing functions and a framework for checking time series models. The idea is that we have some library of methods for frequently used checks, and giving users a framework for adding their own checks. Currently the checks are programmed as functions that always take a pastas model as their first argument. Beyond that, any number of custom keyword arguments are allowed. Example:

```python
# given some solved model
ps.check.rsq_geq_threshold(ml, threshold=0.7)  # returns DataFrame
```

Multiple checks can be applied using `ps.check.checklist()` (function name can probably be improved). A check can be provided as:
- str, which means it must be the name of an existing check function in `ps.check`
- a function, either from `.ps.check` or programmed by the user themselves, taking a model as its first (and only argument)
- a dictionary, containing at least a "func" key with some function as its value. Any number of keyword arguments can be specified which will be passed to that function. This gives users full control over checks and keyword arguments.

```python
# apply multiple checks
checks = [
    {"func": ps.check.rsq_geq_threshold, "threshold": 0.6},  # provide dict for the most control
   "parameter_bounds",  # use str, which will use `ps.check.parameter_bounds`
]
ps.check.checklist(ml, checks)
```
The result is reported as a DataFrame containing:
- the test name (index), 
- the computed statistic, 
- the operator used to compare the statistic, 
- the threshold the statistic is compared to 
- the units of the statistic/threshold (if applicable)
- a boolean whether the check is a pass (true) / fail (false)
- an optional comment

This is an example of the output of multiple checks. The report function allows for some coloring of output in the DataFrame when running as a notebook:
image

## Questions
 - What are your general thoughts about this implementation?
 - Perhaps some function names can be improved?
 - Do we want to add a subclass `ml.check` to expose these methods from the model object directly?
    - I added this in later commits: e.g. `ml.check.rsq_geq_threshold(threshold=0.7)`
 - Any other checks we should program and add?

- [x] closes issue #625 and discussion #658 
- [x] is documented
- [x] tests added / passed
- [ ] TODO: Example Notebook (for new features)


pastas/check.py
```diff
@@ -0,0 +1,550 @@
...
+            Parameters of the response function.
+        rfunc: pastas.rfunc
+            Response function instance.
+
+        Returns
+        -------
+        tmem: float
+            Time to the cutoff value, i.e. the memory of the response function.
+        """
+        t = rfunc.get_t(p, dt=1.0, cutoff=0.999)
```
dbrakenhoff, line 137: add warning if cutoff > 0.999

pastas/check.py
```diff
@@ -0,0 +1,550 @@
+import logging
+from typing import Callable, Optional, Union
+
+import numpy as np
+from IPython.display import display
```
martinvonk, line 5: i don't think IPython is a dependency

dbrakenhoff: Isn't automatically installed if you install jupyter?

martinvonk: Yes but `jupyter` is not a dependency either.

pastas/check.py
```diff
@@ -0,0 +1,555 @@
...
+            ">",
+            n_std * std,
+            guess_unit(parameter),
+            check,
+            "Assumes estimate of σ is reliable.",
+        ]
+    return df
+
+
+def guess_unit(parameter):
```
martinvonk, line 356: This function returns both units and dimensions. So maybe we should only give dimensions. Because units are undoable I think.

pastas/check.py
```diff
@@ -0,0 +1,555 @@
...
+    elif stressmodel is not None and not isinstance(stressmodel, list):
+        stressmodel = [stressmodel]
+
+    df = DataFrame(
+        columns=["statistic", "operator", "threshold", "unit", "pass", "comment"]
+    )
+    df.index.name = "check"
+    unit = "days"
+
+    def interp_step(cutoff: float, p: np.ndarray, rfunc: RfuncBase):
```
martinvonk, line 119: For the `np.ndarray` type hint. I think you can also do; `np.ndarray[float]`. Otherwise you could also use `ps.typing.ArrayLike`.

martinvonk (approved):
Very nice addition. Only have some comments about some naming "conventions".

One thing we might consider as well, since we are supporting python >= 3.10 is to replace the `Union` with the `|` operator. So `Optional[Union[list[str], str]]` can be written as `list[str] | str | None` which is a lot cleaner imo. You can do this here or I can make a pull request for it later where I do it everywhere.

pastas/check.py
```diff
@@ -0,0 +1,555 @@
...
+
+    Returns
+    -------
+    df: pandas.DataFrame
+        DataFrame with the results of the check.
+    """
+    return _diagnostic_test(ml, "stoffer_toloi", p_threshold=p_threshold, **kwargs)
+
+
+def _diagnostic_test(ml: Model, test: str, p_threshold: float = 0.05, **kwargs):
```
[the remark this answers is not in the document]
dbrakenhoff: `alpha` is used several places in Pastas, and doesn't always mean the same thing. But I agree alpha is commonly used in this context, so I'll change it to alpha.

pastas/check.py
```diff
@@ -0,0 +1,555 @@
...
+                check,
+                "",
+            ]
+    return df
+
+
+def uncertainty_gain(
+    ml: Model,
+    n_std: float = 2.0,
+    stressmodel: Optional[Union[list[str], str]] = None,
```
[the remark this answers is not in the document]
dbrakenhoff: I guess we use `name` in other functions to denote stressmodel name, so I guess that's the way to go?

pastas/check.py
dbrakenhoff: For some cases we definitely know the units, so it felt weird not to return them, but I suppose going for just dimensions is more consistent.

dbrakenhoff: I've switched it to returning dimensions, but sort of keeping support for both. I renamed the unit column to dimensions. Once others take a look we can determine what we want to do with this...

raoulcollenteur: Did you want to rename the method to dimensions or can we resolve this comment?

dbrakenhoff: the method is now called `guess_unit_or_dims()` which returns dimensions by default (through `return_dims=True`).

raoulcollenteur (approved):

PR #831, opened 2024-10-01
Run uncertainty estimation benchmark notebook with multiprocessing

martinvonk:
# Short Description
Makes the notebook run much faster, if not on Windows 🪟 or MacOS  🍏 
Also updates the readthedocs section because the notebook was still missing.


doc/benchmarks/pastas_uncertainty_benchmark.ipynb
dbrakenhoff, line 1: Just one question: didn't macOS also require some adjustment to be able to fork the current process because the default was spawn?

dbrakenhoff (approved):

doc/benchmarks/pastas_uncertainty_benchmark.ipynb
mbakker7: I don't know, but will let you know after I try it.

mbakker7: Doesn't work on MacOS indeed. Long error something like

```
Process SpawnProcess-2:
Traceback (most recent call last):
  File "/Users/mark/anaconda3/lib/python3.11/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/Users/mark/anaconda3/lib/python3.11/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/mark/anaconda3/lib/python3.11/concurrent/futures/process.py", line 244, in _process_worker
    call_item = call_queue.get(block=True)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mark/anaconda3/lib/python3.11/multiprocessing/queues.py", line 122, in get
    return _ForkingPickler.loads(res)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: Can't get attribute 'model1' on <module '__main__' (built-in)>
```
And then more like that.

martinvonk:
I'll remove the parallel option for MacOS as well. Biggest benefit is for the GitHub CI I think.

doc/benchmarks/pastas_uncertainty_benchmark.ipynb
martinvonk: > Any reason we don't use the multiprocess package? https://pypi.org/project/multiprocess/

The `concurrent.futures` library does the same as the `multiprocessing` library, but the code is cleaner / simpler to use.

Issue #826, opened 2024-09-17
[BUG] Model parameters are reinitialized for a model with adjusted parameters in pas file


martinvonk:
**Describe the bug**
Load pas file where parameter (bounds) are adjusted. Solve the model. Parameter bounds are set back to Pastas' defaults.

PR #833, opened 2024-10-17
Load parameter bounds and vary if changed in pas file

martinvonk:
# Short Description
Currently, if the `Model.parameters` `DataFrame` was changed in a .pas file only the initial parameters were reloaded correctly. The `pmin`, `pmax` and `vary` values were not parsed to the loaded model. Even though the values from the pas file were visible in the `ml.parameters` `DataFrame`, the old/default parameters were still present in the `ml.stressmodels[name].parameters` `DataFrame` because it was not loaded correctly. 

- [x] closes issue #826

### DIFF
--- a/pastas/io/base.py
+++ b/pastas/io/base.py
@@ -150,9 +150,15 @@ def _load_model(data: dict) -> Model:
     # Convert parameters to numeric
     ml.parameters = ml.parameters.infer_objects()
 
-    # When initial values changed
-    for param, value in ml.parameters.loc[:, "initial"].items():
-        ml.set_parameter(name=param, initial=value)
+    # When parameter initial values and bounds changed
+    for pname, data in ml.parameters.iterrows():
+        ml.set_parameter(
+            name=pname,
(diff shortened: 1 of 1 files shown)


pastas/io/base.py
```diff
@@ -150,9 +150,15 @@ def _load_model(data: dict) -> Model:
     # Convert parameters to numeric
     ml.parameters = ml.parameters.infer_objects()
 
-    # When initial values changed
-    for param, value in ml.parameters.loc[:, "initial"].items():
-        ml.set_parameter(name=param, initial=value)
+    # When parameter initial values and bounds changed
+    for pname, data in ml.parameters.iterrows():
```
dbrakenhoff, line 154: not a big issue, since its the last thing in this function, but you're overwriting data variable from before.

dbrakenhoff (approved):
I would rename the variable in the loop. Then good to go.

pastas/io/base.py
martinvonk: good eye

PR #841, opened 2024-11-13
Fix typos

mwtoews:
This fixes a handful of typos identified using interactive [codespell](https://github.com/codespell-project/codespell).

[-] closes issue #847


raoulcollenteur:
Hi Mike, Thanks for the PR! Codespell looks like a very useful tool. Do you have any experience using it as a pre-commit tool on GI Actions by any chance? I'll go through the PR later, but looks good after briefly looking at it.

Cheers, Raoul

mwtoews:
I've not set it as a pre-commit, but I can amend a commit here to add the configuration for it.

martinvonk:
How fast is `codespell`? If it takes over 0.1 s or so I'd be against using it with a pre-commit hook. 

Also, imo typo's do not matter that much that we would need to check for it in each commit/pr. So just running code spell now and somewhere in the future would be fine for me. 

/ 

Thanks for the contribution @mwtoews !

raoulcollenteur:
So perhaps a pre-release hook? I agree it would not make sense for every commit, but it would be nice when releasing a new version?

mbakker7:
+1 as a pre-release hook
(Also a good idea: Install a spell checker for the software that is running your Jupyter Notebooks.)

mwtoews:
I've added a configuration for codespell to this PR as a first step. It is added to pyproject.toml, so either it needs Python >3.11 or `tomli` installed to work.

The tool is slow, around 9 seconds, so not ideal as a pre-commit hook. As for a pre-release hook, I don't see any existing GHA for this. Should it just be mentioned in `doc/developers/new_release.rst`?

mwtoews:
This is ready for review. The spell check happens in CI.

raoulcollenteur (approved):

raoulcollenteur:
Thanks @mwtoews, looks good to me! 

@martinvonk can you have another look? Looks fine to me. I think Codacy is just complaining because it is a fork so we can ignore that.

mbakker7:
Just for my information, @mwtoews , I presume you checked 'american english' as the language to check against? (is there even an option for other english versions?).

martinvonk:
I don't think that this is the way we want to set up the CI for `codespell`. Because now it will be run on every push. I think we want to run it only on PR's to the `master` branch. So that requires a separate `yaml` in the `.github/workflows`.

mwtoews:
codespell doesn't use a conventional (e.g.) US or UK dictionary, but contains a massive list of English typos with corrections. For instance, it would change "mimimize->minimize" and "mimimise->minimise", but it would never change between "minimize<->minimise" (which would be annoying!)
I don't believe there are any plans to support any other languages, but custom dictionaries can be added, which can contain anything.

PR #842, opened 2024-11-18
Benchmark notebook on response identification via cross-correlation

martinvonk:


martinvonk:
@mbakker7 can you review this ? :)

dbrakenhoff:
It might be worth including a brief mention why we're using `pastas_plugins.ccf` instead of `ps.stats.ccf` :)?

mbakker7:
This notebook looks nice, but I think it can be even nicer. So let's move it to the next release.

Issue #860, opened 2025-02-03
[ENHANCEMENT] Update plots.series with colors and other enhancements


mbakker7:
The `plots.series` function is very useful, but all series are plotted black. 
Suggestion: Plot heads as black dots and use default matplotlib colors to plot the rest by default.
Enhancement: Let the user specify colors.

Vishnu-sai-teja:
Can take this up

martinvonk:
I also propose to add a `sharex` keyword argument that shares the x-axis of only the time series plots. This can easily be done with:
`from pastas.plotting.plotutil import share_xaxes`

martinvonk:
And while we are at it, I think the `kde=True` does not always work and shows only the kernel density estimate for the head time series.

GovertAlkemade:
i can take this up today.

PR #865, opened 2025-02-03
Enhance plot colors in series function

Vishnu-sai-teja:
Related to #860


mbakker7:
Thanks @Vishnu-sai-teja for your contribution. I agree with @martinvonk. 
Furthermore, we are having our first Pastas coding sprint this Friday and this issue would be a great one for one of our first-timers to solve. So if you don't mind, I close your PR for now and see if one of the first-timers can solve it on Friday. If you don't mind checking back in a week or so to see  if this is resolved, that would be great. Thanks again for submitting this.

PR #875, opened 2025-02-07
Adding more colors by default and also the option to specify colors for plots.series()

GovertAlkemade:
# Short Description
Adding more colors by default and also the option to specify colors for plots.series()

- [x] closes issue #860
- [x] is documented

### DIFF
--- a/pastas/plotting/plots.py
+++ b/pastas/plotting/plots.py
@@ -200,7 +200,7 @@ def series(
         for i, stress in enumerate(stresses, start=rows - len(stresses)):
             stress = stress[tmin:tmax].dropna()
             if colors_stresses is None:
-                colors_stresses = list(mcolors.BASE_COLORS.keys())
+                colors_stresses = list(mcolors.TABLEAU_COLORS.keys())
             try:
                 color_stress = colors_stresses[i-1]
             except Exception as e:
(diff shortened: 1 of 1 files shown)


mbakker7:
Thanks for your first PR to Pastas! There is a bit of work left, but you are getting close. 

Your code doesn't run yet on my example, I think, because the `colors_stresses` is not a keyword argument of the `ps.plots.series` method. So please add that.

Also, to check if something is None, you should use `if color_stresses is None` rather than `if colors_stresses == None`.

mbakker7:
Please change `mcolors.BASE_COLORS` to the default matplotlib colors.

GovertAlkemade:
Hi Mark, i changed it to TABLEAU_COLORS, which is the default matplotlib uses.

mbakker7:
This all looks nice. Final change is that in `pastas` we don't print warnings to the screen. So if the color list is is too short, my suggestion is to simply plot black and not print anything. And update the doc string to state that. Something like:
List with the matplotlib colorcodes to use for plotting each stress timeseries. If list is shorter than number of stresses, the remaining stresses are plotted in black. If None (default), default matplotlib colors will be used.

GovertAlkemade:
ah, i understand. Should be fixed now.

mbakker7 (approved):
**Looks good. More enhancements of function are possible, such as grids, but that is for another issue**

Issue #819, opened 2024-09-07
[ENHANCEMENT] Remove rmsn from ml.stats.summary() if no noise model


raoulcollenteur:
**Describe the proposed enhancement**
Remove rmsn from ml.stats.summary() if no noise model is present in the model, instead of returning nan like now.

nicolesullivam:
I can pick this up

dbrakenhoff:
Closed by #876

PR #876, opened 2025-02-07
Do not compute RMSN when no noise model is present in the model

nicolesullivam:
Issue is solved, added a test to check if there are NaN values in the stats summary.
A test model without the noise model would be needed. Thanks

[ ] Closes issue #819


dbrakenhoff:
In conftest.py you can add a new fixture:

```python
@pytest.fixture
def ml_no_noisemodel() -> ps.Model:
    ml = ps.Model(obs, name="Test_Model")
    sm = ps.RechargeModel(prec=rain, evap=evap, rfunc=ps.Gamma(), name="rch")
    ml.add_stressmodel(sm)
    return ml
```

In your test you can then use this ml_no_noisemodel

dbrakenhoff (approved):
LGTM :shipit:

dbrakenhoff (approved):
Nice! :shipit: 

Merging after tests pass.

PR #886, opened 2025-02-10
Matplotlib keyword args

mbakker7:
# Short Description

- [ x] is documented
- [ x] Format code with [ruff formatting](https://docs.astral.sh/ruff/)
- [ x] type hints for functions and methods
- [ ] tests added / passed

### DIFF
--- a/pastas/plotting/plots.py
+++ b/pastas/plotting/plots.py
@@ -101,7 +101,7 @@ def series(
         List with the labels for each subplot.
     figsize: tuple
         Set the size of the figure.
-    kwargs: 
+    kwargs:
         keyword arguments passed to plotting functions of stress timeseries
 
     Returns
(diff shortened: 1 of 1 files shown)


mbakker7:
I think it is ready to go

martinvonk:
very nice, 
I also made some changes:
- kde is now fixed for stresses due to min/max error
- kde is grey if hist else same color of series
- histogram is not plotted twice anymore if kde=True (and hist=True) 
- use prop_cycle colors instead of tableau colors (which are by default the same of course)
- use pastas `share_xaxes` function

martinvonk (approved):

mbakker7:
Looks good, @martinvonk . Only question I have is why to use pastas share_xaxes rather than standard (and easy) matplotlib functions. Removes one more function we have to keep current.

martinvonk:
> Looks good, @martinvonk . Only question I have is why to use pastas share_xaxes rather than standard (and easy) matplotlib functions. Removes one more function we have to keep current.

The share_xaxes pastas wrapper does the same as matplotlib does by default when you call `sharex=True` in a figure. Then not only the axes is shared (`sharex()`), but also tick labels are hidden. David and I once spend a lot of time to figure this out :)

> One more thing, why not print an xlabel for the time series. Can we then at least add an option, as a label along an axes is often required in journals and reports.

Pandas by defaults plot the xlabel from the index name. And that index name is never the label you would want in a publication. And the xlabel of all series plots often overlaps with the title of the figure below.  Additionaly, for time series the indices already make it clear the x-axis is time so it is a bit redundant.

Issue #894, opened 2025-02-25
Update unit tests (fixtures)


martinvonk:
I would like to be able to debug the unit tests locally. This is sometimes difficult to do outside of pytest due the fixtures. Also the path to the fixtures data files is relative. Something to look into and should be possible:
https://docs.pytest.org/en/stable/deprecations.html#calling-fixtures-directly

PR #895, opened 2025-02-26
894 update unit tests fixtures

martinvonk:
# Short Description
 I restructured the following:
- Let fixtures use fixtures
- Allow for running pytest from any directory instead of only pastas base directory.
- Fix PICP tests because they were not run because of an indent

I wasn't able to solve #894 completely but the issue can be closed I think. Debugging the tests locally should just require some extra, self written, code to replace the fixture input.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -43,19 +43,19 @@ def head() -> Series:
 
 
 @pytest.fixture
-def rm() -> ps.RechargeModel:
+def rm(prec: Series, evap: Series) -> ps.RechargeModel:
     rm = ps.RechargeModel(prec=prec, evap=evap, rfunc=ps.Gamma(), name="rch")
     return rm
 
 
 @pytest.fixture
-def sm_prec() -> ps.StressModel:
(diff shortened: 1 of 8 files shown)


raoulcollenteur:
Something with a notebook going wrong in the tests, can you have a look?

raoulcollenteur (approved):

raoulcollenteur:
Looks good to me. Just as a clarification, we should not use rm for the name of the RechargeModel anywhere else (I don't mind it here), because it is also a terminal command and can cause issues in scripts and notebooks..

I think you can merge this after fixing the notebook issues, something with the dates I think.

martinvonk:
> Looks good to me. Just as a clarification, we should not use rm for the name of the RechargeModel anywhere else (I don't mind it here), because it is also a terminal command and can cause issues in scripts and notebooks..

I always use `rm` for RechargeModel 😎  #livingontheedge

raoulcollenteur:
image

raoulcollenteur (approved):

Issue #853, opened 2025-01-13
[BUG] ACF function sometimes returns values above 1 or below -1


raoulcollenteur:
**Describe the bug**

I noticed that the ACF function sometimes returns values above 1 or below -1, which should not be possible. I'll add an example later, but wanted to open the BUG issue now
 
**To Reproduce**
`#Code to reproduce the behavior`

**Expected behavior**

**Screenshots**

**Python package version**
 - run `ps.show_versions()` for this

**Additional context**

dbrakenhoff:
This has (maybe) been noted before, posting links here in case it is related? 

- In https://github.com/pastas/pastas/issues/810 
- And this note to an earlier fix in #96 :

> Note: the use of the number of overlapping elements to scale the result can lead to calculated correlations larger than 1.0, or to perfectly matching series having a correlation that is smaller than 1.0. This is caused by the fact that the two series are normalized at start of the calculation. When shifting one series relative to another, only the overlapping part of the series is considered and scaled by the number of overlapping elements. This overlapping part of the series might not satisfy the conditions of a normalized series (mu=0.0 and sigma=1.0 ). This means the maximum correlation of the overlapping parts can be either larger or smaller than the number of overlapping elements. When scaling by the number of overlapping elements this can yield correlations unequal to 1.0 for series that match one another perfectly. This effect is presumably smaller when the maximum lag << observation period and the total number of measurements is large.

PR #896, opened 2025-02-26
Improve the CCF/ACF function and make it more robust by using gaussian kernels by default

raoulcollenteur:
# Short Description

- Change standard bin_method to "regular" which was already the case under the hood, but not clear to the user.
- Add fallback options for irregular data, but set to gaussian which is much more robust then the previous default (rectangle). The latter was sometimes causing values above |c| > 1 
- If values above |c|>1, throw a warning to inform the user, as it indicates that the CCF/ACF could not be estimated well.
- adapt regular function to return nan lag exceeds the lag (prevents errors in the rare case).

- [x] closes issue #853 
- [x] is documented
- [x] tests added / passed


raoulcollenteur:
This one is ready for review! I think it would be good to address this soon, as the acf function returning values above and below 1 is not something we want and users should be warned.

dbrakenhoff:
So, a few questions about the CCF implementation in pastas that I think we should tackle and (maybe) solve here:

When I use `scipy.signal.correlate()` on two 1d-arrays, I get the following result:

```python
from scipy.signal import correlate

a = np.array([0, 1, 2, 3, 4, 5])
b = np.array([5, 4, 3, 2, 1, 0])

correlate(a, b, mode="full")

# --> array([ 0,  0,  1,  4, 10, 20, 35, 44, 46, 40, 25])
````

When I do something similar to what `pastas.stats.ccf()` does, I get the following:
```python
import numpy as np

cc = np.zeros(len(a))

for lag in range(len(a)):
    n = len(a)
    cc[lag] = np.correlate(a[: n - lag], b[lag:])
cc

# --> array([20., 10.,  4.,  1.,  0.,  0.])
```

So pastas is returning the first part of the results in reverse order. Not a big deal, since flipping `a` and `b` will give you the other half. But when computing the cross-correlation (as compared to e.g. `pastas_plugins.cross_correlation.ccf()`) you need to flip the terms to get the same result. In a synthetic example, you can show that you get back the block response:

```python
from pastas_plugins import cross_correlation as ppcc
import pastas as ps

# given some series head and stress (e.g. precipitation)

cc_plugin = ppcc.ccf(head, stress)
cc_pastas = ps.stats.ccf(stress, head)
```

image

**Question 1:** Should we make these two behave similarly? And if so, what is the "logical" order of the arguments? For what it's worth the statistics textbook referenced in the CCF example at pastas_plugins essentially uses CCF(head, stress).

**Question 2:** `pastas.stats.ccf` returns a timedelta index, wouldn't it be more convenient (e.g. for plotting) to return a float index (and store the unit in the index name)?

**Question 3:** `pastas.stats.ccf` doesn't return the 0 lag, which makes sense for the ACF since it is 1 by definition, but wouldn't it make sense to return the 0 lag result for the CCF?

pastas/stats/core.py
```diff
@@ -118,12 +127,13 @@ def ccf(
     x: Series,
     y: Series,
     lags: ArrayLike = 365,
-    bin_method: str = "rectangle",
+    bin_method: str = "gaussian",
```
dbrakenhoff, line 130: How come we use "gaussian" as default here rather than "regular" like in the `acf()` function?

pastas/stats/core.py
```diff
@@ -162,15 +176,35 @@ def ccf(
     Examples
     --------
     >>> ccf = ps.stats.ccf(x, y, bin_method="gaussian")
+
+    Notes
+    -----
+    The CCF method primarily tries to estimate the autocorrelation using common
```
dbrakenhoff, line 182: autocorrelation -> cross-correlation or just correlation

dbrakenhoff (commented):
Two notes on the changes. And some additional questions in my previous comment :).

pastas/stats/core.py
raoulcollenteur: That should be regular here indeed!

raoulcollenteur:
Thanks for the review! I just updated the code based on your comments. The default argument for bin_method should indeed be regular, which it was already under the hood. My responses to the questions:

Q1: I think it should be dependent on the input order, and the current implementation is correct. I think scipy.signal.correlate uses convolution, which always returns the same ccf regardless of the order. See also the figure here: https://en.wikipedia.org/wiki/Cross-correlation

Q2: This could be a separate Issue/PR, I like the intervals, but we can discuss. Changing it costs some effort from our side with little benefits, I think.

Q3: That is the default, if a user supplied that lags argument with an array that includes 0, it should be returned. I think I mainly chose this because for the ACF you don't want to test for lag-0. Changing this default could be another Issue/PR.

I would prefer to make new issues and PRs if we want to make any changes to code. 

Cheers
Raoul

dbrakenhoff:
Fine to tackle these questions in separate issues, but since the discussion is here, I'll just briefly comment here:

Q1:
> scipy.signal.correlate uses convolution, which always returns the same ccf regardless of the order.

The CCF result in `scipy.signal.correlate` is flipped when you flip the input arguments, as you would expect I think. I'm fine keeping our order, but it should be noted that with the same order of the arguments pastas is returning the first half of the CCF flipped as compared to `scipy.signal.correlate`.

Q2: I finally found out that PeriodIndex has a `.days` argument, so never mind my complaint :P

Q3: I will open a separate issue for this, it should be really easy to default CCF to return lag 0 and not for the ACF.

dbrakenhoff (approved):

dbrakenhoff:
Merging, tackling other issues in separate PR.

Issue #866, opened 2025-02-04
Add documentation string to ml.stats.diagnostics


raoulcollenteur:
**Describe the proposed enhancement**
The current implementation of ml.diagnistics does not have a docstring. This PR proposes to add one, that can probably be copied from ps.stats.diagnostics (perhaps even in the code).

**Expected behavior**
ml.plots.diagnostics? should return a docstring.

`ml.plots.diagnostics?`

PR #897, opened 2025-02-26
Use decorator to copy the docs from one method to another

raoulcollenteur:
# Short Description
This PR adds a docstring to ml.stats.diagnostics via a decorator. Not good for VScode, but good for the docs and works as soon as the code is run. So instead of copying the exact same docstring around I think this is the best option, but open for alternatives.

- [x] closes issue #866
- [x] is documented
- [x] tests added / passed

### DIFF
--- a/pastas/decorators.py
+++ b/pastas/decorators.py
@@ -190,6 +190,7 @@ def latexfun(
 
     return latexify_decorator
 
+
 def documented_by(original: Optional[Function] = None) -> Function:
     def wrapper(target):
         target.__doc__ = original.__doc__
--- a/pastas/decorators.py
+++ b/pastas/decorators.py
@@ -189,3 +189,10 @@ def latexfun(
         return latexify_decorator(function)
 
     return latexify_decorator
+
+def documented_by(original: Optional[Function] = None) -> Function:
+    def wrapper(target):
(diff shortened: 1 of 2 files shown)


pastas/decorators.py
```diff
@@ -189,3 +189,11 @@ def latexify_decorator(f: Function) -> Function:
         return latexify_decorator(function)
 
     return latexify_decorator
+
+
+def documented_by(original: Optional[Function] = None) -> Function:
```
martinvonk, line 194: I don't think original can be Optional?

raoulcollenteur: Oui. I'll change that too. Thanks !

OnnoEbbens:
The `diagnostics` method and function do not have the same arguments, so the docstring will be incorrect if you copy it.

Personally I would not add the extra complexity of decorators and instead manually copy the docstring. Especially if it is just for this function

mbakker7:
I agree with @OnnoEbbens if it is just for a couple of functions

raoulcollenteur:
Thanks @OnnoEbbens for looking at it, I agree with you the decorator is probably a bit of overkill :) I will close this PR and create a new one with a docstring copied.

Issue #902, opened 2025-02-26
[BUG]Starting values of walker values in EmceeSolver can be negative


mbakker7:
**Describe the bug**
The starting values of the walkers are `pinit = pinit + 1e-2 * np.random.randn(self.nwalkers, ndim)`. Hence when `pinit` is small, the starting values can be negative, which is undersirable.

**Expected behavior**
Positive starting values. Maybe simply `pinit = pinit + np.absl(pinit) * 1e-2 * np.random.randn(self.nwalkers, ndim)`

mbakker7:
If `pinit` is negative, the values remain negative. So that should work.
It may be good to make `pinit` a keyword argument that can be set by the user, as the choice we make now (based on the recommendations of the Emcee developer) is just one of many options to choose initial values.

Issue #890, opened 2025-02-20
[BUG]Some incompatibility between the doc string, names, and code of the Log Likelihood Functions


mbakker7:
The doc string, equations, and names of variables are not consistent in the Log Likelihood functions.

For the `GaussianLikelihood`, the docstring has $\sigma^2$ (correct) but this variable is called `sigma` in the code. 

For the `GaussianLikelihoodAr1`, the same problem occurs with $\sigma^2$. Furthermore, in the docstring the autoregressive parameters is called $\phi$, but in the code it is called `theta`, and there is talk about $\mu$, which doesn't exist at all. $\phi$ is actually a function of the time interval $\Delta t$ between observations, so it should be stated somewhere that the current implementation is for constant time intervals ($\sigma^2$ is also a function of $\Delta t$). It would be nice to have a reference to the literature for the `GaussianLikelihoodAr1` function.

raoulcollenteur:
I'll work on this over the next couple of days and put this into PR with Emcee initial values and notebook so we update the whole emcee worksflow at once with the new release. I thought I added the reference but apparently not, so I will add it. I have to look at some local files because I thought I also programmed another likelihood function for heteroscedastic errors as well but I can't find that either. Perhaps I forgot to push some changes....

mbakker7:
Would indeed be great if this is all fixed together. My suggestion for `sigma` is either `sigsq` or `var` (for variance). And use `phi` rather than `theta` (which we already use for response functions elsewhere).

raoulcollenteur:
I just updated the code on the [emcee branch](https://github.com/pastas/pastas/pull/903). I renamed `sigma` to `var` and `theta` to `phi`.

Here's the reference for the formulas (*Table 1, eq. A and D) : https://www.sciencedirect.com/science/article/abs/pii/S0022169415004011

I'll add the likelihood for heteroscedastic, autocorrelated errors (Eq. G) as well, because I would like to test that for some models I encountered that have this problem.

PR #903, opened 2025-02-26
Update EmceeSolver class and Likelihood functions

mbakker7:
# Short Description

This PR implements some essential changes and documentation improvements to the Emcee Solver and related likelihood functions.

- [x] closes issue #902, and #890


mbakker7:
ReadTheDocs is failing. Which makes it hard to check the new docs....

mbakker7:
The Bayesian uncertainty analysis Notebook looks weird in ReadTheDocs. Seems like the progress bar doesn't work and finally throws an error.

raoulcollenteur:
> The Bayesian uncertainty analysis Notebook looks weird in ReadTheDocs. Seems like the progress bar doesn't work and finally throws an error.

That is actually a timeout of RTD, which stops running a notebook if it takes longer than 30 seconds to run the entire notebook. I updated the notebooks section to try and get these run on RTD.

mbakker7:
The Bayesian uncertainty notebook runs now (although different name in Navigation panel on left than on the Examples page). The corner plot doesn't look as good anymore (fewer steps? fewer walkers?). I couldn't find the new notebook anywhere yet. Where did that land?

mbakker7:
On a related note: the notebook "Parameter (uncertainty) estimates benchmarked with synthetic time series" also throws an error on ReadTheDocs. Probably taking too long as well.

raoulcollenteur:
The new notebook is now listed and can be found here: https://pastas--903.org.readthedocs.build/903/examples/uncertainty_ls_mcmc.html

It suffers the same problem as the other notebooks that take too long to run. I think for now however we should accept that because the validity of the results depend on sufficient steps. For the other MCMC notebook this is less important because it only shows how it can be used wiuthout checking the results. 

I will open another issue to discuss and solve the issue of NBs taking to long to run and not being displayed well on RTD so we can merge this one.

Issue #784, opened 2024-06-27
Check response tmax with warmup period and add warnings


dbrakenhoff:
**Describe the proposed enhancement**
As discussed in #782. It would make sense to not just compare the response tmax to the calibration period, but also the warmup period. Too short warmups relative the the response function mean that simulated results can change when the tmin/tmax are adjusted. This is confusing for users. 

**Expected behavior**
- Add a warning to the fit_report at the bottom when tmax exceeds warmup.
- Optionally add a warning when this occurs: 
  `"Response tmax for '{stressmodel name}' > {warmup} warmup period. Note that simulated heads can change when adjusting tmin."`

raoulcollenteur:
This check is also implemented in the checks module I think. But adding it to the fit report might still be an option or not?

PR #910, opened 2025-02-28
#868 add tmin tmax to CompareModels.plot

AnneDeGraaf:
# Short Description
Modifies ps.plots.compare to take in tmin/tmax, such that one can compare the models better.

- [x] closes issue #868 
- [x] is documented
- [x] tests added / passed


pastas/plotting/modelcompare.py
```diff
@@ -948,15 +1010,15 @@ def plot(
...
+        _ = self.plot_simulation(tmin=tmin, tmax=tmax)
 
         # res
-        _ = self.plot_residuals()
-        _ = self.plot_noise()
+        _ = self.plot_residuals(tmin=tmin, tmax=tmax)
+        _ = self.plot_noise(tmin=tmin, tmax=tmax)
 
         # smn, rfn
-        _ = self.plot_contribution(smdict=smdict, normalized=normalized)
```
AnneDeGraaf, line 959: I didn't add the `tmin` `tmax` kwargs to `plot_contribution`. Does that seem right to you? @raoulcollenteur

pastas/plotting/modelcompare.py
```diff
@@ -936,8 +990,16 @@ def plot(
...
             contributions in one subplot, please also provide smdict for best results.
+        tmin : TimestampType, optional
+            start date for the simulation period (E.g. '1980'). If none is provided, the
+            tmin from the oseries is used.
+        tmax : TimestampType, optional
+            end date for the simulation period (E.g. '2010'). If none is provided, the 
+            tmax from the oseries is used.
         legend_kwargs : dict, optional
             pass legend keyword arguments to plots.
+        **fig_kwargs
```
AnneDeGraaf, line 1001: documentation was missing here so added this

AnneDeGraaf:
example output:
```
mc = ps.CompareModels(models=[ml1b, ml1a])
mc.plot(tmin="2011", tmax="2014")
```
4a5577f0-2c9b-40d5-b4e4-ca21bd02f2cf

martinvonk:
Hi @AnneDeGraaf, 
Thank you for this pr :) Looks good!

I think it would be nicer if the tmin and tmax are attributes. That way you don't need tmin and tmax as arguments to each plot function. But you can just call self.tmin and self.tmax.
So something like this:
```python
class CompareModels:
def __init__(self, models, tmin: Optional[pd.Timestamp] = None, tmax: Optional[pd.Timestamp] = None)
    self.models = models
    self.tmin = tmin
    self.tmax = tmax

def plot_simulation():
    simulation = ml.simulate(tmin=self.tmin, tmax=self.tmax)
    simulation.plot()
```

AnneDeGraaf:
@raoulcollenteur  I added the `tmin` and `tmax` kwargs separately to six different CompareModels methods here. It works, but it's not very pretty.  I have a few thought and I'm interested in your view on these:

1. These are `plot_*` methods. It seems to me these kinds of methods should only *visualize* results and not recompute anything. In other words, the tmin and tmax values should only work on the plotting functionality e.g. setting xlims through matplotlib. However, the way these changes are implented now, the `tmin` and `tmax` values are fed back to various computing methods which recompute some results which are then plotted. Notably in:
    a. Model.simulate
    b. Model.residuals
    c. Model.noise
    d. Model.get_contributions
I __think__ these recomputations are only aesthetical, but the behavior is very obscure. Also, from an architecture standpoint, it's inconsistent: through these `plot_*` methods you give access to the above mentioned `Model` methods, but you're being selective about which kwarg can be changed and which can't. Would it be okay to change the other kwargs like `name`, `freq`, `warmup` etc too? And what would happen if in the future we were to make some changes to the `Model.simulate` method that involve `tmin` and `tmax`, could that lead to unexpected behavior when a using the `plot_*` methods?

2. If what we want ís in fact recomputing results and then plot those, then I think these `tmin` and `tmax` kwargs should not be input to the `plot_*` functions, but in fact be input on a higher level. For example as a `CompareModels` or better even, a `Model` class attribute. 

My suggestion would be to use the `tmin` and `tmax` kwargs only on the plotting level, e.g. use them to "zoom in and out". So `tmin` and `tmax` would only be used to set `xlims` in matplotlib. If this does not generate the same results as are intended in #868 , then I strongly believe the changes shouldn't be made to the `plot_*` methods, but instead on a higher level.

dbrakenhoff:
Tough question... The whole design of CompareModels is so hideous (sorry about that) because it wants to be too flexible.

A lot of `ml.plots.*` methods take tmin/tmax to compute time series between tmin/tmax. 

So maybe you would expect `ps.CompareModels([ml0, ml1]).plot(tmin="2000")` to do the same. On the other hand, it would also be nice if you could just set tmin/tmax in CompareModels and not have to think about any of the plotting after that: `ps.CompareModels([ml0, ml1], tmin="2000").plot_contributions()`.

So in keeping with the hideous design: maybe both :scream:? And then when both are specified the one passed to the method wins?

And this is a good summary of the whole class really :sweat: :

> It works, but it's not very pretty.

AnneDeGraaf:
I like @martinvonk 's idea of setting tmin tmax as `CompareModels` attributes, and then passing them directly to `plots.compare` like so: `ps.plots.compare([ml1, ml2], tmin=tmin, tmax=tmax)`.

Not sure I agree with your suggestion @dbrakenhoff to allow the attributes to be changed through the `CompareModels.plots` method as well, that doesn't seem like good practice. I think the `ml.plots.*` functions are different here, as they create a brand new object instead of changing an existing one. 

However, after all, I still wonder why we can't just set the xlims on the plots instead of recomputing a bunch of things? When I visually compare the results of a CompareModels plots wíth tmin and tmax set, to a CompareModels plots with the original extent, it seems there aren't any differences in the plotted values?

If there áre subtle differences in the plotted values, I really wonder why we wouldn't give access to all the (kw)args of the `Model.simulate`, `.residuals`, `.noise` and `.get_contributions` methods. So this question I also posed above:
> Would it be okay to change the other kwargs like name, freq, warmup etc too?
@raoulcollenteur when we spoke about this during the Pastas days you said this wasn't possible because it would change results in a different way. But when you change the `tmin` and `tmax` kwargs there are also recomputations taking place. So what is really the difference then?
Also, will the user understand what happens when they set `tmin` and `tmax`? Will they just think they are zooming in on the plots, or will they understand that recalculations are taking place on the simulation, residuals, noise and contributions? 

Sorry to keep nagging so much about these seemingly small changes. I personally find it interesting and worth the time, but if you all have a different priority and prefer I just ship these changes asap let me know, I would totally understand.

dbrakenhoff:
> I like @martinvonk 's idea of setting tmin tmax as CompareModels attributes, and then passing them directly to plots.compare like so: ps.plots.compare([ml1, ml2], tmin=tmin, tmax=tmax).

I definitely agree with this, so I think this can be implemented. 

My additional question is whether users might expect the `CompareModels.plot_*` methods to also accept `tmin`/`tmax` since `Model.plots.*` methods mostly accept them as well, allowing you to quickly modify a plot. So very much reasoned from the user's perspective, and not so much the design of the class. 

Having said that, I think 95% of people will use `ps.plots.compare([ml1, ml2], tmin=tmin, tmax=tmax)`. So I guess once you start using the `CompareModels` class individual methods, those "power"-users can figure out that tmin/tmax are passed to class here and not the plot methods themselves. Also, the code is a lot cleaner and simpler this way. So I can live with that too :). 

 
> @raoulcollenteur when we spoke about this during the Pastas days you said this wasn't possible because it would change results in a different way. But when you change the tmin and tmax kwargs there are also recomputations taking place. So what is really the difference then?

The `ml.simulate` method is quite complex, because it also contains a lot of logic to calibrate your model using a different timestep, different tmin/tmax, etc. For example, it allows you to do `ml.solve()` and then `ml.solve(freq="7D", tmin="2000")`. All that logic is built into the `simulate` methods, but it does not make sense to use those options when you're only inspecting the result. We just happen to use the same method to plot the result. 

So in the plotting case, only modifying tmin/tmax makes sense I think. And that should generally give you the same result, even though it does perform a recomputation. The only time this causes issues is when the response time is longer than the warmup period (for more info, see #784).

> However, after all, I still wonder why we can't just set the xlims on the plots instead of recomputing a bunch of things?

Good question, I don't have a very good answer, but I guess because the rest of pastas does it this way... :innocent: 

> Also, will the user understand what happens when they set tmin and tmax? Will they just think they are zooming in on the plots, or will they understand that recalculations are taking place on the simulation, residuals, noise and contributions?

I think a lot of users probably don't realize there is actually a recomputation going on when they modify tmin/tmax  in the plot functions. This isn't that big of an issue I think, and it won't affect most users, but it can be confusing when it does matter... (#782).

AnneDeGraaf:
Hmm okay. Then I would like to suggest the following: let's make `tmin` `tmax` only act by setting the xlims for the plots. Nothing is recalculated (saving compute) and there is no unexpected behaviour due to #784 and #782. 

> Good question, I don't have a very good answer, but I guess because the rest of pastas does it this way... 😇

That can be changed, and is perhaps a better way forward 🙃 

I'll update the PR later this week, incorporating these changes.
Changing this PR to Draft for now.

martinvonk:
> My additional question is whether users might expect the `CompareModels.plot_*` methods to also accept `tmin`/`tmax` since `Model.plots.*` methods mostly accept them as well, allowing you to quickly modify a plot. So very much reasoned from the user's perspective, and not so much the design of the class.

I agree with @dbrakenhoff.  So I'd go for parsing `tmin` and `tmax` as arguments to `ps.plots.compare`. So `ps.plots.compare([ml1, ml2], tmin=tmin, tmax=tmax)` and only as attributes to the `CompareModels` class. Handling the class is intended for advanced users who know what they are doing with mosaics and everything. The `ps.plots.compare` is for people who want a quick figure and "familiar" Pastas behavior.
Only thing we could consider is letting the `CompareModels.plot` method have `tmin` and `tmax` arguments that overwrite the `tmin` and `tmax` attributes.

I would not only use `tmin` and `tmax` for only setting the xlims. This is already done sideways in the `CompareModels.plot` method with the call to `CompareModels.get_tminmax`. Changing the xlims can easily be done afterwards by the user as well. 

Recomputing everything all the time for `ml.simulate` and `ml.residuals` etc. is a separate issue. Maybe we can do something smart with caching.

dbrakenhoff:
@AnneDeGraaf, we talked about your question about the tmin/tmax doing a recompute, and eventually figured out this is intentional. I quoted your question in discussion #914 and provided a brief answer for our future selves and anyone else wondering why it was designed this way.

AnneDeGraaf:
great discussion folks, thanks. I processed the changes, ready for review now.

dbrakenhoff (approved):
Looks good to me! Thanks for picking this up, and helping us remember why pastas was written this way :P.

pastas/plotting/modelcompare.py
raoulcollenteur: Such improvements are always welcome! :)

raoulcollenteur (approved):

raoulcollenteur:
Thanks for your work in this issue @AnneDeGraaf ! What started as a 'good-first-issue' turned out to be a bit more challenging but the result is very good :) Looking forward to the next Pastas course where we can finally do `ps.plots.compare([ml1, ml2], tmax="2024")` 

I will merge this into dev after the test are run.

Cheers,
Raoul

pastas/plotting/modelcompare.py
```diff
@@ -907,7 +931,7 @@ def plot(
         grid: bool = True,
         legend: bool = True,
         adjust_height: bool = False,
-        legend_kwargs: Optional[dict] = None,
+        legend_kwargs: Optional[dict] = {},
```
martinvonk, line 934: I think this should be `None` as default argumnt, not `dict` right? Otherwise ruff and/or codacy will start complaining about mutable default arguments.

raoulcollenteur: Oui. good point. And add the if statement again.

raoulcollenteur (approved):

Issue #917, opened 2025-03-05
modify CCF computation to match output of numpy/scipy correlate functions


dbrakenhoff:
### Describe the task
Currently, to obtain the same results, the order of series must be flipped in pastas to match the output in numpy/scipy:

```python
np.correlate(a, b) 
ps.stats.ccf(b, a)
```

So we want to modify pastas to match the results of numpy/scipy. Discussion in #896. 

### Issue checklist
- [x] Add labels
- [x] Add issue type
- [x] Assign yourself, a Pastas developer or no one
- [x] Check if there is not already an existing issue for this task

Issue #915, opened 2025-03-05
irregular timestep CCF rectangle method returns the mean CCF


dbrakenhoff:
### Describe the bug
For irregular timesteps the CCF is computed in such a way that the forward lag and backward lag are added together and averaged. This does not affect the gaussian method since it doesn't take the absolute value of the time lag between x[i] and y[j]. 

![Image](https://github.com/user-attachments/assets/31caa057-860b-42ed-b647-ab6b50e4dca4)

### Issue checklist
- [x] Add labels
- [x] Add issue type
- [x] Assign yourself, a Pastas developer or no one
- [x] Check if there is not already an existing issue for this bug

Issue #912, opened 2025-03-04
[ENHANCEMENT] return lag 0 for ccf function


dbrakenhoff:
**Describe the proposed enhancement**
Return lag 0 by default

**Additional context**
See discussion in #896.

Issue #846, opened 2024-12-17
[ENHANCEMENT] ACF 'rectangle' parallelization


AviadAvr:
Currently, `_compute_ccf_rectangle()` works by utilizing nested for-loops for comparing the difference in lags to the `bin_width`. This approach works well but is substantially slower compared to the `regular` and `gaussian` methods, and especially when compared to FFT-based methods, which is understandable due to the constraint of having irregular timeseries.

A suggested improvement would be to compute each lag as a parallel process using `parallel=True` as part of the Numba `@njit` decorator.

Instead of:
```
@njit
def _compute_ccf_rectangle(
    lags: ArrayLike,
    t_x: ArrayLike,
    x: ArrayLike,
    t_y: ArrayLike,
    y: ArrayLike,
    bin_width: float = 0.5,
) -> Tuple[ArrayLike, ArrayLike]:
    """Internal numba-optimized method to compute the ccf."""
    c = empty_like(lags)
    b = empty_like(lags)
    n = len(t_x)

    for k in range(len(lags)):
        cl = 0.0
        b_sum = 0.0
        for i in range(n):
            for j in range(n):
                d = abs(t_x[i] - t_y[j]) - lags[k]
                if abs(d) <= bin_width:
                    cl += x[i] * y[j]
                    b_sum += 1
        if b_sum == 0.0:
            c[k] = nan
            b[k] = 0.01  # Prevent division by zero error
        else:
            c[k] = cl / b_sum
            b[k] = b_sum / 2  # divide by 2 because we over count in for-loop
    return c, b
```

Replace with:
```
from numba import prange

@njit(parallel=True) # Use Numba's parallel processing
def _compute_ccf_rectangle(
    """ no changes here"""

    for k in prange(len(lags)): # Use prange for parallel loops in Numba
        cl = 0.0
        b_sum = 0.0
        for i in range(n):
            """ rest of the code, no changes here"""
    return c, b
```

This could use a bit more testing, but from some benchmarking i've noticed an improvement by an order of magnitude or so. 
*This would of course vary based on the CPU used.

martinvonk:
Hi @AviadAvr ,

Thank you for making this issue. Can you provide a small example script where you run the benchmark and test the results? This way we can check the speedup.

We have been discussing the topic of parallelization with the Pastas development team lately. We have implemented some parallel methods in Pastastore where parallelization is built around models. Also implementing parallelization within models could cause some nasty conflicts so we have to think about how large the benefits are parallelization within models..

[...]

PR #918, opened 2025-03-05
CCF improvements

dbrakenhoff:
This PR introduces some breaking changes, and I'm not sure how to deal with that:
 - Previous code `ps.stats.ccf(a, b)` is now calculated as `ps.stats.ccf(b, a)`. Do we add a warning?
 - The 0 lag for the ccf is returned. I think it's okay to just return the new result. 

The ACF is unaffected. One slight issue is that I'm getting ACFs > 1.0 warnings in the autocorrelation notebook. The max exceedance is <1e-3 so it's a computational issue having to do with the binning methods, as the current warning suggests. Not sure what to do about that...  

I don't think many people will be affected by these breaking changes since I don't think many people were actively using the CCF except for computing the autocorrelation, which is unchanged. So I think it would be okay to include these changes, and perhaps include a warning for flipped order of x and y.

# Short description
Deals with the following issues:
- #846 
- #912 
- #915 
- #917

### Lag 0 for CCF
Lag 0 for the CCF is now returned. In `ps.stats.acf()` this value is dropped if lags is passed as an integer. 

### Flip order of x, y to match numpy/scipy.correlate

The results now match numpy/scipy: `ps.stats.ccf(a, b)` is similar to `np.correlate(a, b, mode="full")[len(a) - 1:]`

### Performance
Improves performance considerably:
- Traverse only the lower diagonal of the dot product between x and y, in other words only consider the lags in one direction.
- In `_compute_ccf_gaussian`, check whether abs(time lag) is smaller than 6 *  std.dev. Beyond that limit the computed weight becomes negligible. See comment below.
- Use `numba.prange` for the outer loop
- Set `parallel=True, nogil=True, cache=True` in `@njit` decorator.
- Minor performance improvements (e.g. array access once per loop, etc.).

Many thanks @AviadAvr for suggesting the implementation of `numba.prange`!

Timings (N=1000, lags=365):
For `_compute_ccf_regular()`
```
Original       : 26.2 ms ± 1.6 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
This PR        : 4.25 ms ± 5.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)  # <-- I reverted this change
Pastas plugins : 993 µs ± 133 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```
EDIT: unfortunately, `njit`-ing the `_compute_ccf_regular` made it rather flaky. I was getting significant speedups, but the `test_acf_hourly` went from very fast to ~15s. So I reverted that change. For very fast CCFS for time series with regular timesteps `pastas_plugins` is recommended. 

For `_compute_ccf_gaussian()`
```
Original       : 4.51 s ± 69.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
This PR        : 559 ms ± 16.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

For `_compute_ccf_rectangle()`
```
Original       : 248 ms ± 1.85 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
This PR        : 56.1 ms ± 1.86 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

- [x] closes issues
   - [x] #846 
   - [x] #912 
   - [x] #915 
   - [x] #917
- [x] is documented
- [x] type hints
- [x] tests added or expanded

### DIFF
--- a/pastas/decorators.py
+++ b/pastas/decorators.py
@@ -146,7 +146,7 @@ def deprecate_args_or_kwargs(
         raise DeprecationWarning(msg)
 
 
-def njit(function: Optional[Function] = None, parallel: bool = False) -> Function:
+def njit(function: Optional[Function] = None, **kwargs) -> Function:
     def njit_decorator(f: Function):
         try:
             if not USE_NUMBA:
@@ -154,7 +154,7 @@ def njit(function: Optional[Function] = None, parallel: bool = False) -> Functio
             else:
                 from numba import njit
 
-                fnumba = njit(f, parallel=parallel)
+                fnumba = njit(f, **kwargs)
                 return fnumba
         except ImportError:
             return f
--- a/pastas/stats/core.py
+++ b/pastas/stats/core.py
@@ -223,7 +223,7 @@ def ccf(
     elif bin_method == "gaussian":
         c, b = _compute_ccf_gaussian(lags, t_x, x, t_y, y, bin_width)
     elif bin_method == "regular":
-        c, b = _compute_ccf_regular(arange(1.0, len(lags) + 1), x, y)
+        c, b = _compute_ccf_regular(arange(0, len(lags), dtype=float), x, y)
     else:
         raise NotImplementedError
 
--- a/pastas/stats/core.py
+++ b/pastas/stats/core.py
@@ -244,7 +244,10 @@ def ccf(
     result = result.where(result.n > min_obs).dropna()
 
     # Raise a warning if the correlation is above 1 or below -1
-    if (result.ccf > 1).any() or (result.ccf < -1).any():
+    # NOTE: Using 1.01 to avoid excessive warnings when using binning methods for
+    # irregular timesteps. In those cases correlations sometimes exceed 1 by a small
+    # amount.
+    if (result.ccf.abs() > 1.01).any():
         msg = (
             "The correlation is above 1 or below -1. This can occur due to the "
             "binning method used. Please check the data and the binning method and "
--- a/pastas/stats/core.py
+++ b/pastas/stats/core.py
@@ -348,13 +348,12 @@ def _compute_ccf_gaussian(
     return c, b
 
 
-@njit(parallel=False, nogil=False, cache=False)
 def _compute_ccf_regular(
     lags: ArrayLike, x: ArrayLike, y: ArrayLike
 ) -> Tuple[ArrayLike, ArrayLike]:
     c = empty_like(lags)
     n = len(x)
-    for i in prange(len(lags)):
+    for i in range(len(lags)):
         lag = int(lags[i])
         if lag < n:
             # flip x, y to match numpy/scipy correlate output order
--- a/pastas/stats/core.py
+++ b/pastas/stats/core.py
@@ -117,7 +117,10 @@ def acf(
         full_output=full_output,
         alpha=alpha,
         fallback_bin_method=fallback_bin_method,
-    ).iloc[1:]  # drop first value, which is always 1
+    )
+    # drop value for lag=0 by default, unless explicitly included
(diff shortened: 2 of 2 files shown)


raoulcollenteur (approved):

raoulcollenteur:
Thanks for this PR, looks nice! I think we can safely reverse the order for the CCF to match scipy/numpy, I don't think many/any people are using this function. And if so, they probably know what they are doing and will note changes that can be tracked back to this PR.

Regarding the use of `numba.prange` (here, but also elsewhere), I wanted to implement this a while back but @martinvonk  mentioned that this could cause problems when running multiple pastas models in parallel (i.e., with pastastore). Does this mean the use of `prange` does not cause any issues and we can implement it in other parts of Pastas too? 

Perhaps we should only raise the warning when acf > |1.01| or somewhere around that? 

Good to go I think if prange is safe to use now :)

dbrakenhoff:
> Regarding the use of numba.prange (here, but also elsewhere), I wanted to implement this a while back but @martinvonk mentioned that this could cause problems when running multiple pastas models in parallel (i.e., with pastastore). Does this mean the use of prange does not cause any issues and we can implement it in other parts of Pastas too?

Pastastore doesnt´t currently provide any direct methods for computing the ACF/CCF in parallel. So implementing it here isn't much of a risk. Additionally, I just found out that `numba.set_num_threads(1)` can be used to make parallel jitted functions run in a single thread. We might have to explicitly start using that in pastastore if it turns out that numba parallel affects the naive parallelization implemented in pastastore, but I think it's okay to start implementing it. 

> Perhaps we should only raise the warning when acf > |1.01| or somewhere around that?

Yeah, agreed, I'll modify the threshold.

martinvonk:
I think, before making stuff parallel in Pastas we should first make a discussion / issue on how and what. Think this through a bit. 

Because maybe we want PastaStore non-parallel and Pastas in parallel or the other way round?
Maybe the default for Pastas should be to run everything not in parallel? Add something to the rcParams["num_cores"] = 1?
Do we need if `__name__ == "__main__"` statements to ensure memory safety? 
How about linux/macos/windows differences for parallel?

dbrakenhoff:
> Because maybe we want PastaStore non-parallel and Pastas in parallel or the other way round?

It'll probably end up as a bit of both. As long as it's easy to configure, i.e. turn parallel off in both modules, it should be simple to accommodate both approaches. And so far, that seems simple to do. 

> Maybe the default for Pastas should be to run everything not in parallel?

I would say the opposite, pastas tries parallel, unless you tell it not to. If pastastore parallel is affected by this, pastastore can turn off pastas parallel.

> Do we need if __name__ == "__main__" statements to ensure memory safety?

For numba prange we don't need this.

> How about linux/macos/windows differences for parallel?

Numba takes care of this.


The above questions are valid if we start implementing multiprocessing or multithreading within pastas, but as long as we remain in the numba parallel space, these things are taken care of by numba and we don't really need to worry. 

So I would say, go ahead on the numba front, but if we want to start implementing other forms of parallel computing, we shoud, definitely revisit these questions.

martinvonk:
But can we then do something like this in this PR:
`@njit(parallel=ps.rcParams["parallel"])`
and 
```python
if ps.rcParams["parallel"]:`
    from numba import prange as range
```

dbrakenhoff:
I think it would make sense to include a pastas parallel setting so it can be turned off on the module level. Unfortunately this only works on import, so you can never configure this value after importing pastas. So you would have to modify the rcparams.py file manually, which is a nice option to have, but doesn't allow interactive configuration.

The solution is to use `numba.set_num_threads(1)` if you want to toggle between serial/parallel execution using numba after importing pastas. 

Also `numba.prange` defaults to `range` when parallel is off.

Issue #893, opened 2025-02-24
Plots results plots interpolated residuals series


raoulcollenteur:
**Describe the bug**
Do not draw an interpolated line when there is a big gap..

**To Reproduce**
`ml.plots.results()`

**Screenshots**

![Image](https://github.com/user-attachments/assets/3da69eff-b4a6-4d1c-ac88-e4919f6fe105)

**Python package version**

Pastas version: 1.8.1
Python version: 3.10.12
NumPy version: 1.26.4
Pandas version: 2.2.3
SciPy version: 1.15.1
Matplotlib version: 3.10.0
Numba version: 0.60.0

mbakker7:
I presume you mean the interpolated line in the residuals, correct? That means we have to define what a 'big' gap is. Any thoughts?

raoulcollenteur:
@martclanor let me know if you want to pick this up and I can assign you to this issue :)

martinvonk:
We could also do something like this. Where we cut the line if the gap is larger than 1 day. Or any other Timedelta; maybe the model frequency? 
```python
res_split = (residuals.index.diff() > pd.Timedelta(days=1)).cumsum()
residuals.groupby(head_split).plot()
```

martinvonk:
Code snippet what this will look like:
```python
# %%
# load packages
import pandas as pd

import pastas as ps

# create model
data = ps.load_dataset("collenteur_2023")
head = data["heads"]["Davos"].dropna()
ml = ps.Model(head)
sm = ps.RechargeModel(
    data["precipitation"]["Davos"],
    data["evaporation"]["Davos"],
    temp=data["temperature"]["Davos"],
    rfunc=ps.Gamma(),
    name="rch",
    recharge=ps.rch.FlexModel(snow=True),
)
ml.add_stressmodel(sm)
ml.solve()

# plot residuals with gap if gap is larger than model frequency
residuals = ml.residuals()
residuals_split = (
    residuals.index.diff() > pd.Timedelta(1, unit=ml.settings["freq"])
).cumsum()
residuals.groupby(residuals_split).plot(figsize=(7, 3), grid=True, color="k")
```

![Image](https://github.com/user-attachments/assets/1d931769-e735-4fe7-94e6-ba27cd87dcc9)

martclanor:
[...]

PR #925, opened 2025-03-07
Re-add dropped timestamps before plotting residuals in ml.plots.results()

martclanor:
# Short description
Iiuc, the issue is that the residual series lacks "gaps" (timestamps with missing values) because they were dropped from the related `TimeSeries` object.

The proposed solution re-inserts dropped timestamps with `NaN` values, effectively preventing interpolation during plotting.

- [ ] closes issue #893
- [x] is documented
- [x] type hints _(N/A)_
- [ ] tests added or expanded
- [x] example notebook (for new features) _(N/A)_


mbakker7:
Thanks for trying to contribute to pastas!
I am afraid this does not solve the problem described in #893 
The interpolated line is drawn because there are no measurements, and hence no residuals can be computed.
If I am misinterpreting what your solution is, please add an example where this works.

martclanor:
Hi @mbakker7, I might not have explained clearly the solution that I tried here. But first, just to elaborate a bit further on that:

When a `ps.Model` object is instantiated with the observation heads (e.g. with nan-values), nan-values are dropped and thus `ml.oseries.series` does not include the timestamps with the nan-values anymore. And since `ml.residuals()` is calculated from `ml.oseries.series`, residuals series also does not include those timestamps.

My intention in the solution is to re-insert these dropped timestamps, adding them with nan-values into `ml.residuals()` based on timestamps from the original observation series (`ml.oseries.series_original`). This results into a residual series that can be plotted without the interpolation.

That approach sort of works, for example, in `doc/examples/example_snow.py` which can be tested by adding this line:
```python
head.loc["1996-01-03":"2007-12-31"] = np.nan 
```
Before: `ml.plots.results()`
pastas-issue893-before

After: `ml.plots.results()`
pastas-issue893-after

HOWEVER, as I test more cases, I see that this does not always work specifically when the "dropping of nan-values" part is not performed, for example in:
- `doc/examples/example_tarso.py` where the `ml` object is loaded
- `doc/examples/example_well.py` where the raw csv file excludes already the missing timestamps (instead of having nan values)
- any other example with irregular frequency timesteps

I'm not sure just yet how to fix this from here and I don't have a better idea atm (apart from just plotting the residuals as markers instead of line 😅). Please feel free to close the PR if this seems like a dead-end. Thanks!

raoulcollenteur:
Hi @martclanor ,

Thanks for helping! I was actually also looking into this but didn't add the changes yet as it needs some testing. Here's my solution that I was working on (L235 modelplots.py):

```
        res.plot(ax=ax2, color="k", x_compat=True)
        # create nans to not plot periods without residuals
        dt = res.index.to_series().diff().min()
        res = res.asfreq(dt)
        res.plot(
            ax=ax2,
            color="k",
            x_compat=True,
        )
        if self.ml.settings["noise"] and self.ml.noisemodel:
            noise = self.ml.noise(tmin=tmin, tmax=tmax)
            noise = noise.asfreq(dt)
```

If you want, perhaps you can test and improve this solution and adapt your PR accordingly?

Particularly getting the dt seems challenging, but perhaps this work already...

Cheers,
Raoul

martclanor:
@martinvonk, I modified a bit the implementation to classify the gap based on the index range instead of series frequency because with the latter, I'm still getting "unplotted lines" on some examples.

I still get one failing test locally (windows) but it seems unrelated and I'm guessing it won't be caught in the CI here.

martinvonk:
> @martinvonk, I modified a bit the implementation to classify the gap based on the index range instead of series frequency because with the latter, I'm still getting "unplotted lines" on some examples.
> 
> I still get one failing test locally (windows) but it seems unrelated and I'm guessing it won't be caught in the CI here.

Thanks for working on this. I don't know if a percentage of tmax minus tmin as an indicator is the way to go still workable. Also, I'd prefer the if gap to the `plot_series_with_gap` is a Timedelta, not a float. That keeps the function generic and makes it clear what it's intended behaviour is.

martinvonk:
> I understand, it does seem preferable indeed. I first tried continuing with that approach but I'm a bit stumped on how to determine the number of `Timedelta` units that can be considered a "gap" especially with irregular frequency series. I couldn't come up with a robust enough solution for this so I went for the "x% (tmax-tmin)" solution instead.
> 
> If you can help further on this, I'd be glad to test the solution.

I was thinking of testing;
`gap = max(series.index.diff().quantile(0.95), 100)`

martinvonk:
I made some commits. The gap is now plotted for the 95'th quantile of the differences, or 50 if the gap is larger than fifty. If only one point remains, a scatter plot is made of that point. So you'll always see the a line/point for residual.

martclanor:
I think this is great already especially with the flexibility that it offers. I've added a couple of commits to address minor issues I encountered while testing.

raoulcollenteur (approved):

Issue #932, opened 2025-03-12
Plot two responses for TarsoModel and ChangeModel


martinvonk:
### Describe the proposed feature
Both the TarsoModel and ChangeModel have two response functions. For the TarsoModel only one response is plotted in the results plot. For the ChangeModel neither response is plotted. 

### Expected usage
Plot both responses in the rfunc plot (of plot results and other plots).

Issue #933, opened 2025-03-13
Update of the model results plot


martinvonk:
Some things that were bothering me about the results plot:
- yticks sometimes not clearly visible for paramter plots
- parameters sometimes not clearly readable due to length of the parameter names (or length of the parameter values)
- `Step response` or `Block response` plotted as a title, not a label
- matplotlib gridspec is awfull to debug compared to a `subplots_mosaic`
- difficult to update axes afterwards since axes are difficult to identify (for instance to update the xlabels or ylabels)
- ylims set on the min and max value instead of a buffer around the limits (as Matplotlibs default).
- #932 

### Describe the proposed feature
Address the above issues with an update results plot via the subplots_mosaic

PR #934, opened 2025-03-13
Update model results plot with mosaic and double response for specific stressmodels

martinvonk:
## Short description
Update on the ml.plots.results() figure.
- create a new `results_mosaic` method that gives a similar result as the original results plot but adresses #933 
- leaves the `results` method untouched except for a few efficiency changes (which only partially worked https://github.com/pastas/pastas/pull/934/commits/197e518e3c931a3d248e2dd56f4ae36f648aff4e)
- make the method `_plot_parameters_table` to nicel
- update the `_plot_response_in_results` such that `GridSpec` is no longer needed
- Note that the results_mosaic does not support `split`. That gave me too much of a headache. I did update the split method in the original results plot though to save some lines of code. If we want the `results_mosaic` to replace the `results` method than we should consider adding the `split` method to ensure backwards compatibility.
## Test code
```python
import pastas as ps
data = ps.load_dataset("collenteur_2023")
ml = ps.Model(data["heads"]["Davos"].dropna())
rm = ps.RechargeModel(
    data["precipitation"]["Davos"].copy().rename("davos_rain"),
    data["evaporation"]["Davos"].copy().rename("davos_evap"),
    ps.Gamma(),
    temp=data["temperature"]["Davos"].copy().rename("davos_temp"),
    recharge=ps.rch.FlexModel(snow=True),
)
ml.add_stressmodel([rm])
# ml.add_noisemodel(ps.ArNoiseModel())
ml.solve()
# axes = ml.plots.results(stderr=True)
axd = ml.plots.results_mosaic(stderr=True)
```
### `ml.plots.results_mosaic`
image

### `ml.plots.results`
image


martinvonk:
While I was at it I also fixed #932

dbrakenhoff:
> While I was at it I also fixed https://github.com/pastas/pastas/issues/932

Nice :), but won't users get a bit confused by the different lengths responses? Not sure if that's something we need to address but it is something that I saw immediately in the new figure.

rubencalje (approved):

martinvonk:
> > While I was at it I also fixed #932
> 
> Nice :), but won't users get a bit confused by the different lengths responses? Not sure if that's something we need to address but it is something that I saw immediately in the new figure.

Good question. I don't know. I figure that if people use TarsoModel or ChangeModel to some degree know what they are doing. So they know that the stressmodel has two responses. Thus only showing one response, when you actually have two might be more confusing than showing two lines?

dbrakenhoff:
Ah my bad, I thought this was linear recharge on first glance, I agree for Tarso and ChangeModel users don't necessarily have an expectation on that front, and I agree this is better than hiding the second response.

Issue #954, opened 2025-06-12
Erroneous warning that tmax > warmup


mbakker7:
In the current version of Pastas (1.9.0), the solver report gives an incorrect warning that the tmax > warmup. 
Check out, for example, the basic notebook A Basic Model on the readthedocs page: https://pastas.readthedocs.io/stable/examples/basic_model.html

PR #956, opened 2025-06-12
Fix warmup check

dbrakenhoff:
# Short description
Add a short description describing the pull request here.

- [x] closes issue #954 
- [x] is documented
- [x] type hints
- [x] tests added or expanded

### DIFF
--- a/pastas/check.py
+++ b/pastas/check.py
@@ -3,7 +3,7 @@ from typing import Callable, Optional
 
 import numpy as np
 from matplotlib.colors import rgb2hex
-from pandas import DataFrame, concat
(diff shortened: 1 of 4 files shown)


mbakker7:
The warning is indeed gone on `basic_notebook.ipynb`.

Just to check if it works correctly now: Is there a test (or example) somewhere that plots the warning when tmax is indeed longer than warmup?

mbakker7:
Test looks good. Just out of curiosity. Why
```
assert not check.loc["rch", "check_warmup"].item()
```
and not just
```
assert not check.loc["rch", "check_warmup"]
```

raoulcollenteur (approved):

PR #962, opened 2025-07-08
Pastas uncertainty notebook fixes

mbakker7:
This PR fixes the pastas_uncertainty_notebook

The notebook had a few errors and (inadvertently) some things that were tried, including parallel computing. As a result, the notebook didn't run.  
The notebook was updated and the results are now also better as a newer version of least_squares was used. 
Type hints were removed from functions in the notebook, as it makes little sense to add those to temporary functions in a notebook while they do make them more difficult to read.


martinvonk:
If we don't want't to test this notebook / wait for the CI to complete after 30 minutes, then we should remove this notebook from the `test_notebooks_bench.py`

Issue #749, opened 2024-04-24
Possible different in results between different versions of Python packages


raoulcollenteur:
**Describe the bug**
During the recent Pastas workshop it seemed that different results were obtained between different computers and version of Python packages, but we could not test this systematically. I'd be happy if others can run the notebook I created and report here in this issue the values that are printed in the notebook at the end.

**To Reproduce**
The the notebook in the zip-file.

[test_case.zip](https://github.com/pastas/pastas/files/15091415/test_case.zip)

And report the values here

**Expected behavior**

This is what I get on a Mac M2:

**EVP= 82.74855784337282**
Python version: 3.11.8
NumPy version: 1.26.4
Pandas version: 2.1.4
SciPy version: 1.13.0
Matplotlib version: 3.8.3
Numba version: 0.59.0
LMfit version: 1.2.2
Latexify version: Not Installed
Pastas version: 1.5.0

Here's what I get on Mac M1 with very old packages:

**EVP= 82.74835542311386**
Python version: 3.9.12
NumPy version: 1.21.6
Pandas version: 1.4.2
SciPy version: 1.11.4
Matplotlib version: 3.5.1
Numba version: 0.55.1
LMfit version: 1.0.3
Latexify version: Not Installed
Pastas version: 1.5.0

mbakker7:
EVP= 82.74835371251245
Python version: 3.11.4
NumPy version: 1.24.3
Pandas version: 2.1.1
SciPy version: 1.10.1
Matplotlib version: 3.7.1
Numba version: 0.57.0
LMfit version: 1.2.2
Latexify version: Not Installed
Pastas version: 1.5.0

raoulcollenteur:
The problem appears to be related to Scipy version 1.13.0, version 1.13.1 seems to have fixed it: https://github.com/scipy/scipy/issues/20531

[test_case_2.zip](https://github.com/user-attachments/files/15746532/test_case_2.zip)

The attached notebook does show large differences between 1.13.0 and 1.13.1 on my system.

dbrakenhoff:
Here is a tox.ini file that allows testing of multiple scipy/python versions. Youll need to expose different python versions to tox to get it to work for all configurations. But the most important thing is checking scipy. 

Run in parallel with `tox -p auto`

```config
[tox]
env_list = py{310,311,312}-scipy{1.12.0, 1.13.0, 1.13.1, 1.14.0rc1}
minversion = 4.15.1

[testenv]
description = run the tests with pytest
[...]

PR #967, opened 2025-08-07
Fix NumPy version higher than 2.0

martinvonk:
# Short description
To prevent different outcomes in pastas solving

- [x] closes issue #749


mbakker7 (approved):
I think @martinvonk has convincingly shown that variations in NumPy prior to 2.0 gave differences in optimization results. Those differences have disappeared for Numpy >= 2.0. So I am in favor of making this change. It prepares us well for the future, even though we should keep monitoring consistency in optimization. Regarding that last point. some research into #982 may be worth while.

Issue #935, opened 2025-03-13
Update type hints for python 3.10


martinvonk:
### Describe the task
We now use the type hints compatible with python <3.10. But after python 3.10 type hints got updated and made more readable. https://docs.python.org/3/whatsnew/3.10.html

### Additional context
```python
from typing import Union, Optional, Dict, List
a: List[int] = [1, 2, 3]
b: Union[int, None] = None
c: Optional[int] = None
d: Dict[str, float] = {"key": 1.0}
```
would become:
```python
a: list[int] = [1, 2, 3]
b: int | None = None
c: int | None = None
d: dict[str, float ]= {"key": 1.0}
```

PR #970, opened 2025-08-11
Type Hints Update for Python 3.10 (PEP 585 & 604)

martinvonk:
# Short description
- [PEP 604](https://peps.python.org/pep-0604/): New Type [Union Operator: |](https://docs.python.org/3/whatsnew/3.10.html#pep-604-new-type-union-operator)
- [PEP 585](https://peps.python.org/pep-0585/): Support for subscripting ([]) lists, dicts and tuples 
- [x] closes issue #935


pastas/check.py
```diff
@@ -82,7 +82,7 @@ def _response_memory(
     threshold: float,
     label: str,
     cutoff: float = 0.95,
-    names: Optional[list[str] | str] = None,
```
raoulcollenteur, line 85: So no need to add Optional anymore right? This is recognized automatically?

raoulcollenteur (approved):

martinvonk:
Accidentily did merge instead of squash and merge..

Issue #884, opened 2025-02-07
ZeroDivision error when well series always zero


FransSchaars:
if all fluxes in a well are zero, `get_init_parameters` (in rfunc)
gives a ZeroDivision error in: ` 1 / self.gain_scale_factor`

Possible solutions: 
1) inform user he is doing something silly
2) remove stress
3) make sure there is a zero contribution

raoulcollenteur:
Thanks for the example! I think one solution might be to check whether or not the gain_scale_factor equals zero, and set the initial parameter as 1 in that case. Something like this:

`1 if self.gain_scale_factor is 0 else 1 / self.gain_scale_factor `

For all instances in rfunc. Want to implement this in the code and submit a PR @FransSchaars ?

dbrakenhoff:
Could we not set the initial guess to 0.0 if the variation in the series is 0? Or does that cause issues in the optimization?

FransSchaars:
It will also fail when the recharge is constant, because we are looking at the std, so raoul's solution would be better

dbrakenhoff:
> It will also fail when the recharge is constant, because we are looking at the std, so raoul's solution would be better

Good point, I wasn't thinking clearly ;)

PR #992, opened 2025-09-03
Make sure gain_scale_factor is not 0

raoulcollenteur:
by making sure the gain_scale_factor is not 0.0. This simple fix seems to work well :)

- [x] closes issue #884


martinvonk (approved):
Maybe we could add a little log (INFO/DEBUG?) message that the gain_scale_factor was too close to zero so we set it to one. And the user could do something themselves about their initial  `_A` value?

raoulcollenteur:
Thanks @martinvonk, I added a short warning message and will merge this after the checks have passed.

Issue #982, opened 2025-08-21
Should we compute our own Jacobian?


mbakker7:
Many (if not all) the least squares options in Pastas use the Jacobian to zoom in on the optimal parameters. 
Computation of the Jacobian is now done automatically by the solver (scipy or lmfit) using a numerical derivative. It is often better to use an analytic Jacobian (Menyanthes did that for some of the parameters), but that may be difficult for many parameters. I recently read a cool trick to compute an accurate derivative of a real function using a small complex step. This is very accurate and does not suffer (at all?) from finite machine accuracy. It may be worthwhile to use this trick to compute our own Jacobian and feed that to lmfit/least_squares. 

Before we do that, we may want to do some investigating and compare the Jacobian using the current lmfit/least_squares, an analytic derivative (when possible), and the complex trick that I am proposing.

martinvonk:
I think the complex trick you propose is also available in the `numdifftools`  library, https://numdifftools.readthedocs.io/en/latest/reference/generated/numdifftools.core.Jacobian.html

martinvonk:
I'm in favor of this change. SciPy also supports complex jacobians in it's `approx_derivative` function via method="cs". The `approx_derivative` function is also called by 'least_squares` internally. Se we could move this function `appprox_derivative` into Pastas, wrap it, and use it for our own purposes like so:

```python
def jacobian(
    fun: Callable[[ArrayLike], ArrayLike],
    x0: ArrayLike,
    method: Literal["2-point", "3-point", "cs"] = "2-point",
    rel_step=None,
    abs_step=None,
    bounds=(-np.inf, np.inf),
    args=(),
    kwargs=None,
) -> ArrayLike:
    """Compute finite difference approximation of the derivatives of a
    vector-valued function.

    If a function maps from R^n to R^m, its derivatives form m-by-n matrix
    called the Jacobian, where an element (i, j) is a partial derivative of
    f[i] with respect to x[j].

    From Scipy:
    https://github.com/scipy/scipy/blob/f5841beffea5036832652bfc92b2a1356b7d1007/scipy/optimize/_numdiff.py#L278-L617

    Parameters
    ----------
    fun : callable
        Function of which to estimate the derivatives. The argument x
        passed to this function is ndarray of shape (n,) (never a scalar
        even if n=1). It must return 1-D array_like of shape (m,) or a scalar.
    x0 : array_like of shape (n,)
        Point at which to estimate the derivatives.
[...]

PR #993, opened 2025-09-08
Compute our own jacobian

martinvonk:
# Short description
This PR adds the methodology to compute our own jacobian via the `ps.solver.jacobian` function. It calls the same methods as SciPy already did internally with `least_squares` so the outcomes are / should be the same.

- [x] closes issue #982 
- [x] is documented
- [x] type hints
- [x] tests added or expanded


martinvonk:
Can we change this codacy check @raoulcollenteur ?
```
MEDIUM
Code complexity

Method ... has 65 lines of code (limit is 50)
```
I think it is super arbitrary, and docstring counts as well.

martinvonk:
> If it is indeed possible to provide a callable Jacobian function to the solver, that would be my preference for now. We can test things out and don't have to change anything yet.

This is possible. But I think it might be useful to store these different Jacobian functions in one place then. Maybe Pastas-Plugins?

Issue #1003, opened 2025-09-15
`ml.stats.kge()` not working when freq='7D'


OnnoEbbens:
When I solve a model using `ml.solve(freq='7D')` and then compute the kge using `ml.stats.kge()` I get the error below. This is especially annoying because `ml.stats.summary()` won't work anymore because it computes the `kge` by default.

I got this error locally and in the `basic_model.ipynb` notebook.

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[20], [line 1](vscode-notebook-cell:?execution_count=20&line=1)
----> [1](vscode-notebook-cell:?execution_count=20&line=1) ml.stats.kge()

File ~\02_python\pastas\pastas\decorators.py:65, in model_tmin_tmax.<locals>._model_tmin_tmax(self, tmin, tmax, *args, **kwargs)
     62 if tmax is None:
     63     tmax = self.ml.settings["tmax"]
---> [65](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/oebbe/02_python/pastas/doc/examples/~/02_python/pastas/pastas/decorators.py:65) return function(self, tmin, tmax, *args, **kwargs)

File ~\02_python\pastas\pastas\modelstats.py:334, in Statistics.kge(self, tmin, tmax, weighted, modified, **kwargs)
    332 sim = self.ml.simulate(tmin=tmin, tmax=tmax)
    333 obs = self.ml.observations(tmin=tmin, tmax=tmax)
--> [334](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/oebbe/02_python/pastas/doc/examples/~/02_python/pastas/pastas/modelstats.py:334) return metrics.kge(
    335     obs=obs, sim=sim, weighted=weighted, modified=modified, **kwargs
    336 )

File ~\02_python\pastas\pastas\stats\metrics.py:683, in kge(obs, sim, missing, weighted, max_gap, modified)
    680     logger.warning("Time indices of the sim and obs don't match.")
    681     return nan
--> [683](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/oebbe/02_python/pastas/doc/examples/~/02_python/pastas/pastas/stats/metrics.py:683) r = pearsonr(obs=obs, sim=sim, weighted=weighted, max_gap=max_gap)
    685 mu_sim = mean(sim, weighted=weighted, max_gap=max_gap)
[...]

PR #1006, opened 2025-09-23
fix for #1003

OnnoEbbens:
When I solve a model with a frequency of `7D` -> `ml.solve(freq='7D')` and then compute the kge using `ml.kge()` I get an error. More info in this issue #1003.

The problem here is that the `kge` and `pearsonr` is required. The `pearsonr` is calculated from the observations and simulations. These are obtained using:
```
obs = self.ml.observations(tmin=tmin, tmax=tmax)
sim = self.ml.simulate(tmin=tmin, tmax=tmax)
```
Which works fine if the frequency is `D` but not for a frequency of `7D` because then `sim` and `obs` are different sizes. This same problem occurs when calculating residuals for models with a frequency of `7D`. So I think the solution is to do the same operation as for the residuals:
```
if self.interpolate_simulation:
    # interpolate simulation to times of observations
    sim_interpolated = np.interp(oseries_calib.index.asi8, sim.index.asi8, sim.values)
else:
    # All the observation indexes are in the simulation
    sim_interpolated = sim.reindex(oseries_calib.index)
```

Only the `kge` expects a series for sim so I modified this to:

```
# Get simulation at the correct indices
if self.ml.interpolate_simulation:
    # interpolate simulation to times of observations
    sim_interpolated = Series(interp(obs.index.asi8, sim.index.asi8, sim.values), index=obs.index)
else:
    # All the observation indexes are in the simulation
    sim_interpolated = sim.reindex(obs.index)
```

This seems to fix the problem although I think it is good of someone looks into this because I don't know if this is the best place to fix this.

### DIFF
--- a/pastas/modelstats.py
+++ b/pastas/modelstats.py
@@ -22,8 +22,8 @@ These methods may be used as follows.
     nse        0.929136
 """
 
-from numpy import nan
-from pandas import DataFrame
+from numpy import nan, interp
+from pandas import DataFrame, Series
 
 from pastas.typing import Model, TimestampType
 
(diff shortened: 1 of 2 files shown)


raoulcollenteur:
Looks like a good pragmatic solution to me! OI'll merge this into the dev branch now, thanks for the fix!

raoulcollenteur (approved):

Issue #961, opened 2025-06-28
Caching stressmodel contribution during Model.solve to make solving faster


HenrikSpa:
It struck me that it might be possible to make solving models faster by caching stressmodel contributions that are not calibrated (all parameter have vary=False). There might be something I'm missing though, which makes this not work.

How I interpret Model.solve to work:

1.  Model.solve calls solver.solve.
2.  The solver, calls solver.objfunction during the minimization process. objfunction is (i believe) called each time the minimize function tries new parameter values.
3.  solver.objfunction calls solver.misfit, which in turn calls Model.residuals.
4. Model.residuals calls Model.simulate, which in turns calls every Stressmodel.simulate. 

If I interpret this correctly, it means that for every minimize step, contribution for every stressmodel is calculated. However, some stressmodels are rather (relatively) heavy to simulate, like RechargeModel. If the RechargeModel is not calibrated by setting vary=False for every parameter, recalculating the contribution each step is wasteful. It would be enough to calculate the contribution just once and then reuse it for the rest of the minimize steps. 

So my suggestion (if I get the workings of the solving process right):

1. If vary=False for every parameter of a stressmodel, calculate the contribution of that stressmodel only the first time during the solve process and store the contribution within the model object.
2. Retrive the cached contribution for the rest of the steps during the whole solve process.

Any thoughts on this?

PR #1016, opened 2025-10-03
implement simulate caching

dbrakenhoff:
To test:
- set `max_cache_size` in StressModelBase in stressmodels.py (kwarg should be exposed to certain StressModels eventually). Currently `max_cache_size=32`.
- Compare solve times (in Notebook environment to allow usage of %timeit):

```python
ps.set_use_cache(False) # default is off, but to be sure
t0 = %timeit -o ml.solve(report=False)
ps.set_use_cache(True) # turn on caching
t1 = %timeit -o ml.solve(report=False)
(t1.average - t0.average) / (t0.average) * 100  # speedup %
```

# Short description
- add USE_CACHE global variable
- add method to set USE_CACHE `ps.set_use_cache()`
- add caching decorator that respeces USE_CACHE setting.
- add `._cache` attribute to `StressModelBase`
- rename `simulate()` to `_simulate()`  with hashable arguments (p-array is now tuple) and cache this method
- add simulate that converts param array to tuple.

- [x] closes issue #961 
- [x] is documented
- [x] type hints
- [x] tests added or expanded

### DIFF
--- a/.github/workflows/test_benchmark_notebooks.yml
+++ b/.github/workflows/test_benchmark_notebooks.yml
@@ -5,6 +5,13 @@ on:
     branches:
       - master
 
+concurrency:
+  # Group by workflow name and ref (branch or pull request ref).
+  # For pushes the ref is the branch (refs/heads/<branch>).
+  # For PR events the ref is refs/pull/<number>/merge — same across commits for that PR.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   test:
     name: ${{ matrix.name }}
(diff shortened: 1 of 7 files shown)


raoulcollenteur:
Hi @dbrakenhoff how far is this PR? Would be nice to include in the next release I think!

pastas/stressmodels.py
```diff
@@ -97,6 +105,11 @@ def __init__(
 
         self.stress = []
 
+        if CACHETOOLS_AVAILABLE:
+            self._cache = LRUCache(maxsize=max_cache_size)
```
martinvonk: so here you have self._cache = pastas.decorators.get_cacher()
```
def get_cacher(max_cache_size: int) -> Cacher | None:
    try: 
        import ...
    except:    
        ...
```

Then we can potentially use it in other places too right?

dbrakenhoff:
> Is this ready for review?

Yes!

martinvonk (approved):

Issue #931, opened 2025-03-12
Replacing series_original in a TimeSeries created as oseries fails


HenrikSpa:
When a TimeSeries is created for an observed series (using `settings = 'oseries'`), the series is validated as a non-equidistant series:

https://github.com/pastas/pastas/blob/cfe72569124f60c1bd99feec21ac79a81a2a4810/pastas/timeseries.py#L111-L113

However, when trying to replace this series using `series_original.setter`, the function `validate_stress` is called which requires an equidistant series instead of `validate_oseries`, and the replacement fails:

https://github.com/pastas/pastas/blob/cfe72569124f60c1bd99feec21ac79a81a2a4810/pastas/timeseries.py#L181-L184

It would be better if `validate_oseries` is always used if the TimeSeries was created using `settings = 'oseries'`.

rubencalje:
I encountered this issue today. I could not set the series_original of the oseries of a pastas Model. For example:

```
import numpy as np
import pandas as pd
import pastas as pt

n = 12

index = pd.date_range("2025-1-1", periods=n, freq="MS")
values = np.random.rand(n)
s = pd.Series(values, index=index)
ml = pt.Model(s)

ml.oseries.series_original = s
```

will fail as `s` is not equidistant. We should find a way to update the oseries without performing all the validation-checks of a stress. The way I solved it not is setting oseries by a TimeSeries-object:

`ml.oseries = pt.timeseries.TimeSeries(s, settings="oseries")`

raoulcollenteur:
To summarize for future reference, the new pattern to replace an existing oseries is by using `ml.set_oseries(new_oseries)`.

Closing issue as PR #1024 fixed this

PR #1024, opened 2025-10-31
Add `ps.Model.set_oseries()`

rubencalje:
# Short description
This PR adds the method set_oseries, to replace the oseries of a Pastas Model, see issue #931.

- [ ] closes issue #931
- [ ] is documented
- [ ] type hints
- [ ] tests added or expanded

### DIFF
--- a/pastas/model.py
+++ b/pastas/model.py
@@ -1116,7 +1116,9 @@ class Model:
         This method replaces the existing oseries with a new TimeSeries object while
         preserving the original metadata if no new metadata is provided.
         """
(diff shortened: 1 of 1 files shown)


raoulcollenteur:
> Maybe we could/should call `set_oseries` too on line 102 in the `__init__` method? `self.oseries = TimeSeries(oseries, settings="oseries", metadata=metadata)` to: `set_oseries(s=oseries, metadata=metadata)`

I agree this would be more consistent. Then in the future we can built around the set_oseries method :)

Issue #1018, opened 2025-10-06
Doc strings are not informative about default (keyword) values


mbakker7:
The docstrings have evolved over time, but are now not informative anymore regarding the default values.
For example, the doc string of the `solve` method is 
```
ml.solve(
    tmin: Optional[~TimestampType] = None,
    tmax: Optional[~TimestampType] = None,
    freq: str | None = None,
    warmup: float | None = None,
    noise: bool | None = None,
    solver: Optional[~Solver] = None,
    report: bool = True,
    initial: bool = True,
    weights: pandas.core.series.Series | None = None,
    fit_constant: bool = True,
    freq_obs: str | None = None,
    initialize: bool = True,
    **kwargs,
) -> None
Docstring:
Method to solve the time series model.
```
For many keywords, the default is `None`, but that is not the default value. For example, I think the `warmup` default value is 3650 (days=10 years), and for many others I don't even know what it is. Can we make those more informative?

raoulcollenteur:
we'll pick this up in #1031

raoulcollenteur:
Closed with merging #1031

PR #1031, opened 2025-11-06
Update solvers

### LINKED ISSUE #988: Make `Model` initialization solver dependent
When calling `ps.Model.solve`, the model is initialized with `ps.Model.initialize`. However, `ps.Model.initialize` is build mainly around the `ps.LeastSquares` solver with weights and everything. If initialize would be a method of the solver, the `Model` instance would be cleaner and methods would be specifically built for that solver. 

For instance, the `EmceeSolver` requires that there is a `dist` column in the `model.parameters` `DataFrame`, while the other solvers don't require this. If the `initialize` method would be in the solver, the `dist` column could be added to the parameters `DataFrame` via this method.

### LINKED ISSUE #1084: Model fit report always warns parameters could not be estimated well after load
Initially raised here: https://github.com/pastas/pastastore/issues/208

### Describe the bug
When saving a model to a pas file, and reloading, the fit_report will show the warning "Model parameters could not be estimated well". 

This is caused by the fact that initializing the model will set `ml._solve_success` to False, which is overwritten when the model is optimized. This value is not stored in the pas-file, meaning that it will always revert to False when a model is loaded from file. 

False is not a good initial value though, since it implies the solve was unsuccessful from the get-go. None would be a better initial value. The warnings should then only be triggered on False. This would solve the warning appearing when it shouldn't, but it won't solve the issue of the warning disappearing on save/load for models that did *not* solve successfully.

So, the next question is if we want to maintain this flag on load. That will require adding it to the load function and either adding it to the init (not very nice, imo) or adding logic to set the value in load.

### Expected behavior
Warning should not be shown after loading a model.

### Code to reproduce
```python
import pastas as ps
ml = ... # any pastas model
ml.to_file("test.pas")

ml2 = ps.io.load("test.pas")
print(ml2.fit_report())
```

### Issue checklist
- [x] Add labels
- [x] Add issue type
- [x] Assign yourself, a Pastas developer or no one
- [x] Check if there is not already an existing issue for this bug

### LINKED ISSUE #1152: Allow floats (1e4) for n in get_parameter_sample function

Allow floats (1e4) for n in get_parameter_sample function. Because I don't like writing 10000 and am lazy.

### Expected usage
A clear and concise description of what you expect to happen.

```python
# Code to show the expected the behavior
ci = ml.solver.ci_simulation(n=1e4, alpha=0.05)
```

Now this throws an error:

TypeError: 'float' object cannot be interpreted as an integer

### LINKED ISSUE #1157: Make model with custom solver settings reproducible
When I:
1. solve a pastas model using `x_scale=True`
2. save it to a pas file,
3. read it from a file
4. solve it again 

I get different results. It would be nice if the solver settings are saved in the model to make the model reproducible from a .pas file.

```
ml.solve(x_scale=True) # results
ml.to_file('model.pas')
ml2 = ps.io.load('model.pas')
ml2.solve() # different results
```

martinvonk:
# Short description
Update the solver routine.

- [x] closes issue #988
- [x] closes issue #1152
- [x] closes issue #1084
- [x] closes issue #1093 
- [x] closes issue #1018 
- [x] is documented
- [x] type hints
- [x] tests added or expanded


raoulcollenteur:
Let's try not use any AI to work on this PR because we really want to keep this as clean and simple as possible.

martinvonk:
I'm done for now with this PR. I think I'm done with most of the refactoring. 

The emceesolver does not work anymore now however... It needs a method for adding the dist column to the model parameters dataframe.

Feel free to work on this from now on. I'll be back in May.

martinvonk:
Summary of Changes:
This PR updates the solvers to make them more flexible for future implementations, successfully addressing issue #988.

Key Updates:

dist Column Removed: The Parameters DataFrame schema drops the dist column and removes old references to "dist" and "uniform".

API Change in set_parameter: The set_parameter function no longer accepts a dedicated dist argument. Instead, arbitrary properties are now passed through via **kwargs.

Safe Initialization for MCMC: Added a new initialize() method in EmceeSolve that safeguards models by automatically adding a dist column if it's missing (and logging a warning) to prevent crashes during MCMC runs.

Maintenance & Formatting: Updated relevant example notebooks (e.g., the Emcee notebook) to reflect the changes, and ran code formatting/linting using ruff.

dbrakenhoff:
I merged dev to get #1178 into this PR.

raoulcollenteur:
One option I was thinking of for the parameter dataframe and changing columns is to make a ParameterDataFrame class where the solver would register a new column that would than show up in all dataframes. Don't know if this is possible but if so, that might be a plan for the future. 

For now I kinda like the simple approach taken now in `ml.get_init_parameters` where adding the parameters for the solver adds the parameter columns, and setting them through ml.set_parameter even adds it back into the object..

martinvonk:
> One option I was thinking of for the parameter dataframe and changing columns is to make a ParameterDataFrame class where the solver would register a new column that would than show up in all dataframes. Don't know if this is possible but if so, that might be a plan for the future.
> 
> For now I kinda like the simple approach taken now in `ml.get_init_parameters` where adding the parameters for the solver adds the parameter columns, and setting them through ml.set_parameter even adds it back into the object..

I agree that the approach in this PR is good! However, I'm thinking that we might want to remove the stderr column by default as well, and onlly add that if you have a LeastSquares solver...

But I think we should make a new issue on how to handle parameters. If you have for instance 100 realizations with 100 initial parameter sets, then we might want to change things. Good to have a discussion on that in an issue.

martinvonk:
> > One option I was thinking of for the parameter dataframe and changing columns is to make a ParameterDataFrame class where the solver would register a new column that would than show up in all dataframes. Don't know if this is possible but if so, that might be a plan for the future.
> > For now I kinda like the simple approach taken now in `ml.get_init_parameters` where adding the parameters for the solver adds the parameter columns, and setting them through ml.set_parameter even adds it back into the object..
> 
> I agree that the approach in this PR is good! However, I'm thinking that we might want to remove the stderr column by default as well, and onlly add that if you have a LeastSquares solver...
> 
> But I think we should make a new issue on how to handle parameters. If you have for instance 100 realizations with 100 initial parameter sets, then we might want to change things. Good to have a discussion on that in an issue.

I opened a discussion: https://github.com/orgs/pastas/discussions/1223

raoulcollenteur:
- @martinvonk remove stderr everywhere and have LS add it
- @raoulcollenteur fix fit_report for emceesolve.

After this the PR will be opened for review.

martinvonk:
I'm done :)

raoulcollenteur:
This PR is ready for review! The actual solutions from ml.solve won't change (i.e., see the notebooks on the docs), but the internal structure is made future-proof to accommodate other solvers better, such as the upcoming PestSolve and MaximumLikelihoodSolve in Pastas > 2.0.

- solvers can now add columns to the ml.parameter dataframe, see for example the emceesolve.
- solvers have a custom fit_report
- There is a solver baseclass
- Documentation is strongly improved

It should be simpler to implement a new solver now, which we tested when I reimplemented mcmc again (it worked well :))

pastas/solver/least_squares.py
```diff
@@ -510,34 +720,98 @@ class LeastSquares(BaseSolver):
...
+        f_scale: float = 1.0,
+        max_nfev: int | None = None,
+        diff_step: float | ArrayLike | None = None,
+        tr_solver: Literal["exact", "lsmr"] | None = None,
+        tr_options: dict | None = None,
+        callback: Callable | None = None,
         pcov: DataFrame | None = None,
-        nfev: int | None = None,
         **kwargs,
     ) -> None:
```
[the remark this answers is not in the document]
raoulcollenteur: @martinvonk please click resolve on this.

pastas/io/base.py
```diff
@@ -135,29 +138,60 @@ def _load_model(data: dict) -> Model:
...
+                "the file with Pastas 1.3."
+            )
+        solver_name = data[solver_key].pop("class")
+        solver = getattr(ps.solver, solver_name)
+        ml.add_solver(solver(**data[solver_key]))
+
+    # Merge defaults with file parameters: file values win, defaults fill gaps.
+    file_parameters = data["parameters"].copy()
+
+    # Fix old parameter names for One response functions to match the naming convention from Pastas 2.0
```
dbrakenhoff, line 182: Maybe add a note here that this extra logic can be removed when pas-files < v2.0 are no longer supported?

pastas/io/base.py
```diff
@@ -135,29 +138,60 @@ def _load_model(data: dict) -> Model:
...
         ml.add_noisemodel(n)
 
-    # Add solver object to the model from pas-files < 1.3.0  TODO Deprecate
-    if "fit" in data.keys():
+    solver_name = None
+    for solver_key in ("fit", "solver"):
+        if solver_key not in data:
+            continue
+        if solver_key == "fit":
+            logger.warning(
```
dbrakenhoff: So are these files still supported now? So this is our warning to say, that these files will potentially no longer work in the future? Should we make that more explicit? If this is just a warning, can't they just load/save with Pastas 2.0?

pastas/io/base.py
```diff
@@ -135,29 +138,60 @@ def _load_model(data: dict) -> Model:
...
+    if "dist" in file_parameters.columns and solver_name != "EmceeSolve":
+        file_parameters = file_parameters.drop(columns=["dist"])
+
+    # Remove stderr column if it is empty, as this column is no longer used by default in the parameters table since 2.0
+    if "stderr" in file_parameters.columns and file_parameters["stderr"].isnull().all():
         logger.warning(
-            "The solver object is stored in the model.solver attribute since Pastas "
-            "1.3. Please update your pas-file to the new format by loading and saving "
-            "the file with Pastas 1.3."
+            "The 'stderr' column in the parameters table is longer used if "
```
dbrakenhoff: Why can't users load/save with 2.0?

pastas/solver/least_squares.py
```diff
@@ -0,0 +1,1220 @@
...
+        Parameters
+        ----------
+        name: str
+            Name of the block response for which to calculate the confidence interval.
+        n: int, optional
+            Number of random samples drawn from the bivariate normal distribution to compute the confidence interval. Default is 1000.
+        alpha: float, optional
+            Significance level for the confidence interval. Default is 0.05, which corresponds to a 95% confidence interval.
+        max_iter: int, optional
+            Maximum number of iterations for truncated multivariate sampling, default is 10. Increase this value if number of accepted parameter samples is lower than n.
```
dbrakenhoff: line lengths?

pastas/solver/least_squares.py
```diff
@@ -0,0 +1,1220 @@
...
+        p: ArrayLike,
+        noise: bool,
+        weights: Series | None = None,
+        callback: CallBack | None = None,
+        returnseparate: bool = False,
+    ) -> ArrayLike | tuple[ArrayLike, ArrayLike, ArrayLike]:
+        """
+        Wrapper for the shared `objfunction` to calculate residuals or noise.
+        """
+        return misfit(
```
dbrakenhoff: nitpicky: bit confusing to shadow method name

pastas/solver/least_squares.py
```diff
@@ -0,0 +1,1220 @@
...
+        pmin = (
+            parameters.loc[:, "pmin"].fillna(-np.inf).to_numpy(dtype=float, copy=True)
+        )
+        pmax = parameters.loc[:, "pmax"].fillna(np.inf).to_numpy(dtype=float, copy=True)
+
+        # Set the boundaries
+        if self.method == "lm":
+            logger.info(
+                "Method 'lm' does not support boundaries. Ignoring Pastas'"
+                "`pmin` and `pmax` parameter bounds and setting them to `nan`."
```
dbrakenhoff, line 798: ~+/-inf actually?~

nvm, this was about pmin/pmax

pastas/solver/least_squares.py
```diff
@@ -0,0 +1,1220 @@
...
+             (absolute_sigma=False) =pcov(absolute_sigma=True) * chisq(popt)/(M-N)``
+
+        Returns
+        -------
+        pcov: array_like
+            numpy array with the covariance matrix.
+
+        Notes
+        -----
+        This method is copied from Scipy (version 1.14.1):
```
dbrakenhoff: Separate issue, but should we check if this method ever got any updates in scipy in the meantime?

pastas/solver/least_squares.py
```diff
@@ -0,0 +1,1220 @@
...
+            "tr_solver": self.tr_solver,
+            "tr_options": self.tr_options,
+        }
+        return settings
+
+
+class LmfitSolve(LeastSquaresBase):
+    """Solving the model using the LmFit :cite:p:`newville_lmfitlmfit-py_2019`.
+
+    This is basically a wrapper around the scipy solvers, adding some cool
```
dbrakenhoff: not very scientific or formal language, but it is cool :)

pastas/solver/mcmc.py
```diff
@@ -0,0 +1,531 @@
...
+logger = getLogger(__name__)
+
+
+class EmceeSolve(SolverBase):
+    """Solver based on MCMC approach in emcee :cite:p:`foreman-mackey_emcee_2013`.
+
+    Parameters
+    ----------
+    objfunction: func, optional
+        An objective function to be minimized. See the pastas.likelihood_functions module for more information.
```
dbrakenhoff: pastas.solver.likelihood? Or just refer explicitly to GaussianLikelihood, etc?

pastas/solver/mcmc.py
```diff
@@ -0,0 +1,531 @@
...
+    References
+    ----------
+    https://emcee.readthedocs.io/en/stable/
+
+    See Also
+    --------
+    emcee.EnsembleSampler
+    emcee.moves
+    emcee.backend
+    pastas.objective_functions
```
dbrakenhoff: pastas.solver.objective_function

pastas/solver/mcmc.py
```diff
@@ -0,0 +1,531 @@
...
+
+    """
+
+    def __init__(
+        self,
+        name: str = "solver",
```
dbrakenhoff, line 92: Is it every useful for solvers to set their name attribute? Seems like it's never used and always set to solver now?

pastas/solver/mcmc.py
```diff
@@ -0,0 +1,531 @@
...
+
+        # Set priors
+        self._set_priors()
+
+        self.nsteps = steps
+
+        # Set initial positions of the walkers
+        pinit = self.initial[self.vary]
+        ndim = pinit.size
+        pinit = pinit + np.abs(pinit) * 1e-2 * np.random.randn(self.nwalkers, ndim)
```
dbrakenhoff, line 184: is this 1e-2 value something you would potentially want to tweak (per parameter)?

pastas/solver/mcmc.py
```diff
@@ -0,0 +1,531 @@
...
+
+        # Create sampler and run mcmc
+        if self.parallel:
+            logger.info("Going into the parallel universe")
```
dbrakenhoff, line 188: 🤓

pastas/solver/mcmc.py
```diff
@@ -0,0 +1,531 @@
...
+    ) -> float:
+        """Log-likelihood function.
+
+        Parameters
+        ----------
+        p: numpy.Array
+            Numpy array with the parameters.
+        noise: bool
+
+        weights
```
dbrakenhoff: docstring somewhat incomplete

pastas/solver/trackers.py
```diff
@@ -0,0 +1,427 @@
+"""Place to put TrackSolve"""
```
dbrakenhoff: Haha, great description

pastas/solver/trackers.py
```diff
@@ -0,0 +1,427 @@
...
+        pastas Model to track
+    tmin: pandas.Timestamp or str, optional
+        start time for simulation, by default None which defaults to first index in
+        ml.oseries.series
+    tmax: pandas.Timestamp or str, optional
+        end time for simulation, by default None which defaults to last index in
+        ml.oseries.series
+    update_iter : int, optional
+        if visualizing optimization progress, update plot every update_iter
+        iterations, by default nparam
```
dbrakenhoff, line 32: missing pause in docstring

pastas/model.py
```diff
@@ -828,21 +820,24 @@ def solve(
...
-            used for the calibration period.
+            used for the calibration period. If None, the warmup from ml.settings is
+            used (defaults to 3650 days).
         solver: Class pastas.solver.Solver, optional
-            Instance of a pastas Solver class used to solve the model. Options are:
-            ps.LeastSquares() (default) or ps.LmfitSolve(). An instance is needed as
-            of Pastas 0.23, not a class!
-        report: bool | Literal["full"], optional
+            Instance of a pastas Solver class used to solve the model. See pastas.
+            solver for the options. If None, the solver from ml.solver is used, and if
```
dbrakenhoff: This is a bit confusing? If you provide nothing you get LeastSquares, when is ml.solver not None? When you're loading an already solved model I guess? Sentence could be clearer.

pastas/model.py
```diff
@@ -828,21 +820,24 @@ def solve(
...
+            that is also None, the default ps.LeastSquares() is used.
+        report: bool | Literal["full"] | dict, optional
             Print a report to the screen after optimization finished. Set to
             True (default) to print a standard report, "full" to print a
             report including the correlation matrix and standard errors of the
-            parameters, or False to suppress the report. This can also be
-            manually triggered after optimization by calling print(ml.fit_report(
-            )) on the Pastas model instance.
+            parameters, or False to suppress the report. To have full conrol over the
```
dbrakenhoff: Also a bit confusing? What is the extra control I can have, and why would I want that? Maybe just mention the other methods without too much extra and maybe unnecessary info?

pastas/model.py
```diff
@@ -874,9 +869,14 @@ def solve(
...
             is False and deprecated since version 2.0.0.
+
+            .. deprecated:: 2.0.0
+                The initialize argument is deprecated in favor of the `reset_settings`
+                argument.
+
         reset_settings: bool = False,
-            If True, the model settings are reset to their default values before solving.
-            This calls the Model.set_settings() method with default values.
+            If True, the model settings are reset to their default values before
```
dbrakenhoff, line 882: So initial is deprecated in favor of reset_settings. reset_settings mentions returning all model settings back to their defaults. It does not mention anything about initial parameters. Should we add that?

Also the name reset_settings to me indicates we're resetting the settings, which could in theory be separate from the initial parameter values? So this isn't immediately clear to me from reading this.

pastas/model.py
```diff
@@ -977,34 +976,34 @@ def solve(
...
         # Solve model
-        noise = True if self.noisemodel else False
-        success, optimal, stderr = self.solver.solve(
-            noise=noise, weights=weights, **kwargs
-        )
-        if not success:
-            logger.warning("Model parameters could not be estimated well.")
+        solve_success, result = self.solver.solve(weights=weights, **kwargs)
+
+        # Update the parameters with the results from the optimization
```
dbrakenhoff, line 986: These parameters do not have to be set using ml.set_parameter() because they only live in the model class, right? E.g. optimal, stderr, etc.

pastas/model.py
```diff
@@ -1138,11 +1137,13 @@ def set_parameter(
             maximum value for the parameter.
         optimal: float, optional
             optimal value for the parameter.
-        dist: str, optional
-            Distribution of the parameters.
         move_bounds: bool, optional
-            Reset pmin/pmax based on new initial value. Of move_bounds=True, pmin and
-            pmax must be None.
+            Reset pmin/pmax based on new initial value. Of move_bounds=True,
```
dbrakenhoff: Of --> If

pastas/model.py
```diff
@@ -2036,7 +2061,7 @@ def _generate_warnings_report(self, log: bool = True) -> list[str]:
 
         msg = []
         # model optimization unsuccessful
-        if not self._solve_success:
+        if solve_success is False:
             msg.append("Model parameters could not be estimated well.")
```
dbrakenhoff, line 2070: can we provide any guidance with this message where to find more diagnostics?

dbrakenhoff (approved):
Really nice work. Much cleaner, and ready for future developments! 

I commented on a few minor docstring things! But approving in advance!

pastas/solver/mcmc.py
dbrakenhoff: Ah I see it was imported so should be `pastas.likelihood`

pastas/io/base.py
martinvonk: User can load the 'stderr' column, but only if the solver is `LeastSquares` or `LmfitSolve`. Adding `LeastSquares` or `LmfitSolve` to the model, adds the `stderr` column to the model parameters dataframe. If this column (and solver) is present it, the 'stderr' from the pas file is added to the model parameters dataframe.

pastas/io/base.py
martinvonk: Good point.

pastas/solver/least_squares.py
martinvonk: agree, we don't need it anymore as well.

martinvonk:
Thanks for the review @dbrakenhoff. I fixed some of your comments and replied to some other. Some comments are still open for @raoulcollenteur :)

I also made a small change to the fit_report @raoulcollenteur made because it was incompatible with the SolverBase class. Also I changed the name `all_options` to `full_output` because I thought `all_options` was a but vague.. Maybe `full_report` would be even better as a variable name..

Do you also have some time to look at this PR @OnnoEbbens? I thought you had some things you wanted to test with the new solver?

pastas/io/base.py
dbrakenhoff: So if I understand you correctly, there are pas-files that cannot be loaded with 2.0 when stderr is present in the parameters dataframe, and the solver is not LmfitSolve or LeastSquares. And that is why the warning suggests using 1.5 to update the pas file?

My guess is the number of saved models out there with alternative solvers is very very small? EmceeSolver never supported to_dict, and I'm not sure if some of your other solvers support that? So there is a very high probability you are able to load these older pas-files with 2.0? Just wondering out loud if it's worth the warning, when there is a very high chance the load will succeed with pastas 2.0.

dbrakenhoff:
> I also made a small change to the fit_report @raoulcollenteur made because it was incompatible with the SolverBase class. Also I changed the name `all_options` to `full_output` because I thought `all_options` was a but vague.. Maybe `full_report` would be even better as a variable name..

I think `full_output=True` is nice, similar to what scipy uses in some of their optimization funcs. If you're considering `fit_report(full_report=True)` I think calling `fit_report(full=True)` is also pretty clear to me.

pastas/io/base.py
martinvonk: I see the total message better now. The Pastas 1.5 update suggestion should be removed. That has nothing to do with this..

To answer the rest, no the probability is not very high that: there is a not null "stderr" column while the solver is None or EmceeSolve. So I'd be fine with removing this message.

raoulcollenteur:
@martinvonk looks like you made the required changes already and this can be merged after or so I miss something?

pastas/solver/least_squares.py
```diff
@@ -1069,7 +1069,15 @@ def to_dict(self) -> dict:
         return settings
 
 
-class LmfitSolve(LeastSquaresBase):
+@PastasDeprecationWarning(
+    version="2.3.0", reason="The EmceeSolve class is renamed to Emcee."
```
dbrakenhoff: wrong message?

PR #1056, opened 2025-11-19
Make `StressModel.stress` a property option 1

### LINKED ISSUE #985: Make `StressModel.stress` a `@property`
Making the stress attribute of a pastas.StressModel a property would allow individual stresses to be accessed and modified by their name, which is more intuitive and flexible than the current method. For instance, usefull with the forecasting module (#945).

Currently, the stress attribute of a StressModel is a simple list of pastas.TimeSeries objects. If you want to change one of the stresses, you have to create an entirely new list with the updated time series and then reassign it to the stress attribute.

### Usage
```python
prec: pd.Series = ...
prec_new: pd.Series = ...
evap: pd.Series = ...
rm = ps.RechargeModel(prec, evap)
# rm.stress would yield: [ps.TimeSeries(prec), ps.TimeSeries(evap)]
```
#### Current
For example, if a model has prec and evap stresses, and you want to update the prec data, the current workflow is:
```python
# current behavior to change stress
rm.stress = [ps.TimeSeries(prec_new), ps.TimeSeries(evap)]
```

#### Expected
The proposed feature would change the stress attribute from a list to a property. This change would allow you to directly access and modify a stress by name, similar to a regular object attribute.

With this change, the process of updating the prec data would become much cleaner:
```python
rm.prec = prec_new
# rm.stress would yield: [ps.TimeSeries(prec2), ps.TimeSeries(evap)]
```

### Additional context
This might give some issues with the standard `StressModel` where the `stress` is parsed as keyword argument on initialization but converted to a list as the attribute.

martinvonk:
# Short description
This ensures that the stresses can't be directly manipulated anymore, in line with #1014. Additionaly, changing stresses is now more logical where manipulation now occurs via the attribute that the user provided.

old behavior for changing a stress in the stressmodel
```python
rm: ps.RechargeModel = ...
rm.stress[0] = ps.Timeseries(series)
```
new behavior:
```python
sm.prec = ps.ts.TimeSeries(series)
```

This does introduce a breaking change with previous pastas versions where the first argument to the `StressModel.__init__` is now called `s` instead of `stress`. I thought about naming it `series` or something else but we are not very consistent in pastas. You can find `stresses`, `s`, `series` in different places (same as both `head`, `oseries`, and `obs` for the head observations) but that is a separate issue I think.

- [x] closes issue #985 
- [x] is documented
- [x] type hints
- [x] tests added or expanded


martinvonk:
> So it is not possible to change just 1 stress of the WellModel anymore?

No it still is. I should have been more clear. But you now change it via the `s` argument which the user provided themselves as a list (of pandas series or TimeSeries). Only for the WellModel the original behaviour looks like what I now propose for all other StressModels. So old behaviour:
```
wm: ps.WellModel = ...
wm.stress[0] = TimeSeries(...)
```
is basically the same as the new behaviour but that's only because for WellModel you already had to parse a list:
```python
wm.s[0] = TimeSeries(...)
```

martinvonk:
> Another question: would it not be great if you did not have to supply a TimeSeries, but only a pandas Series (just like you originally provided), copying all settings from the existing TimeSeries?

Yes would be great but that I'm not sure if that should be part of this PR or separate. Because that would be better for the implementation we now have as well. With this PR at least it would be easier to implement where you can either make `self.s`, `self.prec`, `self.evap` etc. a `property` too with a setter method or/and add a `add_stress` or `update_stress` method.

rubencalje:
Hmm, I still do not understand why we cannot just let the first argument of StressModel (or WellModel) be called `stress`. This is much clearer than `s` I think, does not create a breaking change, and saves us a lot of deprecation warnings.

martinvonk:
Maybe we could make `self.stress_tuple` a NamedTuple so you can do both `rm.stress_tuple.prec` and `rm.stress_tuple[0]`

PR #1065, opened 2025-11-24
Make `StressModel.stress` a property and create `add_stress` method

martinvonk:
# Short description
This ensures that the stresses can't be directly manipulated anymore, in line with #1014. Additionaly, changing stresses is now more logical where manipulation now occurs via the attribute that the user provided.

old behavior for changing a stress in the stressmodel
```python
series: pd.Series = ...
rm: ps.RechargeModel = ...
rm.stress[0] = ps.Timeseries(series)
```
new behavior:
```python
sm.prec = series
```

This does introduce a breaking change with previous pastas versions where the way the stresses are stored is now called `_stress` and `stress` is a `@property`. The behavior for calling `stress` is now also different where a `TimeSeries` is returned, not a list of `TimeSeries`. This is an alternative solution to #1056. If all stresses in a `StressModel` are needed users can call `.stresses` property which returns a `named_tuple` with all stresses in a `StressModel`.

### EDIT 04/12/2025
Renamed `stress_tuple` to `stresses` after a discussion with @dbrakenhoff 

- [x] closes issue #985 
- [x] is documented
- [x] type hints
- [x] tests added or expanded

### DIFF
--- a/pastas/stressmodels.py
+++ b/pastas/stressmodels.py
@@ -292,7 +292,7 @@ class StressModel(StressModelBase):
     settings: dict or str, optional
         The settings of the stress. This can be a string referring to a predefined
         settings dictionary (defined in ps.rcParams["timeseries"]), or a dictionary with
-        the settings to apply. For more information refer to Time series settings
+        the settings to apply. For more information refer to time series settings
         section below.
     metadata: dict, optional
         dictionary containing metadata about the stress. This is passed onto the
@@ -413,7 +413,21 @@ class StressModel(StressModelBase):
         settings: str | StressSettingsDict | None = None,
         metadata: dict[str, Any] | None = None,
     ) -> None:
-        """Set the stress time series."""
+        """Set the stress time series.
+
+        Parameters
+        ----------
+        stress: pandas.Series or pastas.TimeSeries
+            pandas.Series with pandas.DatetimeIndex containing the stress or a
+            pastas TimeSeries object.
+        settings: dict or str, optional
+            The settings of the stress. This can be a string referring to a predefined
+            settings dictionary (defined in ps.rcParams["timeseries"]), or a dictionary
+            with the settings to apply. For more information refer to time series
+            settings section on class initialization.
+        metadata: dict, optional
+            Dictionary with metadata of the stress time series.
+        """
         if isinstance(stress, TimeSeries):
             self._stress = stress
         else:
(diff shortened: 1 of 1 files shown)


pastas/stressmodels.py
```diff
@@ -214,34 +191,45 @@ def get_stress(
...
+        freq: str | None = None,
+    ) -> None:
+        """Placeholder method for updating the stress time series."""
+        _ = tmin, tmax, freq
+        pass
 
-    def to_dict(self, **kwargs):
-        """Method to export the stress model object."""
+    def to_dict(self, **kwargs) -> None:
+        """Placeholder method the to_dict method."""
```
dbrakenhoff: weird docstring

pastas/stressmodels.py
```diff
@@ -861,16 +958,24 @@ class WellModel(StressModelBase):
...
+        | StressSettingsDict
+        | Iterable[str]
+        | Iterable[StressSettingsDict] = "well",
         sort_wells: bool = True,
-        metadata: list[dict[str, Any]] = None,
+        metadata: Iterable[dict[str, Any]] = None,
         max_cache_size: int = None,
     ) -> None:
+        if not isinstance(stress, (tuple, list, dict)):
+            msg = f"The stress parameter must be a collection of pandas Series, not {type(stress)}"
```
dbrakenhoff: TimeSeries are also accepted?

dbrakenhoff (approved):
Some comments to think about, but overall I approve. 

There is a little unintuitiveness to the implementation for WellModel where 

```python
# given some wellmodel with N stresses
wm.stress[0] = new_stress  # error
sm.stress = new_stress # works
```

But since 
- we do a lot of internal checking with clear error messages, 
- we don't recommend this approach anywhere, 
- we push users to use `set_stress()` 
 
I think this is okay.

pastas/stressmodels.py
```diff
@@ -1341,8 +1562,12 @@ def __init__(
             recharge = Linear()
 
         # Store the precipitation and evaporation time series
-        self.prec = TimeSeries(prec, settings=settings[0], metadata=metadata[0])
-        self.evap = TimeSeries(evap, settings=settings[1], metadata=metadata[1])
+        self.set_stress(
```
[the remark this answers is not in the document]
martinvonk: good idea

raoulcollenteur (approved):
Thanks for the PR @martinvonk, looks good to me. A couple of minor things only. 

Curious to see who used these attributes and methods that we change and assume are not used, guess we'll see and hope for the best :)

PR #1069, opened 2025-12-08
Remove `oseries_calib` and introduce `ObservationSeries`

rubencalje:
# Short description
This PR removes the ugly `oseries_calib` attribute of a Pastas Model. The observations that the model is calibrated on can now be found in `ml.oseries.series`. `ml.oseries` is a `ObservationSeries`-object, which inherits from `TimeSeries`. Some of the functionality in `ml.observations`, to calculate `oseries_calib`, is now moved to `ObservationSeries.update_series`.

I think in the future, we can also introduce a `StressSeries`-object (which is an idea of @martinvonk), and move many of the methods from `TimeSeries` to `StressSeries`. I wanted to focus on the oseries in this PR however.

The user can still request the measurements on which the model was calibrated using `ml.observations()`, so this should not impact reqular use. Only in scripts where `oseries_calib` is used directly, this would result in an error. Therefore, I added a property called `oseries_calib` that redirects to `ml.oseries.series`, so we can show the user a warning that this has been changed.

- [x] closes issue (discussion) #1042
- [ ] is documented
- [ ] type hints
- [ ] tests added or expanded

### DIFF
--- a/pastas/timeseries.py
+++ b/pastas/timeseries.py
@@ -711,9 +711,9 @@ class ObservationSeries(TimeSeries):
                 series = self._fill_nan(series)
 
             if freq is not None and not series.empty:
-                # sample measurements, so that frequency is not higher than model keep
-                # the original timestamps, as they will be used during interpolation of
-                # the simulation
+                # Sample the measurements so that the observation frequency does not
+                # exceed the model frequency. Make sure to retain the original
+                # timestamps, as they will be used for interpolation of the simulation.
                 sim_index = _get_sim_index(tmin, tmax, freq, time_offset)
                 index = get_sample(series.index, sim_index)
                 series = series.loc[index]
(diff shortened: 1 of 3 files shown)


pastas/timeseries.py
```diff
@@ -627,6 +636,97 @@ def to_dict(self, series: bool | None = True) -> dict:
...
+            # Get the original series to start with
+            series = self._series_original.copy(deep=True)
+            series = series.loc[tmin:tmax]
+
+            # Only fill_nans if necessary
+            if series.hasnans:
+                series = self._fill_nan(series)
+
+            if freq is not None and not series.empty:
+                # sample measurements, so that frequency is not higher than model keep
```
raoulcollenteur: Higher than the model. (?) Keep ...

Not sure how but the sentence should be updated here

raoulcollenteur (approved):
Looks good. Couple of small things but should be easy to answer or fix.

Issue #1073, opened 2025-12-11
Add `moment` method for the response functions


martinvonk:
### Describe the proposed feature
Compute the raw moments for the response functions. Can either be discrete or exact depending on the response function.

### Expected usage

Usage would be:
```python
p = [0.4, 1.1, 100.0]
ps.Gamma().moment(p=p, order=0, method="exact")
```

The code would look something like this:

```python
def moment(a: np.typing.NDArray[np.float64], order: int, dt: float = 1.0) -> float:
    """
    Compute the raw moments of an impulse response array.

    Parameters
    ----------
    a : array-like
        Impulse response values.
    order : int
        The order of the moment to compute (0 through 5).
    dt : float, optional
        Time step between impulse response values. Default is 1.0.

    Returns
    -------
    moment : float
        The computed raw moment of the impulse response.
    """
    t = np.arange(len(a)) * dt

    return float(np.sum((t**order) * a))
```

```python
from pastas.stats import moment as moment_discrete

class Gamma()
    ...

def moment(self, p: ArrayLike, order: int, method: Literal["exact", "discrete"], dt: float = 1.0) -> float:
    if method == "exact":
        """From @OnnoEbbens MSc Thesis (2015)"""
        A, n, a = p # A, alpha, beta in Ebbens
        return A * gamma(n + order) / gamma(n) * a**order
    elif method == "discrete":
        block = self.block(p=p, dt=dt)
        return moment_discrete(block, dt=dt)

```

### Additional context
There is also a `scipy.stats.moment` function but that calculates the nth moment about the mean for a sample, and not the raw moment. 

If the cutoff is too low and the dt is too high then the discrete moment is sligthly inaccurate compared to the exact method. Here is an example for the Gamma reponse for A=0.4, n=1.1, a=100.0.

|                                                                         |         M0 |         M1 |         M2 |         M3 |         M4 |         M5 |
|-------------------------------------------------------------------------|-----------:|-----------:|-----------:|-----------:|-----------:|-----------:|
| gamma.moment (exact)                                                    | 4.0000e-01 | 4.4000e+01 | 9.2400e+03 | 2.8644e+06 | 1.1744e+09 | 5.9895e+11 |
| gamma.moment (discrete) default cutoff (0.999) and dt (1.0)             | 3.9960e-01 | 4.3471e+01 | 8.9234e+03 | 2.6203e+06 | 9.7003e+08 | 4.2011e+11 |
[...]

PR #1074, opened 2025-12-11
Add `moment` method for response functions

martinvonk:
# Short description
Compute the raw moments for the response functions. Can either be discrete or exact depending on the response function.

- [x] closes issue #1073
- [x] is documented
- [x] type hints
- [x] tests added or expanded


dbrakenhoff (approved):
Some small comments.

PR #1094, opened 2026-02-03
Replace `timestep_weighted_resample` by `time_weighted_resample`

rubencalje:
# Short description
Replace `timestep_weighted_resample` by `time_weighted_resample`. This new method is faster (no need for a separate `fast`-option), is more robust, and has the option to choose if the original series needs to be interpreted as stepwise (a flux) or linear (a state/stage).

- [ ] closes issue #xxxx
- [ ] is documented
- [ ] type hints
- [ ] tests added or expanded

### DIFF
--- a/.github/workflows/test_format_lint.yml
+++ b/.github/workflows/test_format_lint.yml
@@ -18,6 +18,9 @@ jobs:
           - name: Formatting and linting with ruff
             python: "3.11"
             toxenv: ruff
+          - name: Formatting and linting with ruff (Preview Mode)
+            python: "3.11"
+            toxenv: ruff_preview
     env:
       # Color Output
       # Rich (pip)
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.4.4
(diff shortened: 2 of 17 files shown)


raoulcollenteur:
What is the status of this PR @rubencalje ? Should I review it already or is it still undergoing changes?

raoulcollenteur:
Looks good, is there a reason for not merging this into the dev-branch such that we can release it with 1.14.0 rather than 2.0?

rubencalje:
> 1. I would prefer to merge this into the dev branch and release it with 1.14.0, but if there's a good reason for 2.0 (so dev2) 

fine by me

> 2. Perhaps the old method could be rewritten to internally use the new method?

The new method does not have to give the exact same results as the new method, so I think it is better to leave the old method intact.

ArtesiaWater:
> 1. I would prefer to merge this into the dev branch and release it with 1.14.0, but if there's a good reason for 2.0 (so dev2) that's fine with me.

I'd merge this with 2.0 so that it works with Pandas 3.0. Because that might give some different outcomes already...

martinvonk:
> > 1. I would prefer to merge this into the dev branch and release it with 1.14.0, but if there's a good reason for 2.0 (so dev2) that's fine with me.
> 
> I'd merge this with 2.0 so that it works with Pandas 3.0. Because that might give some different outcomes already...

This was me :)

PR #1131, opened 2026-03-10
Update `PastasDeprecationWarning` and `deprecate_args_or_kwargs` with `DeprecationWarning` and `Attribute/TypeError`

### LINKED ISSUE #701: List of deprecations for Pastas 2.0
Here a list is maintained of things to remove/change in 2.0:

- [x] `kge_2012`
- [x] `contributions_pie`
- [x] Pandas version >= 2.2
- [x] Remove noise argument from solve and model.__init__
- [x] Remove NoiseModel and ArmaModel and uncomment lines in base.py (from line 119)
- [x] Remove lmfit and latexify arguments from get_version in version.py
- [x] Remove Edelman response function
- [x] Remove `output` argument from `fit_report`

_Originally posted by @martinvonk in https://github.com/pastas/pastas/issues/700#issuecomment-1983099752_

martinvonk:
### Changes
* Refactored `PastasDeprecationWarning` to use a `version` argument (instead of `remove_version`), clarified its behavior: logs a warning if the current version is less than the specified version, and raises `AttributeError` if the current version is greater or equal, signaling that the feature can, but should not per se, be removed.
* Refactored `deprecate_args_or_kwargs` to use a `version` argument (instead of `remove_version` and `force_raise`), clarified its behavior: logs a warning if the current version is less than the specified version, and raises `TypeError` if the current version is greater or equal. Improved documentation and error messages for clarity.
* Replaced usage of `logger.warning` with the standard library `warnings.warn` for deprecation warnings to better align with Python best practices as requested by @rubencalje.

### DIFF
--- a/pastas/decorators.py
+++ b/pastas/decorators.py
@@ -175,14 +175,13 @@ def deprecate_args_or_kwargs(
     deprecate_version: str,
     remove_version: str,
     reason: str = "",
-    force_raise: bool = False,
 ):
     """Provide a warning or error when a function argument is deprecated.
 
     This function raises errors or warnings based on the current Pastas version and the
     deprecation timeline. The behavior is:
 
-    - If current version < deprecate_version: logs a warning (or raises if force_raise=True)
+    - If current version < deprecate_version: logs a warning
     - If deprecate_version <= current version < remove_version: raises DeprecationWarning
     - If current version >= remove_version: raises TypeError which indicates that it can be
     removed from the codebase entirely
@@ -198,15 +197,11 @@ def deprecate_args_or_kwargs(
     reason: str, optional
         The reason why the argument is deprecated, or a message directing users to
         an alternative. Default is an empty string.
-    force_raise: bool, optional
-        If True, raise DeprecationWarning immediately even if the current version has
-        not yet reached deprecate_version. Default is False.
 
     Raises
     ------
     DeprecationWarning
-        If between deprecate_version and remove_version (inclusive), or if force_raise=True
-        and current version < deprecate_version.
+        If between deprecate_version and remove_version and current version < deprecate_version.
     TypeError
         If current version >= remove_version and the argument is used.
     """
@@ -218,10 +213,7 @@ def deprecate_args_or_kwargs(
             f"The {name} argument is deprecated and will not be available"
             f" from Pastas version >={DEPRECATE_VERSION}. {reason}"
         )
-        if force_raise:
-            raise DeprecationWarning(msg)
-        else:
-            logger.warning(msg)
+        logger.warning(msg)
     elif CURRENT_PASTAS_VERSION >= REMOVE_VERSION:
         raise TypeError("got an unexpected keyword argument '%s'" % name)
     else:
--- a/pastas/decorators.py
+++ b/pastas/decorators.py
@@ -121,13 +121,12 @@ def PastasDeprecationWarning(version: str, reason: str = "") -> Any:
 
     - If current version < version: logs a warning and allows execution to continue
     - If current version >= version: raises AttributeError which indicates
-    that it can be removed from the codebase entirely
+    that it can be removed from the codebase entirely in the (near) future.
 
     Parameters
     ----------
     version: str
-        The version in which the function or class will be removed from the Pastas codebase
-        and raises an AttributeError.
+        The version in which the function, method or class raises an AttributeError.
     reason: str, optional
         The reason why the function or class is deprecated, or a message directing users
         to an alternative. Default is an empty string.
@@ -146,7 +145,7 @@ def PastasDeprecationWarning(version: str, reason: str = "") -> Any:
             if CURRENT_PASTAS_VERSION < VERSION:
                 msg = (
                     f"{name} is deprecated and will not be available "
-                    f"from Pastas version >={VERSION}. {reason}"
+                    f"from Pastas version >= {VERSION}. {reason}"
                 )
                 warn(message=msg, category=DeprecationWarning)
             else:
@@ -172,14 +171,14 @@ def deprecate_args_or_kwargs(name: str, version: str, reason: str = "") -> None:
 
     - If current version < version: logs a warning
     - If current version >= version: raises TypeError which indicates that it can be
-    removed from the codebase entirely
+    removed from the codebase entirely in the (near) future.
 
     Parameters
     ----------
     name: str
         The name of the argument that is deprecated.
     version: str
-        The version in which the argument will be removed and TypeError will be raised.
+        The version in which using the argument will raise a TypeError.
     reason: str, optional
         The reason why the argument is deprecated, or a message directing users to
         an alternative. Default is an empty string.
@@ -195,7 +194,7 @@ def deprecate_args_or_kwargs(name: str, version: str, reason: str = "") -> None:
     if CURRENT_PASTAS_VERSION < VERSION:
         msg = (
             f"The {name} argument is deprecated and will not be available"
-            f" from Pastas version >={VERSION}. {reason}"
+            f" from Pastas version >= {VERSION}. {reason}"
         )
         warn(message=msg, category=DeprecationWarning)
     else:
--- a/pastas/decorators.py
+++ b/pastas/decorators.py
@@ -149,8 +149,8 @@ def PastasDeprecationWarning(
             REMOVE_VERSION = parse_version(remove_version)
             if CURRENT_PASTAS_VERSION < DEPRECATE_VERSION:
                 msg = (
-                    f"{name} is deprecated and will not available anymore"
-                    f" in Pastas version {DEPRECATE_VERSION}. {reason}"
+                    f"{name} is deprecated and will not be available "
+                    f"from Pastas version >={DEPRECATE_VERSION}. {reason}"
                 )
                 logger.warning(msg)
             elif CURRENT_PASTAS_VERSION >= REMOVE_VERSION:
@@ -214,7 +214,7 @@ def deprecate_args_or_kwargs(
 
     if CURRENT_PASTAS_VERSION < DEPRECATE_VERSION:
         msg = (
-            f"The {name} argument is deprecated and will not available"
+            f"The {name} argument is deprecated and will not be available"
             f" from Pastas version >={DEPRECATE_VERSION}. {reason}"
         )
         if force_raise:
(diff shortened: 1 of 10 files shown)


pastas/decorators.py
```diff
@@ -112,38 +112,54 @@ def _model_tmin_tmax(
...
                 msg = (
-                    "%s is deprecated and will be removed in Pastas version %s. "
-                    % (name, remove_version)
-                ) + reason
+                    f"{name} is deprecated and will not available anymore"
+                    f" in Pastas version {DEPRECATE_VERSION}. {reason}"
+                )
                 logger.warning(msg)
+            elif CURRENT_PASTAS_VERSION >= REMOVE_VERSION:
+                raise ModuleNotFoundError("no module named '%s'" % name)
```
dbrakenhoff: A ModuleNotFoundError is a bit odd here no?

```python
import pastas as ps
ps.Foo # --> AttributeError
```

pastas/decorators.py
```diff
@@ -154,41 +170,64 @@ def _function(*args, **kwargs):
...
+
+    if CURRENT_PASTAS_VERSION < DEPRECATE_VERSION:
         msg = (
-            "The '%s' argument is deprecated and will be removed in Pastas version %s. "
-            % (name, remove_version)
-        ) + reason
-        logger.warning(msg)
-    else:
+            f"The {name} argument is deprecated and will not available"
+            f" anymore in Pastas version {DEPRECATE_VERSION}. {reason}"
```
dbrakenhoff:
```suggestion
            f" from Pastas version >={DEPRECATE_VERSION}. {reason}"
```

pastas/decorators.py
```diff
@@ -154,41 +170,64 @@ def _function(*args, **kwargs):
...
-                % (name, remove_version)
-            ) + reason
-
+            logger.warning(msg)
+    elif CURRENT_PASTAS_VERSION >= REMOVE_VERSION:
+        raise TypeError("got an unexpected keyword argument '%s'" % name)
+    else:
+        msg = (
+            f"The {name} argument is deprecated and is not available "
+            f"anymore since Pastas version {DEPRECATE_VERSION}. {reason}"
```
dbrakenhoff:
```suggestion
            f"since Pastas version {DEPRECATE_VERSION}. {reason}"
```

pastas/model.py
```diff
@@ -162,18 +162,19 @@ def __init__(
...
                     "anymore, so passing noisemodel=False is not needed anymore. To "
                     "fix this error, do not pass noisemodel=False to Model."
                 )
-                deprecate_args_or_kwargs(
-                    "noisemodel", "2.0.0", reason=msg, force_raise=True
-                )
+            deprecate_args_or_kwargs(
+                name="noisemodel",
+                deprecate_version="2.0.0",
+                remove_version="2.2.0",
```
dbrakenhoff: We should confirm we want this to disappear after 2 minor versions. I believe @rubencalje suggested keeping certain kwargs in until 3.0.

tests/test_deprecator.py
```diff
@@ -6,78 +6,140 @@
...
+    # class will be removed in future version - should log warning
+    @PastasDeprecationWarning(
+        deprecate_version="999.0.0", remove_version="1000.0.0", reason="Boo!"
+    )
     class Deprecated:
         def __init__(self, a: Any) -> None:
             self.a = a
 
-    Deprecated(1)
+    Deprecated(1)  # logs warning, continues execution
```
dbrakenhoff: This only tests execution, not if the warning is shown. We could capture the log output to test?

tests/test_deprecator.py
```diff
@@ -6,78 +6,140 @@
...
 
-    # class was deprecated in older version
+    # method is currently deprecated (between deprecate and remove versions) - should raise DeprecationWarning
     class Deprecated:
         def __init__(self, a: Any) -> None:
             self.a = a
 
-        @PastasDeprecationWarning(remove_version="1.0.0", reason="Boo!")
+        @PastasDeprecationWarning(
+            deprecate_version="1.0.0", remove_version="2.0.0", reason="Boo!"
```
dbrakenhoff: maybe increase remove_version for this test so we don't ever have to think about updating remove_version.

tests/test_deprecator.py
```diff
@@ -6,78 +6,140 @@
...
             self.a = a
 
-    Deprecated(1)
+    Deprecated(1)  # logs warning, continues execution
 
-    # class was deprecated in older version
-    @PastasDeprecationWarning(remove_version="1.0.0", reason="Boo!")
+    # class is currently deprecated (between deprecate and remove versions) - should raise DeprecationWarning
+    @PastasDeprecationWarning(
+        deprecate_version="1.0.0", remove_version="2.0.0", reason="Boo!"
```
dbrakenhoff: increase remove_version so we don't have to think about updating it.

tests/test_deprecator.py
```diff
@@ -6,78 +6,140 @@
...
+            reason="Boo!",
+            force_raise=True,
         )
 
-    # raise error for past deprecation
+    # raise error when between deprecate and remove versions
     with pytest.raises(DeprecationWarning):
-        deprecate_args_or_kwargs("test", remove_version="1.0.0", reason="Boo!")
+        deprecate_args_or_kwargs(
+            "test", deprecate_version="1.0.0", remove_version="2.0.0", reason="Boo!"
```
dbrakenhoff: increase remove_version

dbrakenhoff (commented):
Looking good, I left some comments to consider before we merge.

pastas/decorators.py
martinvonk: You're right. I don't know why I thought it should be an `ModuleNotFoundError`. I'll transform it into an AttributeError.

tests/test_deprecator.py
martinvonk: Yes good point.

rubencalje:
I think we are making this much more complicated than it needs to be.

I think we should just warn the user that functionality will change, from a certain major version far away in the future (3.0 for example). I think we should never raise a warning (is this done in other packages?), but just warn the user using warnings.warn and/or logger.warning.

After that version, we remove the functionality from the code, and the code of the user stops working. But because they should have seen the warning in many versions before, it is their own fault.

pastas/decorators.py
martinvonk: Technically this could also be an `ImportError` if you say `from pastas.noisemodel import Foo`

martinvonk:
> I think we are making this much more complicated than it needs to be.
> 
> I think we should just warn the user that functionality will change, from a certain major version far away in the future (3.0 for example). I think we should never raise a warning (is this done in other packages?), but just warn the user using warnings.warn and/or logger.warning.
> 
> After that version, we remove the functionality from the code, and the code of the user stops working. But because they should have seen the warning in many versions before, it is their own fault.

I agree it adds some complexity under the hood, but the goal of this decorator is actually to automate exactly what you're suggesting. It formalizes the timeline so we don't have to manually keep track of when to use `logger.warning` vs `warnings.warn` vs `DeprecationWarning` vs `FutureWarning`, and when to finally drop the code.

This keeps things highly user-friendly without us having to constantly think about it. Also, the core of this approach was already decided since https://github.com/pastas/pastas/issues/800 and https://github.com/pastas/pastas/issues/801, so this just cleans up the semantics.

pastas/model.py
martinvonk: The problem I have with using a version like 3.0 is that is not formally in sight anywhere. So it could still be years and years till this code could be removed. While with using N minor releases at least we keep sight of when it will be removed.

martinvonk:
On a related note, thinking about the future: we might want to consider dropping our custom decorator entirely at some point. Since Python 3.13, [PEP 702](https://peps.python.org/pep-0702/) introduced a built-in [`@deprecated` decorator](https://typing.python.org/en/latest/spec/directives.html#deprecated) to the standard library that handles this exact workflow natively. Adopting that formal approach could be a great way to simplify our codebase eventually.

rubencalje:
Serious Python libraries do not use raise DeprecationWarning. They use warnings.warn(..., DeprecationWarning) and only raise a normal exception once the feature is actually removed. So that is what we should be doing in these decorators.

martinvonk:
> Serious Python libraries do not use raise DeprecationWarning. They use warnings.warn(..., DeprecationWarning) and only raise a normal exception once the feature is actually removed. So that is what we should be doing in these decorators.

Good thing we're not a serious Python library then :)

On a more serious note: I'm fine with how it is now. I'm fine with how the new proposal in this PR. I'm fine with how serious python packages do it. I'm fine with going with the [PEP 702](https://peps.python.org/pep-0702/) format, which can also work for Python<3.13 via typing-extensions. 

I only made this PR after the discussion in the last Pastas meeting where @raoulcollenteur mentioned that `remove_version` was ambiguous and it was not entirely clear when to actually remove the code. But I can live with any solution.

rubencalje:
I would propose to keep only one version in the decorator (`remove_version`), with a high major version (3.0), use warnings.warn for versions before that version, and raise a ValueError after that version.

martinvonk:
> I would propose to keep only one version in the decorator (remove_version), with a high major version (3.0), use warnings.warn for versions before that version, and raise a ValueError after that version.

I agree this makes sense. However, I'd propose we merge this now and revisit the deprecation strategy for Pastas 2.0. We can use this PR to standardize deprecations for the 1.14.0 release, consistent and clear with how we've handled it for the past two years. Then we can implement the simpler approach you proposed when we transition to Pastas 2.0 onwards. Changing this now and bumping all the deprecations for Pastas 2.0 that have been there for a few years to Pastas 3.0 (which might take another few years) would be unnecessarily cumbersome.

So I'd say: merge this PR and work with what we have for now, and revisit the deprecation strategy after the 2.0 transition (in the `dev2`).

rubencalje:
I would use a `remove_version` of 3.0 only for new deprecations, but 2.2 is fine by me as well. We can remove the deprecated noisemodel-functionality in one of the next versions after 2.0 (see second comment below) , because we have announced this for quite a while. The points I was making are:
- use `warnings.warn` instead of `logger.warning` (or instead of `raise DeprecationWarning`), which is a minor change, but more in line with other python-packages.
- just 1 `remove_version`, not a `deprecate_version` and a `remove_version`, so I want to keep it like it is now. We can agree to completely remove the functionality after one or two years. For example at version 2.2 we can agree to remove all functionality with a `remove_version` of 2.0. At the moment, we do not know when version 2.2 will be released, so there is no point in defining beforehand at what version we will remove certain functionality.
- If we really want to raise an error after `remove_version` (instead of using a warning), I would raise a ValueError with an informative massage ('xxx is removed since version xxx' and not 'module not found')

Let's discuss this tomorrow in person, I think we will agree faster then.

martinvonk:
Updated as per requested by @rubencalje and discussed with @dbrakenhoff and @rubencalje. Only question now is do we merge this in 1.14 or 2.0?

raoulcollenteur (approved):

raoulcollenteur:
I suggest to merge this into 1.14. Looks good, thanks for the changes @martinvonk.

rubencalje (approved):
Great, looks good! I would also merge in 1.14.

PR #1132, opened 2026-03-10
Add pas file testing

dbrakenhoff:
- add script to generate pas files that hopefully covers all known options
- script can be run with `uv --with "pastas==1.13.1"` to generate pas files for a specific pastas version
- add tests that load all generated pas files for a specific pastas version
- add github workflow specifically for pas file testing
- fix bug for loading ChangeModel

---

The following is a draft of a guide for the pas-file upgrade process:

# Developers guide for .pas files

Pastas models are serialised as `.pas` files (JSON under the hood).
[PastaStore](https://github.com/pastas/pastastore) stores these files on disk,
so a breaking change in the `.pas` format can "corrupt" a user's archive.

This document describes how we guard against that.

---

## Compatibility Table

| Pastas version | Min. readable `.pas` version | Notes |
|----------------|------------------------------|-------|
| 1.13.x         | 1.5                          | —     |
| 1.14.x         | 1.5                          | —     |
| 2.0.0          | 1.14                         | breaking — run migration first |

> **Rule:** every release must be able to read all `.pas` files written by the
> previous *minor* release.  A major version bump is the only time we allow
> dropping support for older formats, and even then a migration path must exist.

---

## Migration Path for Users

If you want to upgrade from pastas **X** to pastas **Y** and Y is a new major
version, you must first migrate your stored `.pas` files using the last minor
release before Y:

```
current pastas X.x  →  install pastas 1.14.x  →  run migration  →  install pastas 2.0
```

### Upgrading .pas files

**TODO:** suggestion by copilot, but seems pretty useful, either here or in pastastore or both

```python
import pastas as ps

ps.io.upgrade_pas_files(
    directory="path/to/your/pas/files",
    recursive=True,    # descend into sub-directories
    dry_run=False,     # set True to preview without writing
)
```

This re-serialises every `.pas` file in *directory* with the current version of
pastas, preserving the original as `<name>.pas.bak`.

---

## Developer Checklist

When you make a change that touches `.pas` file structure:

1. Update `_load_model` / `_load_stressmodel` (etc.) to accept **both** the old
   and new keys, with a deprecation fallback.
2. Add the new version string to the compatibility table above.
3. Update the previous version string in `PASTAS_VERSIONS` in
   `tests/test_pas_files.py` so CI generates files written by
   that version.
4. If a *major* version bump is planned, ensure the migration step in
  `pastas/io/migrate.py` is working as intended. This should be covered by the test sin
  step 3, but double check!

---

## How the Tests Work

`tests/generate_pas_files.py` uses `uv run --with pastas==<version>` to write
create a collection of `.pas` files covering every model component.  
`tests/test_pas_files.py` then loads each file with the *current* checkout of
pastas to verify nothing is broken.

The CI workflow `.github/workflows/test_pasfiles.yml` runs this automatically
on every push/PR to `dev` and `master`.

### DIFF
--- a/.github/workflows/test_pasfiles.yml
+++ b/.github/workflows/test_pasfiles.yml
@@ -24,6 +24,7 @@ jobs:
     name: Test suite for .pas files
     runs-on: ubuntu-latest
     env:
+      CI_PASFILES: "1"
       # Color Output
       # Rich (pip)
       FORCE_COLOR: 1
--- a/.github/workflows/test_pasfiles.yml
+++ b/.github/workflows/test_pasfiles.yml
@@ -1,0 +1,43 @@
+name: CI PASFILES IO
+
+on:
+  # Trigger the workflow on push or pull request on master,
+  # and only on pull-requests on dev branch
+  push:
+    branches:
+      - master
+      - dev
+  pull_request:
+    branches:
+      - master
+      - dev
+
+concurrency:
+  # Group by workflow name and ref (branch or pull request ref).
+  # For pushes the ref is the branch (refs/heads/<branch>).
+  # For PR events the ref is refs/pull/<number>/merge — same across commits for that PR.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test_pasfiles:
+    name: Test .pas files
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
+      - name: Install uv with Python 3.13
+        uses: astral-sh/setup-uv@v7
+        with:
+          enable-cache: true
+          python-version: "3.13"
+          activate-environment: true
+
+      - name: Install the project
+        run: |
+          uv sync --extra "dev"
+
+      - name: Run pytest
+        run: |
+          uv run pytest -m "pasfiles"
--- a/.github/workflows/test_pasfiles.yml
+++ b/.github/workflows/test_pasfiles.yml
@@ -40,6 +40,10 @@ jobs:
         run: |
           uv sync --extra "dev"
 
+      - name: Print versions
+        run: |
+          python -c "import pastas as ps; ps.show_versions(optional=True)"
+
       - name: Run pytest
         run: |
           uv run pytest -m "pasfiles"
--- a/.github/workflows/test_pasfiles.yml
+++ b/.github/workflows/test_pasfiles.yml
@@ -1,4 +1,4 @@
-name: CI PASFILES IO
+name: CI PASFILES
 
 on:
   # Trigger the workflow on push or pull request on master,
@@ -25,25 +25,30 @@ jobs:
     runs-on: ubuntu-latest
     env:
       CI_PASFILES: "1"
+      # Color Output
+      # Rich (pip)
+      FORCE_COLOR: 1
+      # Tox
+      PY_COLORS: 1
+      # Pytest
+      PYTEST_ADDOPTS: "--color=yes"
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6
 
-      - name: Install uv with Python 3.13
+      - name: Set up uv
         uses: astral-sh/setup-uv@v7
         with:
+          python-version: "3.11"
           enable-cache: true
-          python-version: "3.13"
-          activate-environment: true
 
-      - name: Install the project
+      - name: Install dependencies
         run: |
-          uv sync --extra "dev"
+          uv venv
+          uv pip install tox-uv
 
-      - name: Print versions
-        run: |
-          python -c "import pastas as ps; ps.show_versions(optional=True)"
+      - name: Setup tox environment
+        run: uv run tox -e pasfiles --notest
 
-      - name: Run pytest
-        run: |
-          uv run pytest -m "pasfiles"
+      - name: Test
+        run: uv run tox -e pasfiles --skip-pkg-install
--- a/.github/workflows/test_pasfiles.yml
+++ b/.github/workflows/test_pasfiles.yml
@@ -24,7 +24,6 @@ jobs:
     name: Test suite for .pas files
     runs-on: ubuntu-latest
     env:
-      CI_PASFILES: "1"
       # Color Output
       # Rich (pip)
       FORCE_COLOR: 1
--- a/.github/workflows/test_pasfiles.yml
+++ b/.github/workflows/test_pasfiles.yml
@@ -21,7 +21,7 @@ concurrency:
 
 jobs:
   test_pasfiles:
-    name: Test .pas files
+    name: Test suite for .pas files
     runs-on: ubuntu-latest
     env:
       CI_PASFILES: "1"
--- a/.github/workflows/test_pasfiles.yml
+++ b/.github/workflows/test_pasfiles.yml
@@ -24,7 +24,7 @@ jobs:
     name: Test suite for .pas files
     runs-on: ubuntu-latest
     env:
-      CI_PASFILES: "1"
+      CI_PASFILES: "1" # generate pas files for these tests
       # Color Output
       # Rich (pip)
       FORCE_COLOR: 1
--- a/.github/workflows/test_pasfiles.yml
+++ b/.github/workflows/test_pasfiles.yml
@@ -23,6 +23,8 @@ jobs:
   test_pasfiles:
     name: Test .pas files
     runs-on: ubuntu-latest
+    env:
+      CI_PASFILES: "1"
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6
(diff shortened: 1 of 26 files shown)


.github/workflows/test_pasfiles.yml
```diff
@@ -0,0 +1,45 @@
...
+          python-version: "3.13"
+          activate-environment: true
+
+      - name: Install the project
+        run: |
+          uv sync --extra "dev"
+
+      - name: Run pytest
+        run: |
+          uv run pytest -m "pasfiles"
```
martinvonk: I'd really like to keep testing this via `tox` and keep the definitions / overview of all different tests pyproject.toml via `tool.tox.env` attributes.. So that we can just run `tox -e pasfiles`

.github/workflows/test_pasfiles.yml
```diff
@@ -0,0 +1,45 @@
...
+    name: Test .pas files
+    runs-on: ubuntu-latest
+    env:
+      CI_PASFILES: "1"
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
+      - name: Install uv with Python 3.13
+        uses: astral-sh/setup-uv@v7
```
martinvonk, line 40: Maybe we should consider uv for all tests. That could speed things up significantly..

pastas/io/base.py
```diff
@@ -172,13 +172,13 @@ def _load_stressmodel(ts, data):
     # Create and add stress model
     stressmodel = getattr(ps.stressmodels, ts.pop("class"))
 
-    if "rfunc" in ts.keys():
-        rfunc_class = ts["rfunc"].pop("class")  # Determine response class
-        rfunc_up = ts["rfunc"].pop("up", None)  # get up value
-        rfunc_gsf = ts["rfunc"].pop("gain_scale_factor", None)  # get gain_scale_factor
-        rfunc = getattr(ps.rfunc, rfunc_class)(**ts["rfunc"])
+    for key in [k for k in ts if k in ("rfunc", "rfunc1", "rfunc2")]:
```
martinvonk, line 175: Just a question but what are rfunc1 and rfunc2? or can we just test if `[k for k in ts if "rfunc" in k]`?

.github/workflows/test_pasfiles.yml
dbrakenhoff: tox away, but I wanted to use uv for a simple way to run the pas file export script with the right pastas version :), seemed simple enough to use uv again after setting it up that way.

dbrakenhoff: And can you get it to run in 39s ;)?

tests/test_pas_files.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+PAS_FILES = (
+    [
+        p
+        for version in PASTAS_VERSIONS
+        for p in sorted(generate_pas_files(version).glob("*.pas"))
+    ]
+    if os.environ.get("CI_PASFILES", "0") == "1"
+    else []
```
martinvonk: Is this really necessary? because now I can't run the test locally.

martinvonk (approved):
Very nice. Only thing I have is that it'd be nice to run the tests locall which can't happen due to the `os.environ.get("CI_PASFILES", "0") == "1"`

.github/workflows/test_pasfiles.yml
martinvonk: done in #1133

pastas/io/base.py
dbrakenhoff: Yeah i was afraid it might trigger on something called rfunc_kwargs or the likes. So I ensured only names that refer to rfuncs in the pastas StressModels were used. rfunc1 and rfunc2 are the rfunc names in ChangeModel.

tests/test_pas_files.py
dbrakenhoff: Well it was, because uv wasn't available on the other testing suites before. Even though the tests aren't run by the other workflows, the pas-files were being generated in test discovery. This errored because uv wasn't available. I guess we have uv everywhere now, so I guess we could turn this off, but this means there is some unnecessary pas file generation on each other workflow... which also isn't ideal.

So I would just leave it and accept you have to run `pytest tests/test_pas_files.py` to do it locally. That isn't that big of a deal I would say :innocent:

martinvonk: Sorry I wasn't clear. I don't mind running it via pytest but when I do the tests are skipped instead of run. And I'm not 100% sure why.. But when I remove this statement it just works.

martinvonk:
```
(pastas) ~/repos/pastas$ uv run pytest tests/test_pas_files.py 
      Built pastas @ file:///repos/pastas
Uninstalled 1 package in 1ms
Installed 1 package in 1ms
Test session starts (platform: linux, Python 3.13.5, pytest 9.0.2, pytest-sugar 1.1.1)
rootdir: repos/pastas
configfile: pyproject.toml
plugins: sugar-1.1.1, cov-7.0.0, anyio-4.12.1
collected 1 item                                                                                                                                     

 tests/test_pas_files.py s                                                                                                             100% ██████████
================================================================= slowest durations ==================================================================

(2 durations < 0.005s hidden.  Use -vv to show these durations.)

Results (0.02s):
       1 skipped
(pastas) ~/repos/pastas$ pytest tests/test_pas_files.py 
Test session starts (platform: linux, Python 3.13.5, pytest 9.0.2, pytest-sugar 1.1.1)
rootdir: repos/pastas
configfile: pyproject.toml
plugins: sugar-1.1.1, cov-7.0.0, anyio-4.12.1
collected 1 item                                                                                                                                     

 tests/test_pas_files.py s                                                                                                             100% ██████████
================================================================= slowest durations ==================================================================

(2 durations < 0.005s hidden.  Use -vv to show these durations.)

Results (0.02s):
       1 skipped
(pastas) ~/repos/pastas$ pytest tests -m "pasfiles"
Test session starts (platform: linux, Python 3.13.5, pytest 9.0.2, pytest-sugar 1.1.1)
rootdir: repos/pastas
configfile: pyproject.toml
plugins: sugar-1.1.1, cov-7.0.0, anyio-4.12.1
collected 539 items / 538 deselected / 1 selected                                                                                                    

 tests/test_pas_files.py s                                                                                                             100% ██████████
================================================================= slowest durations ==================================================================

(2 durations < 0.005s hidden.  Use -vv to show these durations.)

Results (0.21s):
       1 skipped
     538 deselected
```

dbrakenhoff: Ah yes, very good point. So compiling the list of pas files is done on test discovery, so that's why it needs to be empty for the other test workflows. This is controlled by the environment variable. When it has the correct value it will create the pas files and then run the tests, otherwise the list is empty and it will run/skip an empty test. 

I think a neater solution is to default the list to True and add an env variable that switches off pas file generation in the other test workflows. I will fix it tomorrow.

martinvonk: I tried this myself within the pytest framework. I can now run it locally as expected. But the file did not become less complicated, but the environmental variable is no longer needed.. Feel free to remove the change I made if you think it too complicated compared to the environmental variable.

tests/test_pas_files.py
```diff
@@ -0,0 +1,97 @@
+"""Tests for loading .pas gallery files.
```
raoulcollenteur: Gallery files?

tests/test_pas_files.py
```diff
@@ -0,0 +1,97 @@
...
+        pas_files = _get_pas_files()
+        metafunc.parametrize(
+            "pas_file",
+            pas_files,
+            ids=[f"{p.parent.name}/{p.stem}" for p in pas_files],
+        )
+
+
+XFAIL = {
+    "ChangeModel.pas": "Known issue with ChangeModel in in <1.13.1",
```
raoulcollenteur: remove "in"

raoulcollenteur: Plus we only test 1.13.2 right, so is this line necessary?

raoulcollenteur:
Looking good, let's hope it is easy to maintain, but it does look like that. 

I like the suggestion ps.io.upgrade_pas_file, let's think about that in another issues and/or PR.

raoulcollenteur (approved):

tests/test_pas_files.py
dbrakenhoff: I was chatting to copilot and may have mentioned creating a "gallery" of pas files and it got very enthusiastic about the term :P. I removed some of the mentions, but not all it turns out.

tests/test_pas_files.py
dbrakenhoff: Should be 1.13.2 and yes this is currently necessary to avoid failing tests for 1.14.

tests/test_pas_files.py
dbrakenhoff: I'm impressed you got it to work, but I think just defaulting to True, and turning the tests off with the environment variable in the other workflows is a simpler solution that I will still understand in the future :)

martinvonk: I was impressed as well. But totaly fine to ignore it and go for this approach :)

PR #1145, opened 2026-03-27
Replace internal Model checks with centralized check utilities

### LINKED ISSUE #1144: Use check module for internal Model check methods
### Describe the task
Currently, when generating warnings after a model is optimized (e.g., via _generate_warnings_report / fit_report), the codebase relies on the internal methods `_check_response_tmax` and `_check_parameter_bounds` to determine if the response time exceeds the calibration period and if any parameters are on the bounds. However, the pastas.check module already implements fully-featured methods that do exactly this. 

### Proposed changes
Replace the usages of _check_response_tmax within the Model warning generation with check.response_memory and check.response_memory_vs_warmup. Remove the redundant _check_response_tmax method entirely to avoid code duplication. This centralizes all check and warning logic in the check.py module. and reduces code duplication.

martinvonk:
Refactors model warning checks to use centralized utilities instead of duplicated internal logic.

### DIFF
--- a/pastas/check.py
+++ b/pastas/check.py
@@ -473,13 +473,36 @@ def parameter_bounds(ml: Model, parameters: list[str] | str | None = None):
     elif isinstance(parameters, str):
         parameters = [iparam for iparam in ml.parameters.index if parameters in iparam]
     df = get_empty_check_dataframe()
-    upper, lower = ml._check_parameters_bounds()
+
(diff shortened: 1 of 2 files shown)


martinvonk:
Ideally this is merged before I continue working on #1031. So I can start moving stuff for the fit report to the dedicated solvers.

dbrakenhoff (approved):

PR #1149, opened 2026-04-13
update weights docstring and tests

### LINKED ISSUE #990: Multiply the jacobian with the weights for the covariance matrix
### Describe the question/bug
I don't know if this should be an issue or a discussion, but I have now labeled this as a question. If the answer to the question is yes, this should be a bug/feature. 

For the `LeastSquares` and `LmfitSolve`, weights can be supplied to the objective function (`self.objfunction`, which calls `self.misfit`).  The covariance matrix for the `LeastSquares` solver, using `scipy.optimize.least_squares`, is computed in `pastas.LeastSquares.get_covariances` using some code copied from SciPy. The `LmfitSolve` solver (that wraps `scipy.optimize.leastsq` by default) has the keyword argument `calc_covar: bool = True` that is to be supplied to `lmfit.Minimizer.minimize`, which uses the `numdifftools` package to compute the Hessian.

For both LeastSquares and Lmfitsolve, weights are not taken into account when computing the covariance matrix. So two questions:
1. Shouldn't this be the case? With: $Cov(p) =( J^T W J)^{-1} * s^2$, and not as currently used $Cov(p) = (J^T J)^{-1} * s^2$
2. How about the `ml.noise_weights`, should those be treated the same as the "normal" weights?

### Expected behavior (I think)

```python
# result = scipy.optimize.OptimizeResult class
jacobian = result.jac
cost = 2 * result.cost  # res.cost is half sum of squares!

nobs, npar = jacobian.shape
s_sq = cost / (nobs - npar)  # variance of the residuals
W = numpy.diag(self.weights)
pcov = numpy.inv(jacobian.T @ W @ jacobian) * s_sq
```
SciPy uses some fancy tricks (that we copied) to get the inverse matrix, with the singular value decomposition or the lapack, `trtri` function, but the pcov in the end should be same in most scenarios.

rubencalje:
# Short description
Initially, I thought we currently use a nonstandard interpretation of user-defined weights in Pastas. Conceptually, assigning a weight of 2 to a measurement should make its contribution to the objective function twice as large. However, because all solvers use a least-squares formulation, the weights are effectively squared. As a result, a measurement with weight 2 contributes four times as much to the objective function instead of the intended factor of two. 

Initially this PR corrected that behavior by applying the square root of the weights to the residuals, ensuring that the minimized objective function becomes:

$\sum_i w_i \, r_i^2$

However, after discussion with the Pastas-team, it turned out both PEST and scipy mostly use the same weights as pastas currently does, where a double weight has four times as much influence. This is why we will keep the weights as they are. This PR improves the docstrings and add tests, to better inform the user of the way weights work in Pastas.

- [ ] closes issue #xxxx
- [ ] is documented
- [ ] type hints
- [ ] tests added or expanded


martinvonk:
I think this is a good change. But I think that if we start using the weights not as calibration period indexer, we should think about how weights are properly parsed, especially when computing the parameter covariance matrix: https://github.com/pastas/pastas/issues/990

martinvonk (approved):

raoulcollenteur:
> I think this is a good change. But I think that if we start using the weights not as calibration period indexer, we should think about how weights are properly parsed, especially when computing the parameter covariance matrix: #990

I agree with the above. Before merging this, perhaps we should think about this a bit more and clarify the purpose of the weights as discussed in the last Pastas meeting.

rubencalje:
As discussed yesterday, both PEST and scipy mostly use the same weights as pastsas currently does, where a double weight has four times as much influence. This is why we will keep the weights as they are.

However, it is good to add a notebook, where we demonstrate this, so the user knows he has to supply the square root of the weights themselves, if they want to have a linear influence of weights.

Also, the added tests should be edited so it passes for the weighting scheme that is implemented in pastas.

martinvonk:
I reverted the changes that takes the square root. I updated the test and docstring accordinlgy. I think we should still merge this because the tests are nice checks for our own sanity. And the docstring is much more clear now. Good to have till we get a better implementation of the weights.

PR #1154, opened 2026-04-14
Add solve strategies to Model.solve

rubencalje:
Introduce an optional strategy parameter to `Model.solve` to support two-step solving workflows: 

- "deterministic_then_noise_only"
- "deterministic_then_full".

Both require a noisemodel and temporarily adjust parameters to first fit deterministic components, then handle noise (either only `noise_alpha`/`noise_beta`, or continue to a full solve). The implementation copies and restores parameter states, adjusts which parameters vary, and accumulates nfev across iterations. Added validation and error messages for missing or unknown strategies. Also removed an extra blank line in BaseSolver and added tests covering the new strategies.

# Short description
Add a short description describing the pull request here.

- [ ] closes issue #xxxx
- [ ] is documented
- [ ] type hints
- [ ] tests added or expanded

### DIFF
--- a/pastas/model.py
+++ b/pastas/model.py
@@ -1041,19 +1041,22 @@ class Model:
                     raise ValueError(msg)
                 # copy the existing parameters, so we can set initial and vary back later
                 parameters = self.parameters.copy()
+                noise_parameters = parameters.index[
+                    parameters["name"] == self.noisemodel.name
+                ]
                 # fit model without noisemodel first
-                self.set_parameter("noise_alpha", vary=False)
+                self._parameters.loc[noise_parameters, "vary"] = False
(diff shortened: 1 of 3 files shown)


rubencalje:
As discussed yesterday, this implementation of the solve-strategy is not desired.

It may be better to implement this in the solver-classses, or add separate methods, for example `ps.solve.deterministic_then_noise_only(ml, **kwargs)` or `ps.solve.deterministic_then_full(ml, **kwargs)`.

Let's first discuss this in an issue.

martinvonk:
We should consider if `fit_constant=False` is not a "solve strategy" as well, and if we should deal with it like that in a similar way.

PR #1155, opened 2026-04-14
add verbose option to validate_series

OnnoEbbens:
This PR adds a verbose option to the `validate_series` function. This will print a ✅ or ❌ with a message for every check performed by `validate_series`.

If the verbose option is used no errors are raised. Actually, I think this should be the default because when you raise an error for every check you cannot see the results of the other checks if one fails.

```python
>>> ps.validate_oseries(valid_series, verbose=True)

✅ timeseries is a pandas series
✅ timeseries values are floats
✅ timeseries index is a DatetimeIndex
✅ timeseries index is datetime64
✅ timeseries index has no NaNs
✅ timeseries index is monotonically increasing
✅ timeseries index has no duplicate indices
✅ timeseries values have no nan-values


>>> ps.validate_oseries(invalid_series, verbose=True)
✅ timeseries is a pandas series
✅ timeseries values are floats
❌ Index of series Head_340521 is not a pandas.DatetimeIndex.
❌ Indices of series Head_340521 are not datetime64.
✅ timeseries index has no NaNs
✅ timeseries index is monotonically increasing
✅ timeseries index has no duplicate indices
❌ The Time Series 'Head_340521' has nan-values. Pastas will use the fill_nan settings to fill up the nan-values.
```


martinvonk:
Very nice change. I felt the freedom to make some changes as well because we noticed during the course that the docstrings weren't matching to what was actually happening (e.g. NaT check was missing for the series). I also improved the clarity of some of the messages.

martinvonk:
maybe we can just turn of the cyclomatic complexity in codacy? @raoulcollenteur @dbrakenhoff ?
`Method _validate_series has a cyclomatic complexity of 34 (limit is 8)`

martinvonk (approved):

rubencalje (approved):

PR #1156, opened 2026-04-14
Move calculation of responses from modelplots to stressmodels

### LINKED ISSUE #1199: Move calculation of responses from modelplot to stressmodel
@rubencalje proposed to move the computation of the response function from modelplot.py to the individual stressmodels. See the original code here:
 https://github.com/pastas/pastas/blob/bf4010dc3b2c00961ca86a210577a2df54f8bd78/pastas/plotting/modelplots.py#L534

This would improve the internal method `_plot_response_in_results` method and affect `ml.plots.results` and `ml.plots.results_mosaic`.

Methods like `_get_responses`, `_get_all_responses` or similar should be added to at least `ChangeModel` and `TarsoModel`, perhaps also to `WellModel`. The first two models actually do not work with istress, because it is not depending on the number of stresses if there are multiple responses, but a characteristic of the stressmodel that these work with two response functions.

### LINKED ISSUE #1200: Show all response functions in `ml.plots.results` and `ml.plots.results_mosaic`

For some stressmodels and configurations, the stressmodels internally work with multiple different response functions: `ChangeModel`, `TarsoModel`, `WellModel`, and `RechargeModel(recharge=Linear)`.

In the exception of the `Linear` recharge model, different response functions can be calculated for the precipitation and potential evaporation.

It would be nice if it would be possible to show all the responses in these plotting methods, for example, by providing an argument `all_responses`.

A tricky thing is how to figure out how many responses a stressmodel (can) have. Perhaps each stressmodel just needs information about that, by adding a `self.nrfuncs` attribute? 

i.e., `nrfuncs` depends on:
- The recharge model in `RechargeModel`, 2 if `Linear`, 1 otherwise. self.nsplit could work
- number of pumping wells in `WellModel`, len(self.stresses) could work
- the model structure (Changemodel and TarsoModel just have two defined)
- `StepModel` and `TrendModel` actually have 0.

If a keyword `all_responses` is added, the default would have to be determined (True, False, or None -> stressmodel dependent).

### Expected usage
A clear and concise description of what you expect to happen.

```python
`ml.plots.results(all_responses=True)`
```

### Additional context
#1198 , #1199

### LINKED ISSUE #1202: Argument `split` available in plots.results but not `plots.results_mosiac`

The plotting method `results` supports `split` but the similar method `results_mosiac` does not. It would be nice if these methods have a similar interface. 

### Expected usage
A clear and concise description of what you expect to happen.

```python
`plots.results_mosiac(split=True)`
```
should be possible

### LINKED ISSUE #1208: Rename argument `split` to `split_contribution` to be more explicit
In PR #1156 it is proposed to rename the argument `split` to `split_contributions` to be more explicit about what is done.

rubencalje:
# Short description
This PR makes the plotting of responses in the results-plot more general, by moving the calculation of the responses to the stressmodels. In this way, it is also easier to add your own stressmodel, and plot the reponse in the results-plot.

The user will probably notice the change in this PR because it now also plots the response to evaporation in the Linear Recharge model (see below), just like the two responses in the Tarso model. For the WellModel, all responses are plotted. This PR handles the keyword argument `split=True` better than before, where it would for example not work for WellModel.

image

This PR adds a method `get_responses` to each stressmodel-class, which for `StressModelBase` is as follows:

```
def get_responses(self, ml, block_or_step="step", istress=None) -> list[Series]:
        responses = [
            ml._get_response(
                block_or_step=block_or_step,
                name=self.name,
                add_0=True,
                istress=istress,
            )
        ]
        return responses
```

- [x] closes issue 
- [x] #1199, 
- [x] #1200, 
- [x] #1201, 
- [x] #1202, 
- [x] #1208
- [x] is documented
- [x] type hints
- [x] tests added or expanded


rubencalje:
As discussed yesterday, the moving of the response calculation is a good idea. There were some questions about requesting the response from the stressmodel, which ask the response from the model, which lets the stressmodel calculate the response again. However, I think the current approach is defendable, as the model contains the parameters, and so needs to be in between.

Showing both responses of the RechargeModel can raise questions. Maybe we can add an option 'only_show_first_response`, with a default value of True. And if this is False, and multiple responses are shown, to not plot the first response in blue, but in another color.

TomHottentot:
I believe the moving of the response calculation is a good idea. 

In your example I believe it would be better to add a small legend instead of the title "Step Response" similar to `ml.plots.results_mosaic()`. For a RechargeModel it should make sense which is which, bit a WellModel it might get less intuitive. Or is it better to stick to `stacked_results()` with WellModels?

rubencalje:
Before merging this PR, we have the following options in to plot the responses of stressmodels with multiple responses (`RechargeModel` with `recharge=Linear`, `TarsoModel`, `WellModel`, `ChangeModel`) in `ml.plots.results()` and `ml.plots.results_mosaic()`:

1. Merge the PR unchanged: always plot all responses.
2. Introduce an `all_responses=False` option, where only the first response is displayed (for precipitation in the case of RechargeModel), unless `all_responses=True` is explicitly enabled.
3. Same as option 2, but with `all_responses=True` as the default.
4. Never render multiple responses, and only support `split=True`.

I am in favor of option 3, but can live with option 1 or 2.

raoulcollenteur:
Generally looks good I think, two issues to fix before making a choice on the options:

- `split` argument is not available in results_mosaic, but probably should be?
- `ml.plots.results(split=False)` does not work and still shows all responses.

Regarding the options. I am in strong favor of option 4 for the Rechargemodel after adding split to `results_mosaic` as well, because a) this would not change the default behavior and plot as it has always been in Pastas for Linear recharge, the response to precipitation excess, and b) it otherwise complicates explaining this during Pastas courses and in the examples on the docs where you have to explain where the second response comes from because it is the response to the raw potential recharge and not f*E, plus the fact that it is not used for computation and cannot be computed for other recharge models.

But for TarsoModel and WellModel I favor option 4 with `split=True` as the default. So what about moving the default to the default value of the `get_response` method of the stress model and use None in the plotting? That way we can do both behaviors described above :)

If not option 4, many examples need to be adapted to explain what response function is being shown because this will be entirely unclear to the users otherwise, or adapted by applying split=False. 

Adding a keyword `all_responses` to the plotting functions does not seem necessary when `split` is already there, right? 

Probably the others @pastas/dev should weigh in as well here.

dbrakenhoff:
I agree that showing multiple responses for HantushWellModel, Tarso and DoubleExponential is nice, and in my opinion should also be the default setting. This makes it super clear what is being used in the computation and allows you to compare those responses. So those responses should all be shown on the same plot. 

I also agree that showing the response to evaporation for RechargeModels is awkward and more difficult to explain. We tell people we're making a linear estimate of recharge and then convolving that with a response function, so there suddenly being 2 responses is a bit confusing, even though its possible to compute the theoretical response to evaporation. 

I also like split=True to separate out the different contributions within the HantushWellModel, RechargeModel, etc. into separate plots. So in my mind there are two different options we'd both like to support, showing all responses in a single plot (show_all_responses). Separating out the individual contributions within one StressModel into separate charts (split). 

So I vote option 3, but not for RechargeModel ... :man_facepalming: 

And I can live with option 2 as well.

raoulcollenteur:
Ah now I get it, then I also favor also option 3 but not for recharge. What about `split_responses` as keyword name?

martinvonk:
I'm in favor of option 3, but can live with 1 and 2. I don't mind the linear RechargeModel having two responses. Whether we do convolution with R on one response, or with P and E on two responses of which the latter one scaled with `f` does not matter. I think this might even clarify things a bit for courses.

I also like the name `split_responses`. Maybe we can also rename `split` to `split_contributions` such that it is clear what splits what :)

rubencalje:
> Ah now I get it, then I also favor also option 3 but not for recharge. What about `split_responses` as keyword name?

I do not think `split_responses` is a good description of the feature, because it does not split the responses. It shows all responses when it is True, and only the first one when it is false. Therefore I think `all_responses` is better, or maybe we can use the inverse `only_show_first_response`.

We do not use the response to evaporation for convolution, but you can see the Linear recharge as a system with two responses, one for precipitation and one for evaporation, and you can split the contribution using the `split` argument. So I am in favor of showing both responses as well when `all_reponses` is added as an argument and set to True. This will help the user to understand how the Linear recharge-model works.

I am not in favor of showing just the first response for some stressmodels, and showing all responses for others. But if we do want this, maybe we can add an extra boolean argument `split` to the `__init__` of  `RechargeModel`, to turn on splitting the contribution and showing multiple responses, so I can activate this feature for `RechargeModel` again.

I think `split_contributions` is a better name than `split`, but we need to add a deprecationwarning then. I am currently working on adding support for `split` in `results_mosaic` as well.

dbrakenhoff:
I am in favor of renaming `split` to `split_contributions` with a deprecation warning. As for the other argument name options:

- `plot_all_responses`/`show_all_responses` 
- `show_only_first_response`
- `all_responses`

I think I prefer the last one. 

> I am not in favor of showing just the first response for some stressmodels, and showing all responses for others. But if we do want this, maybe we can add an extra boolean argument split to the __init__ of RechargeModel, to turn on splitting the contribution and showing multiple responses, so I can activate this feature for RechargeModel again.

Yes this is annoying I agree, but from another perspective it could make sense to only show the step responses that were used in the convolution step in pastas. But as you say, doing the convolution separately is mathematically equivalent and since we want to support the splitting of the contributions, it also makes sense to allow the responses to be displayed separately. I also agree with Raoul this will require maybe an extra step for beginners to explain, but it sounds doable. So you've convinced me, I can live with the two responses for RechargeModel as well :).

raoulcollenteur:
I prefer `all_responses` as well I think. 

I still strongly prefer not to show two responses for the linear model and suggest to set `all_responses=None` by default, and then just checking if there is a Linear recharge model and showing one response (the one that is used in computation, and has been shown for the entire existence of Pastas by default), and showing all responses for the other stressmodels. 

(I thought of using a call to `sm.get_stress()` (or using the existing call :) ) to figure out the number of response functions to plot, but unfortunately that doesn't work for Tarso..)

@mbakker7 what would you prefer?

martinvonk (commented):
Here it is still called split. Which is logical I guess but makes me second guess if we should rename split to split_contributions...

pastas/plotting/modelplots.py
```diff
@@ -190,7 +204,10 @@ def results(
         sim = self.ml.simulate(tmin=tmin, tmax=tmax, return_warmup=return_warmup)
         res = self.ml.residuals(tmin=tmin, tmax=tmax)
         contribs = self.ml.get_contributions(
-            split=split, tmin=tmin, tmax=tmax, return_warmup=return_warmup
+            split=split_contributions,
```
martinvonk, line 207: Here it is still called split?

pastas/plotting/plotutil.py
```diff
@@ -65,7 +65,7 @@ def _get_height_ratios(ylims: list[tuple[float, float]]) -> list[float]:
     return [0.0 if np.isnan(ylim[1] - ylim[0]) else ylim[1] - ylim[0] for ylim in ylims]
 
 
-def _get_stress_series(ml, split: bool = True) -> list[Series]:
+def _get_stress_series(ml: Model, split: bool = True) -> list[Series]:
```
martinvonk, line 68: Here it is still called split. Which is logical I guess but makes me second guess if we should rename split to split_contributions...

martinvonk: Same goes for `ml.plots.series`.. What does split do there?

pastas/plotting/modelplots.py
```diff
@@ -533,76 +576,22 @@ def results_mosaic(
...
-                    name=sm_name,
-                    add_0=True,
-                    istress=istress,
-                    **rkwargs,
-                )
-            ]
-
+        responses = sm.get_responses(
+            self.ml, block_or_step=block_or_step, istress=istress
+        )
```
martinvonk: This is much cleaner now. Nice.

pastas/plotting/plotutil.py
martinvonk: If we're not gonna name, what is now `all_responses`, `split_responses`. Then I don't mind keeping the `split` keyword argument name instead of renaming to `split_contributions` .

pastas/stressmodels.py
```diff
@@ -210,9 +209,37 @@ def get_stress(
...
+        -----
+        For the individual options for the different settings please refer to the
+        docstring from the TimeSeries.update_series() method.
+
+        See Also
+        --------
+        ps.timeseries.TimeSeries.update_series
+        """
+        for stress in self.stresses:
+            stress.update_series(freq=freq, tmin=tmin, tmax=tmax)
```
martinvonk, line 243: Nice that this works!

martinvonk (approved):

pastas/plotting/modelplots.py
dbrakenhoff: I think the term `split` in ml.get_contributions is pretty clear, whereas in plots it is more ambiguous what it could mean. So I think this is fine.

rubencalje:
> I prefer `all_responses` as well I think.
> 
> I still strongly prefer not to show two responses for the linear model and suggest to set `all_responses=None` by default, and then just checking if there is a Linear recharge model and showing one response (the one that is used in computation, and has been shown for the entire existence of Pastas by default), and showing all responses for the other stressmodels.
> 
> (I thought of using a call to `sm.get_stress()` (or using the existing call :) ) to figure out the number of response functions to plot, but unfortunately that doesn't work for Tarso..)
> 
> @mbakker7 what would you prefer?

As a solution to this standoff we can also set `all_responses` to `False` by default, so most of the functionality is left unchanged, unless the user explicitly sets `all_responses` to True.

raoulcollenteur:
I took a closer look at the implementation and I think this can be done a bit cleaner, while solving some other problems:

I think we should add a method to the Model class _get_responses directly and call that from the plots rather than this circular implementation now. That way, we can also solve the problem that some stressmodels (TarsoModel and ChangeModel) need a special method to compute the response. Right now using ml._get_response with an istress actually doesn't work for those stressmodels, so that's another issue that can be solved then. All the other stressmodels can make use of the exact same logic.

Apart from that the new .get_response functions are public but undocumentated. So they should either become private, or get documented. 

I'll make this implementation and then I think everyone is gonna be happy and it will do the same :)

raoulcollenteur:
I restructured quite a bit in the last few commits, but the functionality is still the same. By default, for most stressmodels all response functions are shown, but not for the RechargeModel. 

I added a new method `_get_all_response`  to Model, and adapted `_get_response` to allow to use custom _get_response methods from the stressmodel if any exist.

@rubencalje let me know what you think of this implementation :)

@dbrakenhoff , should WellModel actually not also have a custom _get_response function that uses `r` ? Right now (and before, see [here](https://pastas--1156.org.readthedocs.build/1156/examples/multiple_wells.html)) the response for the different pumping wells were still the same.. But before this PR they are different (see stable docs). That is still something to look into I think, probably to do with #1173 from @martinvonk?

rubencalje:
I still do not really see why it is a problem to show both responses for the Linear RechargeModel, and I thought @dbrakenhoff and @martinvonk agreed. The whole goal of this PR was that we were not checking the type of stressmodel in the plotting-module anymore. But if the others agree this is a good change, I am ok.

Then the second point: I would keep the method `get_responses`  (and not `get_response`) in each stressmodel, where you get all responses if `istress` is None. The model-method `get_responses` can then just call the stressmodels `get_responses`, which is more clear. Besides, the new implementation is slower, as some calculations are repeated. We should make sure `istress` in `get_responses` works for all stressmodels and that there is a docstring, but that is no reason to change the whole setup.

pastas/plotting/plotutil.py
raoulcollenteur: Here you split the stresses, so otherwise `split_stresses` would make more sense, but I think this can stay as is.

rubencalje:
> @dbrakenhoff , should WellModel actually not also have a custom _get_response function that uses `r` ? Right now (and before, see [here](https://pastas--1156.org.readthedocs.build/1156/examples/multiple_wells.html)) the response for the different pumping wells were still the same.. But before this PR they are different (see stable docs). That is still something to look into I think, probably to do with #1173 from @martinvonk?

The response for the different pumping wells was shown earlier in this PR, in the notebook you linked to. Your link points to the latest version of the documentation in this PR, so it already reflects the results of your changes.

raoulcollenteur:
Perhaps we have to discuss the default for which response functions we show in a Pastas meeting and park that issue for now? Then after we decide, we can quickly implement that and go ahead. 

Before that, perhaps we can determine here on the technical implementation. But before that it should be clear what issues this PR actually tries to solve now, so I made the following issues:

- #1199
- #1198
- #1200
- #1201
- #1202
- #1203 
- #1207 
- #1208 
- #1165

@rubencalje the issue with the different pumping wells happened after I updated this branch from dev, and thus is a new issue created on the dev branch that we can probably fix in this PR as well? See https://pastas.readthedocs.io/latest/examples/multiple_wells.html

rubencalje:
Maybe it is a good idea then to revert your changes and implement those in another PR?

raoulcollenteur:
> Maybe it is a good idea then to revert your changes and implement those in another PR?

Yes I think that is a good idea. I will do that. I made separate issue for everything this PR has turned out to be trying to fix (so far I identified 6), which I think is just too many. Perhaps the quickest way is to make small PRs for each of these issues? 

Still, what started with 'just' #1199 helped to find many more issues, so I guess that's a good thing 😅

Let me know how you want to proceed, I would not be against separate PRs for each issue. That would make the process much easier I think and we can just copy the code from here?

raoulcollenteur:
I added issues for all the bugs and features this PR was/is trying to solve to make sure all the changes will end up in the release notes and are documented. See the overview in my comment above. 

My proposal is to solve the following issues in this PR:

- #1199 but only with private methods (i.e., `sm_get_responses`). Implementing new public methods is not necessary and can be a step after this PR after making an issue about this. 
- #1200 but with the default to False. That way there is no change in behavior yet, we can discuss that in another issue (does not yet exist) and PR, preferably after a Pastas meeting.
- #1201
- #1202 
- #1208

The issues #1165, #1198, #1203, and #1207 are better to do outside this PR I think. I think an argument can be made to separate #1208 as well but for me it is ok in here.

What do you @pastas/core all think? I can make the changes if we agree on this, should be quick.

raoulcollenteur (approved):

PR #1173, opened 2026-05-01
Add option to use the impulse response function for convolution

### LINKED ISSUE #475: [ENHANCEMENT]Remove Edelman response function
The Edelman response function has been in the code for a while, but was never operational as it was not imported upon regular import of `pastas`. 

The Edelman response function has problems, as it takes infinitely long before the step response reaches 1, and as a result it takes very long for it to reach the cutoff. Almost the same response can be obtained with the Polder function, which reaches 1 within a finite time.

I propose to remove the Edelman response function from the code base.

### LINKED ISSUE #1172: Replace block response by impulse response for some functions
Some of the step response functions (and hence the block response functions) are difficult to compute (especially the four-parameter function but also the Hantush function). They can easily be replaced by the impulse response function centered at the time interval. I expect this will give pretty much the same result and will be very fast.

### LINKED ISSUE #1185: Latexify archived
### Describe the task
The [latexify-py repository](https://github.com/google/latexify_py) was archived by the owner on Apr 18, 2026. It is now read-only. 

Maybe we should consider dropping support for it.. The last release was in 2024. It was a nice feature to have but not necessary, Also, now with large language models one can pretty easily get the latex code from python code. Only we can't show it in the notebooks anymore then.

martinvonk:
This pull request refactors and extends the response function (`rfunc`) classes in `pastas/rfunc.py` to add flexibility in how block responses are computed

I added a new `use_impulse` parameter to `RfuncBase` and all subclasses (`Gamma`, `Exponential`, `Hantush` etc.), allowing users to specify whether the block response should be computed or the impulse response at the middle of the dt time interval. The `block` method then calls `block_from_step` or `block_from_impulse` accordingly.

The notebook shows all the results. There are also tests available to compare the outcomes.


martinvonk:
This is still in draft. Need to merge #1134 first.

martinvonk:
From Pastas meeting
- merge #1134
- rename `step_or_impulse` setting to: `use_impulse: bool = False`
- fix FourParam

martinvonk:
Here's a summary of all the changes in this PR:

1. `RfuncBase` is now an Abstract Base Class.
2. New `use_impulse` parameter added to `RfuncBase.__init__` and propagated to all response function subclasses (`Gamma`, `Exponential`, `HantushWellModel`, etc.). When `True`, the "block response" is computed by integrating the impulse response instead of differencing the step response. Therefor, the `block()` method is refactored. It now dispatches to either:
   - `block_from_step()`: original behavior (difference of step response)
   - `block_from_impulse()`: new method: evaluates `impulse(t_mid, p) * dt`
3. Make `_name` a property
4. Make `nparam` a property
5. Update the `to_dict` methods accordingly
6. Update all the docstrings accordingly and make None behavior, and other keyword arguments, more clear

pastas/rfunc.py
```diff
@@ -263,84 +231,252 @@ def impulse(t: ArrayLike, p: ArrayLike) -> ArrayLike:
...
+            tmax is the minimum of the tmax determined from the cutoff and maxtmax.
+
+        Returns
+        -------
+        b: array_like
+            Block response values.
+        """
+        t = self.get_t(p=p, dt=dt, cutoff=cutoff, maxtmax=maxtmax, **kwargs)
+        t_mid = t - (0.5 * dt)  # compute times at the middle of the interval
+        return self.impulse(t_mid, p) * dt
```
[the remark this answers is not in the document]
martinvonk: Fixed in Model._get_response

martinvonk:
@rubencalje said he wanted to review this. Maybe you want to have a look as well @mbakker7.

martinvonk:
Maybe you also want to have a look @dbrakenhoff, specifically at the HantushWellModel impulse response function.

pastas/rfunc.py
```diff
@@ -143,73 +122,74 @@ def get_init_parameters(self, name: str) -> DataFrame:
             The initial parameters and parameter bounds used by the solver.
         """
 
+    @abstractmethod
     def get_tmax(self, p: ArrayLike, cutoff: float | None = None) -> float:
-        """Method to get the response time for a certain cutoff.
+        """Method to get the response time for a certain cutoff. A cutoff of
+        0.999 returns a tmax when 99.9% of the step response has effectuated.
```
dbrakenhoff: Nicer to put the second sentence in a new paragraph. Also "effectuated" is a very fancy word, I would use simpler language: " A cutoff of 0.999 returns the time when the step response has reached 99.9% of its upper limit. " Or something like that.

pastas/rfunc.py
```diff
@@ -251,117 +231,263 @@ def impulse(t: ArrayLike, p: ArrayLike) -> ArrayLike:
...
+            Fraction of the step response after which the response is truncated.
 
         Notes
         -----
-        The impulse response function for each response function class can be viewed on
-        the Documentation website or using `latexify` by running the following code in a
-        Jupyter notebook environment::
-
-            ps.RfuncName.impulse
+        Only change the settings if values are provided!
```
dbrakenhoff: What does this mean?

pastas/rfunc.py
```diff
@@ -251,117 +231,263 @@ def impulse(t: ArrayLike, p: ArrayLike) -> ArrayLike:
...
+            Response function parameters.
+        dt: float | ArrayLike
+            Time step in days. If an array_like is provided, it is returned as is
+            and the cutoff and maxtmax parameters are ignored. If a float is
+            provided, the times are computed from dt until the tmax determined
+            from the cutoff and maxtmax parameters.
+        cutoff: float | None, optional
+            Fraction of the step response used to determine the response cutoff.
+        maxtmax: float | None, optional
+            Maximum response time to compute, usually the simulation length. Not
```
dbrakenhoff: "usually the simulation length" - Is this frequently used? If not maybe change the language here.

pastas/rfunc.py
```diff
@@ -542,52 +669,68 @@ class HantushWellModel(RfuncBase):
...
+        falls back to `False` and uses the step response.
     quad: bool, optional
-        Use the method 'numba_quad' to compute the step_response.
+        Use `quad_step` to compute the step response using numerical
+        integration. Default is False.
+    approximate_tmax: bool, optional
+        If True, get_tmax will use the fast Lambert W approximation (default). If False,
+        it will use the exact numerical root finding method.
+    log_b: bool, optional
+        Whether to use log10 scaling for parameter b, which is related to the diffusivity
```
dbrakenhoff: "related to diffusivity and distance" ?

Maybe clearer: 

"Whether to use log10 scaling for parameter b. Default is True. In this Hantush implementation parameter b is multiplied by distances squared meaning values of b can get very small under certain conditions. Log scaling can help with optimization when the value of b is very small and the range of b spans several orders of magnitude." ?

pastas/rfunc.py
```diff
@@ -542,52 +669,68 @@ class HantushWellModel(RfuncBase):
...
-
-        ps.HantushWellModel.impulse
-
     where r is the distance from the pumping well to the observation point and must
     be specified. A, a, and b are parameters, which are slightly different from the
     Hantush response function. The gain is defined as:
 
     :math:`\\text{gain} = A K_0 \\left( 2r \\sqrt(b) \\right)`
 
+    The direct impulse response is not implemented for this class. Block responses
```
dbrakenhoff: Still true?

pastas/rfunc.py
```diff
@@ -1067,70 +1248,56 @@ def moment(
...
-        -----
-        The exported dictionary should exactly match the input arguments of __init__.
-
-        """
-        data = {
-            "class": self._name,
-            "up": self.up,
-            "gain_scale_factor": self.gain_scale_factor,
-            "cutoff": self.cutoff,
+        settings = super().to_dict() | {
```
dbrakenhoff, line 963: nice and short :)

pastas/rfunc.py
```diff
@@ -1204,44 +1372,57 @@ def moment(
...
+        Scale factor used to set the initial value and bounds of the gain
+        parameter, computed as `1 / gain_scale_factor`.
 
-    _name = "One"
+    """
 
     def __init__(
         self,
         cutoff: float = 0.999,
+        use_impulse: bool = False,
```
dbrakenhoff: I guess the ABC requires `use_impulse` to be in the __init__? It would be nicer if it was never included in my opinion instead of triggering a warning?

martinvonk: It also has to be in the init becuase it is in the rfunc.to_dict() now by default. 
I can also just implement an impulse method for the ps.One based on the `scipy.signal.unit_impulse` method.

pastas/rfunc.py
martinvonk: If you (I in this case) don't remember it wasn't important :)

pastas/rfunc.py
dbrakenhoff: Instead of leaving it in, you could also set One's to_dict() method to remove that value from the setting dictionary? But just implementing an impulse is also fine of course...

dbrakenhoff (approved):

pastas/rfunc.py
martinvonk: I went for the impulse :)

mbakker7:
I am happy with this PR. Only thing to consider is whether it shouldn't be called 
`use_block=True`. And then use impulse when `use_block=False`. 
It is a more positive way of formulating the default. We normally use the block. Unless we use the impulse.
I also don't think we should say that we use the impulse for the block. No, we simply use the impulse. But internally, we have a variable called 'block' that refers to the 'impulse', but that is not something to bother the user with. 
Also, this method is still widely known as the method of impulse response functions. So if we have `use_impulse=False` may be confusing to some users. 

Writing all this, maybe I should re-read the doc string to see what it says now. But my vote would be for `use_block` rather than `use_impulse`. Thoughts?

mbakker7:
Follow up: Doc string of the `Rfunc` base class now states for `use_impulse`
`Use the impulse response instead of the step response to compute the block response. Default is False.`. 
This is confusing. It should say
`Use the impulse response to compute the response function. Default is False which means it uses the block response to compute the response function`. Or better, if we adopt `use_block`:
`Use the block response to compute the response function. Default is True. When False, it uses the impulse response function`

martinvonk:
Thank you for the feedback and the additions @mbakker7. I think `use_block` is indeed better so I adjusted it. Also the docstring is much better now. If you approve the PR we can merge this I think.

mbakker7 (approved):
I think this looks very nice. Only question I have is whether there is going to be a "rcparams" option that can be set so that the use_block is always false (i.e., the impulse response is always used). Are we going to implement that or will that be a change to consider later?

martinvonk:
> I think this looks very nice. Only question I have is whether there is going to be a "rcparams" option that can be set so that the use_block is always false (i.e., the impulse response is always used). Are we going to implement that or will that be a change to consider later?

I think that's a separate issue for setting globals settings in pastas. Because then you also want a global to turn on/off caching, numba etc via rcParams. So good to make an issue for that.

PR #1178, opened 2026-05-04
add pause to slow down tracksolve

dbrakenhoff:
Small silly change to configure how quickly tracksolve updates the plots (and performs the solve). Useful for demos.


martinvonk (approved):
Need to merge this in #1031 where TrackSolve is moved

PR #1187, opened 2026-05-08
fix cachedmethod deprecation warning

### LINKED ISSUE #1146: `cachedmethod` gives DeprecationWarning
### Describe the bug
```python
WARNING:py.warnings:~/pastas/pastas/decorators.py:277: DeprecationWarning: decorating class methods with @cachedmethod is deprecated
```

dbrakenhoff:
- code didn't get prettier but apparently this is what it should be now... ?


martinvonk:
> * code didn't get prettier but apparently this is what it should be now... ?

Looks a bit awkward but I guess this is fine if it works :). Where did you find this solution? Should we maybe ask the developers of cachetools what's up?

martinvonk (approved):
I'm fine with this solution but see comment above.

dbrakenhoff:
I asked Claude :innocent: ... I think maybe the warning has something to do with the fact we're wrapping it in our own decorator, but I'm not entirely sure. Might be good to check with the developers of cachetools if what we're doing is okay... I'll open a new issue, see #1190. But for now this fixes things :).

PR #1188, opened 2026-05-11
add StatTimer displaying user-specified stats in progress indicator

dbrakenhoff:
# Short description
- updates RMSE value every N iterations in the waitbar
- useful for slow models to check if your solution is getting anywhere (i.e. single-cell UZF MODFLOW models)

- [x] is documented
- [x] type hints
- [ ] tests added or expanded

### DIFF
--- a/pastas/timer.py
+++ b/pastas/timer.py
@@ -19,9 +19,11 @@ This will print the following to the console::
 
 """
 
+from typing import Literal
+
 from pastas.decorators import PastasDeprecationWarning
 from pastas.stats import metrics
-from pastas.typing import Literal, Model
+from pastas.typing import Model
 
 
 @PastasDeprecationWarning(
--- a/pastas/timer.py
+++ b/pastas/timer.py
@@ -20,6 +20,7 @@ This will print the following to the console::
 """
 
 from pastas.decorators import PastasDeprecationWarning
+import numpy as np
 
 
 @PastasDeprecationWarning(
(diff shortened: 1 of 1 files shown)


martinvonk:
I think this is nice. Three questions:
- Is it maybe nicer to show the SSE since that is what we minimize?
- Related to the comment above but I know this works for scipy.optimize.minimize. The callback in scipy.minimize returns the OptimizeResult class for every iteration where (half) the sum of squared residuals is an attribute. This way we don't have to compute the SSE ourselves which might save time. So if we could use that, or the return value of self.solver.objfunction that could be nice(r).
- Maybe we should also move this file to the solver directory as done with the callbacks in #1031?

dbrakenhoff:
> Is it maybe nicer to show the SSE since that is what we minimize?

I think the value of the RMSE is easier to interpret. So I'd like to keep that one :innocent: 

> Related to the comment above but I know this works for scipy.optimize.minimize. The callback in scipy.minimize returns the OptimizeResult class for every iteration where (half) the sum of squared residuals is an attribute. This way we don't have to compute the SSE ourselves which might save time. So if we could use that, or the return value of self.solver.objfunction that could be nice(r).

We actually call the callback ourselves in misfit(), so this won't work at the moment. If we pass the callback onto scipy, then this becomes an option. But I'll leave that to a separate PR. 

> Maybe we should also move this file to the solver directory as done with the callbacks in ...

Wrong PR linked, but good idea :+1:

dbrakenhoff:
> Maybe we should also move this file to the solver directory as done with the callbacks in ..

This one is also easier to do in another PR where the solver directory already exists.

martinvonk:
> > Is it maybe nicer to show the SSE since that is what we minimize?

> I think the value of the RMSE is easier to interpret. So I'd like to keep that one 😇

Yes true. Another thought: maybe to make fit_statistic an keyword argument to SolveTimer initialization? Then people can choose and you don't need a seperate class..

> > Related to the comment above but I know this works for scipy.optimize.minimize. The callback in scipy.minimize returns the OptimizeResult class for every iteration where (half) the sum of squared residuals is an attribute. This way we don't have to compute the SSE ourselves which might save time. So if we could use that, or the return value of self.solver.objfunction that could be nice(r).

> We actually call the callback ourselves in misfit(), so this won't work at the moment. If we pass the callback onto scipy, then this becomes an option. But I'll leave that to a separate PR.

Ah I understand, fine by me. 

> > Maybe we should also move this file to the solver directory as done with the callbacks in ...

> Wrong PR linked, but good idea 👍

Sorry 👼, I meant #1031.  

> This one is also easier to do in another PR where the solver directory already exists.

👍

pastas/timer.py
```diff
@@ -83,3 +84,44 @@ def timer(self, _, n: int = 1):
                     f"Model solve time exceeded {self.max_time} seconds!"
                 )
         return displayed
+
+
+class StatTimer(SolveTimer):
+    """StatTimer that updates a user-specified solve statistic every N iterations."""
+
+    def __init__(
+        self, ml, *args, statistic="rmse", update_interval: int | None = None, **kwargs
```
martinvonk: self, ml: pastas.typing.Model, *args, statistic: Literal["rmse", "sse", ...] = "rmse", update_interval: int | None = None

martinvonk (approved):

raoulcollenteur (approved):

PR #1189, opened 2026-05-12
Complex step for jacobian in LeastSquares solver

### LINKED ISSUE #982: Should we compute our own Jacobian?

dbrakenhoff:
These changes allow users to specify `jac="cs"` when using the `LeastSquares` solver. This estimates the Jacobian by adding small complex step to each parameter. The advantage (I believe) is fewer iterations to compute the Jacobian, and a more accurate result. This requires pastas to support complex values parameters. This PR investigates what would be necessary to implement this. 

We should discuss if this is something we want? If we do, we need to complete the checklist below. 

Supported rfuncs:
 - Exponential
 - Hantush and HantushWellModel (though the latter still needs to be tested)
 - DoubleExponential
 - FourParam (only when use_block=False)
 - Spline
 - One
 - Polder
 - Kraijenhoff

Supported StressModels/RechargeModels
- TARSO
- Linear
- FlexModel

Not supported:
- Gamma (gammainc does not support complex inputs), and mpmaths version is very slow, comparatively
- Non-linear recharge models Peterson and Berendrecht are currently not supported

- [ ] closes issue #xxxx
- [ ] is documented
- [ ] type hints
- [ ] tests added or expanded

Note these changes were implemented based on the solver_initialize branch, which is why I set that as the target branch for this PR

Update 2026-05-26: adjusted support list above based on #1236.


dbrakenhoff:
I think at the moment the calculation isn't accurate for non-linear recharge models since they don't support complex parameters. Supporting this would require the numba functions to accept complex inputs.

martinvonk:
`mpmath` has a `gammainc` function that supports complex values. So maybe we could look into that for supporting the gamma function.

raoulcollenteur:
Would this not require a jump in scipy dependence o 1.16 as well ?

pastas/solver/least_squares.py
martinvonk: jac: Literal["2-point", "3-point", "cs"] = "2-point",

raoulcollenteur (approved):
Nevermind my comment about cs, that's only valid for 'lm' but it was available for trf in earlier versions.

dbrakenhoff:
Alright with #1236 this is somewhat complete now. 

I did some experiments on the Pastas Examples models using different solver settings for `jac` and `x_scale="jac"`.

The results are below. The NaNs for the complex step are caused by models using `Gamma` or `FourParam` with `use_block=True`. 

This is the delta NSE, so the change in NSE relative the currently default Pastas solve. Basically, all methods work similarly, and only somewhat significant changes occur for the snowmodel models.

image


And perhaps also interesting for #1234, these are the relative changes in solve time relative to the current default solve in Pastas:


image

pastas/recharge.py
```diff
@@ -314,19 +315,24 @@ def simulate(
...
     @staticmethod
     @njit
     def get_root_zone_balance(
         pe: ArrayLike,
         ep: ArrayLike,
-        srmax: float = 250.0,
-        lp: float = 0.25,
-        ks: float = 100.0,
-        gamma: float = 4.0,
+        srmax: complex = 250.0,
```
martinvonk: maybe we can do `float | complex`

martinvonk:
How hard would it be to fix this?
> Non-linear recharge models Peterson and Berendrecht are currently not supported

dbrakenhoff:
> How hard would it be to fix this?
> 
> > Non-linear recharge models Peterson and Berendrecht are currently not supported

Not that hard I think, it requires the threshold checks to be rewritten to use the real part of the parameters, similar to what was implemented for FlexModel now.

martinvonk:
> Gamma (gammainc does not support complex inputs), and mpmaths version is very slow, comparatively

Four questions:
- Does gamma also not work for `Gamma(use_block=False)`?
- How much slower is the mpmaths version? 
- Can we maybe have a slow version of mpmaths Gamma, in a seperate rfunc, maybe on Pastas-Plugins?
- Maybe nice to have a check that if your model has a Gamma response and jac="cs", it defaults to "3-point" and logs an error.

martinvonk:
> Alright with #1236 this is somewhat complete now.
> 
> I did some experiments on the Pastas Examples models using different solver settings for `jac` and `x_scale="jac"`.

Interesting results. So my hypothesis is that "3-point" will give some more accurate results for difficult problems, but for most cases "2-point" is fine and faster. For the `x_scale="jac"` I think you always want that, especially if your initial parameters make sense. Because otherwise you can make a very large jump for your first iteration from your initial parameters into parameters space. So it should reduce the nfev significantly, and maybe yield more accurate results.

> And perhaps also interesting for #1234, these are the relative changes in solve time relative to the current default solve in Pastas:

Can you for the last table also show the number of function evaluations?

martinvonk:
> > How hard would it be to fix this?
> > > Non-linear recharge models Peterson and Berendrecht are currently not supported
> 
> Not that hard I think, it requires the threshold checks to be rewritten to use the real part of the parameters, similar to what was implemented for FlexModel now.

If it is not too much work I'd prefer if we just try to support everything..

dbrakenhoff:
> - Does gamma also not work for Gamma(use_block=False)?

This does work in fact! `gamma()` apparently does support complex inputs.  

> - How much slower is the mpmaths version?

Ok i found that there were some improvements to be made, by using different mpmath implementations with lower precision. One call to `step()` with mpmaths `gammainc` is about 100 times slower: 5 us (scipy) vs 500us (mpmath). Performing a solve on one of the example models goes from 1.3 seconds (scipy) to 11.3 seconds (mpmath). (Interestingly, calling solve with mpmath step and jac="cs" lowers the solve time to 4s.)

> - Can we maybe have a slow version of mpmaths Gamma, in a seperate rfunc, maybe on Pastas-Plugins?

Yea, that would be a reasonable solution. Should be simple enough to, just inherit from `Gamma` , and replace `step()`

> - Maybe nice to have a check that if your model has a Gamma response and jac="cs", it defaults to "3-point" and logs an error.

Or just raise an error that tells the user to get their inputs right :innocent: ? But if we want to default to `"cs"` then this might be a logical addition.

> Can you for the last table also show the number of function evaluations?

image

dbrakenhoff:
Or did you want to see nfev + njev?

martinvonk:
> > * Does gamma also not work for Gamma(use_block=False)?
> 
> This does work in fact! `gamma()` apparently does support complex inputs.

Great, I hoped so :)

> > * How much slower is the mpmaths version?
> 
> Ok i found that there were some improvements to be made, by using different mpmath implementations with lower precision. One call to `step()` with mpmaths `gammainc` is about 100 times slower: 5 us (scipy) vs 500us (mpmath). Performing a solve on one of the example models goes from 1.3 seconds (scipy) to 11.3 seconds (mpmath). (Interestingly, calling solve with mpmath step and jac="cs" lowers the solve time to 4s.)

Okay that's pretty slow :)

> > * Can we maybe have a slow version of mpmaths Gamma, in a seperate rfunc, maybe on Pastas-Plugins?
> 
> Yea, that would be a reasonable solution. Should be simple enough to, just inherit from `Gamma` , and replace `step()`

Yes good idea.

> > * Maybe nice to have a check that if your model has a Gamma response and jac="cs", it defaults to "3-point" and logs an error.
> 
> Or just raise an error that tells the user to get their inputs right 😇 ? But if we want to default to `"cs"` then this might be a logical addition.

I would prefer logging, not raising. We can also refer that `use_block=False` does work for Gamma with "cs".

> > Can you for the last table also show the number of function evaluations?
> 
> image

> Or did you want to see nfev + njev?

No this is fine, thanks :) 

Interesting that sometimes you can have more nfev but a faster runtime. Probably because you don't get search in parameters space with long tmax's.

pastas/recharge.py
```diff
@@ -365,29 +371,31 @@ def get_root_zone_balance(
         """
         n = pe.size
         # Create empty arrays to store the fluxes and states
-        sr = zeros(n + 1, dtype=float64)  # Root Zone Storage State
+        sr = zeros(n + 1, dtype=np.complex128)  # Root Zone Storage State
```
martinvonk, line 384: just curious, why complex128 and not complex64?

pastas/rfunc.py
```diff
@@ -536,10 +544,54 @@ def step(
...
+        dt: float = 1.0,
+        cutoff: float | None = None,
+        maxtmax: float | None = None,
+        **kwargs,
+    ) -> ArrayLike:
+        if np.iscomplexobj(p) and self.use_block:
+            raise NotImplementedError(
+                "Gamma.step does not support complex-step Jacobian evaluation "
+                "(jac='cs') by default. Set use_block=False on the "
+                "Gamma response function, set complex_gamma=True or use a "
```
martinvonk: should be: `complex_step=True`

pastas/rfunc.py
```diff
@@ -485,9 +491,11 @@ def __init__(
         self,
         cutoff: float = 0.999,
         use_block: bool = True,
+        complex_step: bool = False,
         **kwargs,
     ) -> None:
         super().__init__(cutoff=cutoff, use_block=use_block, **kwargs)
+        self.complex_step = complex_step
```
martinvonk, line 498: maybe we should also add this to the `.to_dict`?

pyproject.toml
```diff
@@ -55,6 +55,7 @@ formatting = ["pastas[ruffing]"]
 linting = ["pastas[ruffing]"]
 caching = ["cachetools"]
 codespelling = ["codespell[toml]"]
+mpmath = ["mpmath"]
 full = ["pastas[solvers,interactive,datasets,caching]"]
```
martinvonk: add mpmath to full?

pastas/rfunc.py
```diff
@@ -485,9 +491,11 @@ def __init__(
         self,
         cutoff: float = 0.999,
         use_block: bool = True,
+        complex_step: bool = False,
```
martinvonk, line 494: maybe also add a test?

martinvonk (approved):

pastas/recharge.py
dbrakenhoff: 2 float64s make a complex128 :)

pastas/recharge.py
dbrakenhoff: Does numba njit infer types for compiling from these hints? I put complex first just in case...

martinvonk: According to our python parallel course yes. I'm fine with complex first. As long as it is clear that we can do both.

pastas/rfunc.py
dbrakenhoff: So demanding ;)

martinvonk: 😘

martinvonk:
Looks great @dbrakenhoff. If I'm not mistaken the only things that don't work now are:
- `FourParam` with `use_block=True`
- `Gamma` without `complex_step=False` and `use_block=True`
Correct?

I solved the merge conflict from #1219 for you.

dbrakenhoff:
- Aight, Berendrecht and Peterson now also supported. 
- Gamma has a new init kwarg to allow use of the mpmath version of Gamma.step. 

Ready to merge after #1031. I will update target branch after #1031 is merged.

martinvonk:
I tried something for the Gamma.step since: `gammainc(a, x) = 1 - gammaincc(a, x)` but `gammaincc(a, x)` also does not support complex numbers.

martinvonk:
Alternatively, https://github.com/flintlib/python-flint also has a gamma_upper

PR #1214, opened 2026-05-19
Pin dependency upper versions for package versions that will (most likely) break Pastas

### LINKED ISSUE #1107: List dependency package versions on release, and maybe pin upper versions?
I recently got a question about support for older Pastas versions, which made me wonder about installing older pastas versions. 

So question 1: Should we add a list of dependency versions that were used to run the GH actions on latest Python when we do a release? E.g. include the output of `ps.show_versions()` in the release message? This makes it clear what combination of packages will definitely work for that version of Pastas. 

Since we do not use upper limits on package versions, only lower limits, it is difficult to know what recent versions of packages are supported for a specific pastas release. Say I want to know which pandas/numpy to use for Pastas 1.5.0 ? If I tell conda/pip/uv to install this version of pastas, they will all currently install pandas 3.0 (due to the lack of upper version limit). I guess the quickest way is to retrieve a working environment for an older pastas version is to use the lower version limits of packages in the pyproject.toml? 

(Just found out about this command `uv pip install --resolution=lowest-direct`)

So question 2, should we use upper major version limits on packages in Pastas from now on? Similar to the temporary setting where we allow pandas 2.X but not 3.X? I think pandas, numpy and maybe scipy are the really important ones, the rest we can get away with not specifying.

martinvonk:
Pin dependency upper versions for package versions that will (most likely) break Pastas. If a dependency announces a version with big changes we should do a pin the version lower. But that's what I think is the best solution for now.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,13 +19,13 @@ maintainers = [
     { name = "O.N. Ebbens", email = "o.ebbens@artesia-water.nl" },
     { name = "M.A. Vonk", email = "vonk.mart@gmail.com" },
 ]
-requires-python = ">= 3.11"
+requires-python = ">= 3.11, < 3.15"
 dependencies = [
-    "numpy >= 2.0.0",
-    "matplotlib >= 3.9.0",
+    "numpy >= 2.0.0, < 3.0",
+    "matplotlib >= 3.9.0, < 4.0",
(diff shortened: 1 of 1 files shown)


raoulcollenteur:
I forgot which Pastas you are allowed to brake...

raoulcollenteur (approved):

PR #1217, opened 2026-05-19
Make StressModelBase an abstract base class and refactor `stressmodels.py`

### LINKED ISSUE #1067: Make all the Base classes in Pastas an AbstractBaseClass
### Describe the task
I would like to propose converting all Base classes in Pastas into proper AbstractBaseClass (ABC) implementations. This would make the architecture more pythonic, enforce consistency, and clearly define which methods and properties must be present or overridden in all subclasses. By explicitly marking required behavior with `@abstractmethod` and `@property`, contributors will immediately understand what needs to be implemented, reducing ambiguity and runtime errors.

martinvonk:
# Short description
This pull request refactors the `stressmodels.py` file by making:
- StressModelBase an AbstractBaseClass (just like RfuncBase)
- update the `to_dict` method (just like RfuncBase and BaseSolver)
- make `nsplit` a property instead of an attribute such that it is safe to access
- move some methods such that the order of methods is consistent accross stressmodels

Note that this PR does not change any behavior (significantly).

- [x] related to issue #1067


pastas/stressmodels.py
```diff
@@ -214,20 +250,20 @@ def get_stress(
         freq: str | None = None,
         istress: int | None = None,
         **kwargs,
-    ) -> Series:
-        """Returns the stress(es) of the time series object as a pandas DataFrame.
+    ) -> Series | DataFrame:
+        """Get the stress(es) of the time series object as a pandas Series.
```
raoulcollenteur: or DataFrame

pastas/stressmodels.py
```diff
@@ -1641,39 +1656,104 @@ def __init__(
...
 
-            # Check if precipitation is likely in mm/d and not m/d. If the maximum
-            # value of the annual sums is smaller than 12 (m/d), the highest annual
-            # precipitation in the world, then the precipitation is very likely in m/d
-            # and not in mm/d. In this case a warning is given for nonlinear models.
+        # Check if precipitation is likely in mm/d and not m/d. If the maximum
+        # value of the annual sums is smaller than 12 (m/d), the highest annual
+        # precipitation in the world, then the precipitation is very likely in m/d
+        # and not in mm/d. In this case a warning is given for nonlinear models.
+        freq_offset = "YE" if pandas_version >= parse_version("2.2.0") else "A"
```
raoulcollenteur: Given that pandas 2.2.0 is now the minimum let's change this to freq_offset="YE"

pastas/stressmodels.py
```diff
@@ -1599,36 +1619,31 @@ def __init__(
...
         # Store a temperature time series if provided/needed or set to None
-        if self.recharge.snow is True and temp is None:
-            msg = (
-                "Recharge model requires a temperature series. No temperature series "
-                "were provided."
-            )
-            raise TypeError(msg)
-        if temp is not None:
+        if temp is None:
+            if hasattr(self.recharge, "snow") and self.recharge.snow is True:
```
raoulcollenteur: `self.recharge.snow is True` would suffice here because if there is no attribute snow the if statement would fail anyway..

raoulcollenteur (approved):
Couple of small things, can be merged after changing and checking those.

pastas/stressmodels.py
martinvonk: fixed for all checks

PR #1221, opened 2026-05-20
Make NoiseModelBase an AbstractBaseClass

### LINKED ISSUE #1067: Make all the Base classes in Pastas an AbstractBaseClass

martinvonk:
This pull request refactors the noise model base class and its subclasses to use Python's abstract base class (ABC) mechanism, improves parameter handling, and standardizes the interface for noise models in line with StressModels (#1217) and Rfuncs (#1173).

### DIFF
--- a/pastas/noisemodels.py
+++ b/pastas/noisemodels.py
@@ -17,6 +17,7 @@ See Also
 pastas.model.Model.add_noisemodel
 """
 
+from abc import ABC, abstractmethod
 from logging import getLogger
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


pastas/noisemodels.py
```diff
@@ -109,11 +111,10 @@ def _set_dist(self, name: str, value: str) -> None:
 
     def to_dict(self) -> dict:
         """Method to return a dict to store the noise model"""
-        data = {"class": self._name, "norm": self.norm}
-        return data
+        settings = {"class": self._name, "norm": self.norm}
```
raoulcollenteur: Why change this to settings ? I think data was better, settings can be confusing because it already means something in Pastas.

raoulcollenteur (approved):

pastas/noisemodels.py
martinvonk: `settings` is the same variable name in `rfunc.py`

martinvonk: Changed it for noisemodels.py in this pr.

PR #1227, opened 2026-05-21
Improve verbose series validation

### LINKED ISSUE #861: [ENHANCEMENT] Update documentation on `validate_oseries` and `validate_series`.
The documentation states what the requirements are, but it doesn't state which errors `pastas` will fix (for example a Nan value in the head series).

dbrakenhoff:
When verbose=True the behavior of `_validate_series` was a bit inconsistent. Depending on the type of error the function either prints the outcomes of all checks, or raises an error. This PR attempts to avoid raising when verbose=True, and attempts to print the validation results to screen. Using `verbose=True` makes the notebook about preparing timeseries much easier to read in my opinion. 

Improve `_validate_series` when verbose=True:
- when verbose, we want to print all errors, and try to avoid raising errors
- shortcut return when series is DataFrame
- add support for failing equidistant check because index is not datetime etc. Instead of raising, print a more vague error message

Improve notebook
- use verbose everywhere and uncomment validate_series calls
- improve suggested solution for dtype not float
- add info that pastas supports NaNs to to example notebook

- [ ] closes issue #xxxx
- [x] is documented
- [x] type hints
- [x] tests added or expanded

### DIFF
--- a/doc/examples/prepare_timeseries.ipynb
+++ b/doc/examples/prepare_timeseries.ipynb
@@ -1085,14 +1085,6 @@
     "\n",
     "pd.concat([valueskept, duplicates], axis=1)"
    ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "b2cf2beb",
-   "metadata": {},
-   "outputs": [],
-   "source": []
   }
  ],
  "metadata": {
--- a/doc/examples/prepare_timeseries.ipynb
+++ b/doc/examples/prepare_timeseries.ipynb
@@ -68,7 +68,6 @@
     "index = pd.date_range(\"2000-01-01\", \"2000-01-10\")\n",
     "series = pd.DataFrame(data=[np.arange(10.0)], index=index)\n",
     "\n",
-    "# Here's error message returned by Pastas\n",
     "ps.validate_stress(series, verbose=True)"
    ]
   },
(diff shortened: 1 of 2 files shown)


pastas/timeseries.py
OnnoEbbens:
```suggestion
    # 7. Check if the time series has nan-values
```

doc/examples/prepare_timeseries.ipynb
```diff
@@ -67,7 +69,7 @@
     "series = pd.DataFrame(data=[np.arange(10.0)], index=index)\n",
     "\n",
     "# Here's error message returned by Pastas\n",
```
OnnoEbbens: No error message is returned with `verbose=True`

OnnoEbbens: also fore some other cases below

OnnoEbbens (commented):
I had some comments, nothing big.

I was also thinking if `verbose=True` should be the default. The advantage is that you always get all the messages. The disadvantage is that you won't get an error when you run the code without looking at the printed messages in the console. It makes sense to have very strict default settings for this function. Therefore I am happy to leave the default `verbose=False`.

raoulcollenteur (approved):

dbrakenhoff:
Will merge after tests pass.

PR #1232, opened 2026-05-22
Fixes some things in check module

### LINKED ISSUE #1139: Allow extra arguments to be passed on in checks module

For example, when I used the check with the rsq threshold, I wanted to pass on the tmin and tmax. This is not possible right now, but would be nice and simple to implement.

dbrakenhoff:
# Short description
- only check uncertainty for parameters that vary (this fixes issue with checks not passing because fixed parameters would fail uncertainty check). 
- show abs param value in statistic column for uncertainty check
- set name of model as index name for check report dataframe
- support kwargs in stat_ufunc_threshold and rsq_geq_threshold (#1139)

### DIFF
--- a/pastas/check.py
+++ b/pastas/check.py
@@ -133,6 +133,7 @@ def stat_ufunc_threshold(
     ufunc: Callable,
     statistic: str,
     threshold: float,
+    **kwargs,
 ):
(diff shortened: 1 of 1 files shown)


martinvonk (approved):

raoulcollenteur (approved):
Love it! Pass on the kwargs :)

PR #1237, opened 2026-05-28
Improve parnames tt

### LINKED ISSUE #1231: Change parameter names of ThresholdTransform even more
PR #1229 changed the default parameter names of `ThresholdTransform` from `ThresholdTransform_1` and `ThresholdTransform_2` to `threshold_1` and `threshold_2`, which is an improvement. However, as we are changing the parameter names, might it be even better to rename the parameters to `transform_h` and `transform_f`?

So I propose to change the default name to `transform` and use `h` (or `d`?) and `f` (factor) as the names of the two parameters. Thoughts?

rubencalje:
# Short description

This PR implements the change proposed in #1231.

This pull request makes minor adjustments to the `pastas/transform.py` file, primarily renaming parameter identifiers and default names for clarity and consistency.

Parameter and naming updates:

* Changed the default `name` argument in the class initializer from `"threshold"` to `"transform"` to better reflect its purpose.
* Renamed parameter keys in `set_init_parameters` from `"_1"` and `"_2"` to `"_d"` and `"_f"` respectively, for improved clarity.

- [ ] closes issue #1231
- [ ] is documented
- [ ] type hints
- [ ] tests added or expanded

### DIFF
--- a/pastas/io/base.py
+++ b/pastas/io/base.py
@@ -118,8 +118,35 @@ def _load_model(data: dict) -> Model:
         sm = _load_stressmodel(smdata, data)
         ml.add_stressmodel(sm)
 
+    file_parameters = data["parameters"]
     # Add transform
     if "transform" in data.keys():
+        if data["transform"]["class"] == "ThresholdTransform":
+            # the parameters of ThresholdTransform were renamed from 1 and 2 to d and f in pastas 2.0.0,
+            # so we need to check if the old parameter names are present and rename them to the new ones
+            old_param_name = f"{data['transform']['name']}_1"
+            if old_param_name in file_parameters.index:
+                new_param_name = f"{data['transform']['name']}_d"
+                logger.warning(
+                    "Renaming parameter %s to %s to match the naming convention of ThresholdTransform.",
+                    old_param_name,
+                    new_param_name,
+                )
+
+                file_parameters.rename(
+                    index={old_param_name: new_param_name}, inplace=True
+                )
+            old_param_name = f"{data['transform']['name']}_2"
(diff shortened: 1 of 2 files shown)


martinvonk (changes requested):
Looks very nice. But this should also be changed in the `pastas.io.base.py` to allow for reading `<2.0 .pas` files. You can look in #1031 where I also did this for the Step name change from `_d` to `_A`. It will probably cause a small merge conflict but that's fine.

rubencalje:
> Looks very nice. But this should also be changed in the `pastas.io.base.py` to allow for reading `<2.0 .pas` files. You can look in #1031 where I also did this for the Step name change from `_d` to `_A`. It will probably cause a small merge conflict but that's fine.

You are right. I implemented and tested the renaming of parameters from old pas-files.

pastas/io/base.py
```diff
@@ -120,6 +120,32 @@ def _load_model(data: dict) -> Model:
...
+                    "Renaming parameter %s to %s to match the naming convention of ThresholdTransform.",
+                    old_param_name,
+                    new_param_name,
+                )
+
+                data["parameters"].rename(
+                    index={old_param_name: new_param_name}, inplace=True
+                )
+            old_param_name = f"{data['transform']['name']}_2"
+            if old_param_name in data["parameters"].index:
```
martinvonk: I think one check is enough right? If _1 is present, _2 is also there.

martinvonk (approved):

pastas/io/base.py
martinvonk: already fixed it. Also alligned it with #1031

PR #1248, opened 2026-06-04
make fit_constant work again

### LINKED ISSUE #1247: ml.residuals goes wrong when fit_constant=False in dev branch
### Describe the bug

I was checking the forecast notebook and noted the bias-correctio was wrong. After a long, long search, I nailed it down to ml.residuals given different answers for different tmin values when ml.solve(fit_constant=False)

### Expected behavior
Take any model and solve with ml.solve(fit_constant=False). Ie. basic_model.ipynb

### Code to reproduce
```
import matplotlib.pyplot as plt

sim1 = ml.simulate(tmin="2000-01-01", tmax="2014-01-01", warmup="3650D")
sim2 = ml.simulate(tmin="2005-01-01", tmax="2014-01-01", warmup="3650D")

obs1 = ml.observations(tmin="2000-01-01", tmax="2014-01-01")
obs2 = ml.observations(tmin="2005-01-01", tmax="2014-01-01")

residuals1 = obs1 - sim1
residuals2 = obs2 - sim2

fig, axes = plt.subplots(1, 2, figsize=(10.0, 4.0), sharey=True, layout="tight")
residuals1.dropna().plot(ax=axes[0], title="Residuals 2000-2014")
residuals2.dropna().plot(ax=axes[1], title="Residuals 2005-2014")

ml.residuals(tmin="2000-01-01", tmax="2014-01-01", warmup="3650D").plot()
ml.residuals(tmin="2005-01-01", tmax="2014-01-01", warmup="3650D").plot()
```

### Package and Python version

Pastas     : 2.0.0
Python     : 3.13.11
Numpy      : 2.4.6
Pandas     : 3.0.3
Scipy      : 1.17.1
Matplotlib : 3.10.9
Numba      : 0.65.1

raoulcollenteur:
# Short description
Fixes a double bug introduced by set_settings and the new solver. The param constant_d wasn't used because the dataframe was rebuilt so use self.set_parameter instead. And use a private attribute to estimate fit_constant when False. This used to be self.normalize_residuals, but I thought self._fit_constant is better and more clear.
- [x] closes issue #1247


pastas/model.py
```diff
@@ -147,7 +147,6 @@ def __init__(
 
         # some _attributes simulation and solving
         self._interpolate_simulation: bool | None = None
-        self._solve_success: bool | None = None
```
martinvonk, line 150: this should have been removed in #1031, but was left over

martinvonk: so good that it's getting removed now

martinvonk:
Good catch @raoulcollenteur. I was still a bit confused about the code and behavior, resulting from the internal state variables. So I tried to change to the code to make the behavior more obvious via a `subtract_mean` keyword argument to `Model.residuals`. Let me know if you like it. If not, feel free to revert https://github.com/pastas/pastas/pull/1248/changes/b3354d59fe202401a4e2efe961dd32e07c674c30.

raoulcollenteur:
@copilot revert https://github.com/pastas/pastas/commit/b3354d59fe202401a4e2efe961dd32e07c674c30.

Let's keep the code simple here, no need for exposing the user to additional keyword options that are only used internally.

martinvonk (approved):
Also fine :)

PR #1176, opened 2026-05-04
Prepare release for Pastas 2.0.0rc0

raoulcollenteur:
# Short description
Release of Pastas 2.0

# PRs to merge before:
- [x] #1031
- [x] #1134 
- [x] #1156
- [x] #1171 
- [x] #1173
- [x] Make sure conda release works again: https://github.com/conda-forge/pastas-feedstock/tree/main/recipe
- [x] #1193 
- [x] #1195 
- [x] #1270 ?
- [x] #1197
- [x] #1244 
- [x] #1255
- [x] #1260
- [x] #1261
- [x] #1265
- [x] #1267
- [x] #1273 ?
- [x] #1271
- [x] #1291
- [x] #1298
- [x] #1299 
- [x] #1300 
- [x] #1301
- [x] #1302

# Add to release notes
- Parameter values can change when using the default LeastSquares solver due to new default settings for the arguments `xscale` (how to scale the jacobian) and `jac` (how to compute the jacobian).
- The default parameter bounds for the FlexModel recharge model parameters changed, possible causing changes to the calibrated parameter values.

# TODO before merging:
- [x] Check if all deprecations are set to Pastas 2.4.0 for future deprecations. Breaking changes are marked with v2.0.0.
- [x] Check with Pandas 2.3 if all still works
- [x] Check #701 
- [x] Make `__version__` in version.py: `2.0.0rc0`
- [x] Check notebooks/rtd
- [x] Make sure we do a merge commit, not squash and merge

PR #1255, opened 2026-06-11
Update defaults for LeastSquares to jac and 3-point

### LINKED ISSUE #1234: Change x_scale to "jac" and jac to "3-point" by default in LeastSquares
> Let's discuss if we want to hardcode all these options. Alternatively we could just store all used-provided solver settings passed on to scipys least_squares in an attribute to ensure reproducibility. That would require less bookkeeping on our end 

 _Originally posted by @raoulcollenteur in [#1031](https://github.com/pastas/pastas/pull/1031/changes#r3241056172)_

I think this is something we should do as Pastas models are computationally cheap (so 3-point won't hurt) and I think we all experienced these are better defaults (x_scale=jac). 

Users will find it easier to get a good model. Can be implemented after merging #1031 

Related to https://github.com/orgs/pastas/discussions/450

raoulcollenteur:
# Short description

- [x] closes issue #1234 
- [x] is documented


raoulcollenteur:
One thing that I am unsure about is how to make this change clear.... I don't know where to add this, ml.solve seems the best option to me. Thoughts?

mbakker7:
Silly question maybe, but is 3-point really that much better than 2-point? Or is 3-point a freebee as the center point is computed anyway?

martinvonk:
> Silly question maybe, but is 3-point really that much better than 2-point? Or is 3-point a freebee as the center point is computed anyway?

No 3-point is not a freebie. 3 point is central differences, so: current point, step forward and step backward. 2-point is forward differences so: current point and step forward. As you already have simulation of the current point, for 2-point you only need one more simulation run per parameter. For 3-point you would need two extra runs per parameter.

mbakker7:
OK, I got it. I figured 2-point was a central difference scheme. As it is just a forward difference scheme it should really be called a 1-point IMHO. But that doesn't matter. Now that I understand, I am in favor of the 3-point scheme.

martinvonk:
I'll look at the defaults and the scipy version commented by ruben in #1234

raoulcollenteur:
@martinvonk or @dbrakenhoff can you do a review here?

raoulcollenteur:
I think this is ready to merge. I made a note in the PR https://github.com/pastas/pastas/pull/1176 to add a note to the release notes.

martinvonk:
I think this is good to go. I fixed some attributes that @rubencalje that we can't store anyway. So we can leave the minimum scipy version at 1.13.0 if we want.

martinvonk (approved):

martinvonk:
I think it would be good if @dbrakenhoff also had a look for the callback behavior.

raoulcollenteur:
> I think it would be good if @dbrakenhoff also had a look for the callback behavior.

That was actually not the point of this PR si I'll merge for now and we can open a new issue if the changes you made cause any problems. That would be cleaner anyway :)

PR #1258, opened 2026-06-11
Add Ireson recharge model (deficit-based with bypass flow)

### LINKED ISSUE #1257: Create deficit based nonlinear recharge model
Currently, pastas includes storage-based nonlinear recharge models like `FlexModel` and `Berendrecht` and `Peterson`. In Ireson and Butler (2013) a Soil Moisture Deficit (SMD) approach is used, rather than total root zone storage. The model relies on 4 parameters:
- rc: Root constant.
- pwp: Permanent wilting point.
- bf: Bypass fraction (can be fixed to 0 to ignore bypass)
- th: Bypass threshold (can be fixed to 0 by default, as it is often insensitive).  

By default, only 2 parameters (rc and pwp) would need to be calibrated to keep the model lightweight.

martinvonk:
# Short description
This PR introduces the Ireson recharge model, implementing the simple nonlinear recharge model evaluated in [Ireson and Butler (2013)](www.doi.org/10.5194/hess-17-2083-2013). 

The model has 4 parameters (rc, pwp, bf, th), but is configured to calibrate only 2 parameters by default: the root constant (rc) and the permanent wilting point. The bypass fraction (bf) and threshold (th) are set to zero and not calibrated by default.

From nonlinear notebook:
image

- [x] closes issue #1257
- [x] is documented
- [x] type hints
- [x] tests added or expanded


mbakker7:
I am not sure how we should handle this, but there are many recharge models we could introduce. I thought the process is to introduce this first as a Pastas Plugin. Then when it turns out to perform well, we can introduce it in the main version. Thoughts?

rubencalje (approved):
I like this Recharge-model, because it is a bit simpler than the other complex recharge-models, and it is implemented elsewhere as well.

If we put this in pastas plugins, it is harder to find and harder to maintain, so I think it is ok to  merge.

raoulcollenteur:
Let's discuss after releasing 2.0. If added, I think true value would also be in adding a chalk aquifer example or fractured rock. That would actually be very interesting for many international users and show Pastas works for those sorts of aquifers as well.

Issue #1166, opened 2026-04-19
Make parameter `ks` in FlexModel a scale factor of `sr_max`


martinvonk:
### Idea: Redefining `ks` as a scale factor in the FlexModel

I’ve noticed that the current implementation of the FlexModel recharge model often leads to physical inconsistencies during calibration. Because `ks` is defined in **mm/d** and `sr_max` in **mm**, their parameter bounds frequently overlap, allowing `ks` to become larger than `sr_max`. This implies a daily outflow greater than the total rootzone storage capacity, which doesn't make much physical sense. Furthermore, these two parameters are highly correlated, which usually forces me to fix `sr_max` just to get a stable result.

I’m interested in testing whether redefining `ks` as a dimensionless scale factor between **0 and 1** might resolve these issues. Since a scale factor would be capped at 1, it would inherently prevent the outflow from exceeding `sr_max`, ensuring a more physically plausible model. I suspect this change might also reduce the parameter correlation, potentially making the model easier to calibrate without having to fix `sr_max` manually.

I think this is worth exploring, but I'm open to other thoughts on this.

raoulcollenteur:
Found it! As suggested in #1260, the outflow only equals the size of the storage reservoir if gamma is one and ks>srmax. If gamma > 1, the outflow is reduced and ks can be larger than srmax without causing the drainage to exceed the current storage. 

I suggest to close this issue.

raoulcollenteur:
@martinvonk ?

martinvonk:
> Found it! As suggested in [#1260](https://github.com/pastas/pastas/pull/1260), the outflow only equals the size of the storage reservoir if gamma is one and ks>srmax. If gamma > 1, the outflow is reduced and ks can be larger than srmax without causing the drainage to exceed the current storage.

Still don't really understand this. So if gamma > 1, the outflow for one time step can be larger than what is stored? Where is that water then coming from? But maybe that's just my brain hurting.

But even if this behavior is logical, I still think this might be a good idea. Then the `ks` scale factor could be larger than 1 as well; e.g. bounds at 0-2. Which might still be easier to estimate than the current `ks`.

PR #1260, opened 2026-06-11
Update parameter bounds to ensure reasonable values for the FlexModel

raoulcollenteur:
# Short description

This PR changes the default parameter bounds for $k_s$, $\gamma$, and $S_{r,max}$, to improve the calibration, and ensure obtaining more plausible recharge results by default .

- Change $S_{r,max}$ to have a lower bound of 10 mm instead of 1e-5 (almost zero) to ensure that the bucket can store some water.
- Change $\gamma$ to have a lower bound of 1 and upper bound of 5 instead of [1e-5-20]. Below 1, a drier soil would increase the hydraulic conductivity which is not realistic. Above ~5, it becomes so nonlinear (i.e., sr/srmax becomes really small) that larger values of gamma practically have no effect and make calibration difficult. 
- Change bounds of $k_s$ to 1-1000 mm/d from 1e-5-1e4 previously. These would be extreme cases, particularly the upper bound which is a very high for most soils. If ks is that high, it basically always drains all the incoming precipitation and a linear model might work just as well.

- [x] closes issue #1281 
- [x] will be documented in the release notes


pastas/recharge.py
```diff
@@ -225,11 +225,11 @@ def get_init_parameters(self, name: str) -> DataFrame:
         """Get initial parameters and bounds for the FlexModel recharge model."""
         parameters = DataFrame(
             [
-                (250.0, 1e-5, 1e3, True, name),  # srmax
+                (250.0, 10.0, 1e3, True, name),  # srmax
                 (0.25, 1e-5, 1.0, False, name),  # lp
-                (100.0, 1e-5, 1e4, True, name),  # ks
-                (2.0, 1e-5, 20.0, True, name),  # gamma
-                (1.0, 0.25, 2.0, True, name),  # kv
+                (100.0, 1.0, 1e4, True, name),  # ks
```
martinvonk: it is also kinda weird that the pmax of ks (1e4) can be larger than the pmax of srmax (1e3). Related to #1166

pastas/recharge.py
```diff
@@ -225,11 +225,11 @@ def get_init_parameters(self, name: str) -> DataFrame:
...
+                (2.0, 1e-5, 5.0, True, name),  # gamma
+                (1.0, 0.5, 1.5, True, name),  # kv
```
martinvonk: I'd keep the pmax of kv to 2.0. See #1262.

raoulcollenteur:
@martinvonk $k_s$ > $S_{r,max}$ has indeed no effect for when the reservoir is full (sr/srmax=1), but does have an effect when not saturated for gamma > 1. For example, when srmax=250, sr=125, gamma=5, and ks=500, the recharge is 15.625. if gamma is 1, ks > srmax has no effect because it will just drain all the water (recharge is 250 so it will drain all available 125 mm).

EDIT: I thought you opened an issue about this.. perhaps we link that here for reference.

martinvonk:
> EDIT: I thought you opened an issue about this.. perhaps we link that here for reference.

Yes this is #1166

dbrakenhoff (approved):
Only somewhat qualified to approve this PR. I'm guessing this narrows the bounds a bit which will result in more realistic and better models. And further narrowing of some parameter bounds might still be useful but will require a bit more research?

martinvonk (approved):
Seems good to me like this

raoulcollenteur:
> Only somewhat qualified to approve this PR. I'm guessing this narrows the bounds a bit which will result in more realistic and better models. And further narrowing of some parameter bounds might still be useful but will require a bit more research?

Yes I think so. I think a challenge for all parameters is to allow a range wide enough to cover many environments, but narrow enough to make sure most users get good model results by default. There is some tradeoff there. 

Merging this now.

Issue #1198, opened 2026-05-14
Can't get individual response for Tarsomodel and ChangeModel


raoulcollenteur:
### Describe the bug

Right now, not all response functions from the `ChangeModel` and `TarsoModel` can be obtained through `ml.get_block_response` and `ml.get_step_response`. I think it would be nice if that would be possible, through a fix in `ml._get_response` that both these methods use. 

Proposed plan:
- [ ] Create `get_response` or similar method in the stress models to be able to compute these responses
- [ ] Make `ml._get_responses` use the custom method (and not `sm.rfunc`) if available.
- [ ] Finally, we need to decide how to select a response. This is done through istress for the other stressmodels, but actually for these stress models it is not dependent on the number of stresses but rather a stressmodel characteristic. Still, using `istress` might work as well.

## To reproduce this issue
Make one of those models and do

 ml.get_block_response("recharge", istress=1).plot()
ml.get_block_response("recharge", istress=0).plot()

They will be the same but should not be. 

Related to #1199

raoulcollenteur:
#1156 introduces a `_get_responses` method to the stressmodels. A first step to solve this issue. Next step is to make sure this method is called from the Model `_get_response` method.

Issue #1165, opened 2026-04-17
ml._get_response not retrieving all the responses for for sm's with multiple responses


raoulcollenteur:
### Describe the proposed feature

Related to issue #1156, the ml.get_step/block_response methods are not able to return all the responses at once, which would be nice. 

### Expected usage
A clear and concise description of what you expect to happen.

```python
# Code to show the expected the behavior
rfuncs = ml.get_step_response

rfuncs ## multiple response in the case of ie. a wellmodel.
```

### Additional context
Add any other context about the feature here.

PR #1261, opened 2026-06-11
Improve ml.get_step_response and block_response to get all responses

raoulcollenteur:
# Short description
Many internal changes make ml._get_responses great again. This method can now be used to get all the response functions form the stress models, internally used for plotting but also useful for the users. Also made some of the plotting code easier with this implementation.

This also sets the default all_responses to None, and I think all the plots are by default now showing all the responses as we agreed in the pastas meeting. 

- [x] closes issue #1198, #1165, and #all_responses=None
- [x] is documented
- [x] type hints
- [x] tests added or expanded


pastas/stressmodels.py
```diff
@@ -346,18 +346,41 @@ def _get_block(
...
+        dt: float, optional
+            timestep for the response function.
+        istress: int, optional
+            When multiple stresses are present in a stressmodel, this keyword can be
+            used to obtain the response to an individual stress.
+        kwargs: dict: passed to rfunc.step() or rfunc.block()
+
+        """
+        if block_or_step == "step":
+            rfunc = self.rfunc.step
```
rubencalje: the name of rfunc is a bit confusing, as we also use this name for the response-classes. I would name in `resp_fcn` or something like that.

pastas/stressmodels.py
```diff
@@ -612,6 +635,51 @@ def _simulate(
         )
         return h
 
+    def _get_responses(
```
rubencalje: This method is the same as in StressModelBase right? Can we then not remove this?

pastas/stressmodels.py
```diff
@@ -738,6 +806,45 @@ def _simulate(
...
```
rubencalje, line 795: Can also be removed?

pastas/stressmodels.py
```diff
@@ -889,6 +996,36 @@ def simulate(
         trend = trend.cumsum() * p[0]
         return trend.rename(self.name)
 
+    def _get_responses(
```
rubencalje: If we check in StressModelBase if there is an rfunc defined (and if there is not, return None), we can remove this method.

pastas/stressmodels.py
```diff
@@ -956,6 +1093,36 @@ def simulate(p: float | None = None) -> float:
         """Simulate the stressmodel's contribution."""
         return p
 
+    def _get_responses(
```
rubencalje: If we check in StressModelBase if there is an rfunc defined (and if there is not, return None), we can remove this method.

rubencalje (changes requested):
Looks good. All code to calculate responses is in the stressmodels now, which is clearer. I would leave a check if there is an rfunc in the base-stressmodel class, so we do not need to override these methods in some of the stressmodel-classes.

I did not check if the all stressmodels are plotted correctly, but the code looks good!

pastas/stressmodels.py
raoulcollenteur: Good point, changed the name and made it consistent throughout

pastas/stressmodels.py
raoulcollenteur: Good point, done!

pastas/stressmodels.py
raoulcollenteur: Done

pastas/stressmodels.py
raoulcollenteur: p is different so keeping this one.

raoulcollenteur:
Thanks for the review! I made all the changes as suggested. Don't know what happened with the checks that fail. looks like another issue..

martinvonk:
Maybe one more check, does this now work for a ChangeModel with `ml.solver.get_response_ci`?

rubencalje (approved):

martinvonk:
From the docstring and the type hints it is now not clear when `istress` can be `"all"`. Also I think it is quite a lot of possibilities: `istress: int | None | Literal["all"] = None`. Maybe `list[int]` would be better than `"all"`? Otherwise use `int | Literal["all"] | None when appropriate.

martinvonk:
The order of the keyword arguments of `_get_responses` is also different per StressModel, overriding the BaseClass in an incompatible manner:
```
Method "_get_responses" overrides class "StressModelBase" in an incompatible manner
  Parameter 2 name mismatch: base parameter is named "block_or_step", override parameter is named "p"
  Parameter 3 name mismatch: base parameter is named "p", override parameter is named "dt"
  Parameter 4 name mismatch: base parameter is named "dt", override parameter is named "block_or_step"Pylance[reportIncompatibleMethodOverride](https://github.com/microsoft/pylance-release/blob/main/docs/diagnostics/reportIncompatibleMethodOverride.md)
```

martinvonk:
> Maybe one more check, does this now work for a ChangeModel with ml.solver.get_response_ci?

This does not work, from changing_responses.ipynb:
```python
# ChangeModel
ml2 = ps.Model(head, name="change")
ml2.add_noisemodel(ps.ArNoiseModel())

cm = ps.ChangeModel(
    river,
    ps.Exponential(),
    ps.Exponential(),
    tchange=t_step,
    name="riv",
)

ml2.add_stressmodel([cm, step])
ml2.solve(report=False, tmin=tmin, tmax=tmax)
_ = ml2.plots.results(figsize=(10, 6))

ml2.solver.ci_step_response("riv")
```

martinvonk (dismissed):
See comments above. I think these should be solved.

raoulcollenteur:
> > Maybe one more check, does this now work for a ChangeModel with ml.solver.get_response_ci?
> 
> This does not work, from changing_responses.ipynb:
> 
> ```python
> # ChangeModel
> ml2 = ps.Model(head, name="change")
> ml2.add_noisemodel(ps.ArNoiseModel())
> 
> cm = ps.ChangeModel(
>     river,
>     ps.Exponential(),
>     ps.Exponential(),
>     tchange=t_step,
>     name="riv",
> )
> 
> ml2.add_stressmodel([cm, step])
> ml2.solve(report=False, tmin=tmin, tmax=tmax)
> _ = ml2.plots.results(figsize=(10, 6))
> 
> ml2.solver.ci_step_response("riv")
> ```

That should now be 

`ci = ml2.solver.ci_step_response("riv", n=10, istress=0)
`
I updated the order of the arguments, good catch!


> From the docstring and the type hints it is now not clear when istress can be "all". Also I think it is quite a lot of >possibilities: istress: int | None | Literal["all"] = None. Maybe list[int] would be better than "all"? Otherwise use `int |  Literal["all"] | None when appropriate.

Since the stressmodels have no common method to inform about how many responses there are we need the "all" string option, so I am keeping it this way. It can always be "all" it will give all the responses.

martinvonk:
@dbrakenhoff checks if this can give some simplifications to the code of WellModel

raoulcollenteur:
@dbrakenhoff if the changes are internal only you can also do this after 2.0 to speed things up, and make a new issue.

dbrakenhoff:
Yea, I'll have a quick look and let you know here.

dbrakenhoff (approved):

dbrakenhoff:
looks good to me!

PR #1265, opened 2026-06-19
Create `pastas.options` for global configuration settings

### LINKED ISSUE #1204: Expand global settings configuration
It would be nice to expand the existing pastas.rcParams dictionary to support broader global (model) settings.
Some key settings that are excellent candidates for global configuration include:
- `seed`: Setting a global random seed for solvers/statistics to easily ensure reproducibility without passing it to every function call.
- `cache`: Globally enabling or disabling caching (acting as a global equivalent to ps.set_cache(True)).
- `numba`: Globally enabling or disabling numba
- `parallel`: Globally enabling or disabling parallelization inside Pastas.
- `use_block`: Globally toggling the use of block response functions instead of impulse response functions.

### Example of proposed usage:
```python
import pastas as ps
import numpy as np

# Set global settings
ps.rcParams["seed"] = 100

# Internal implementation example:
def prediction_interval(...):
    rng = np.random.default_rng(ps.rcParams["seed"])
    noise = rng.normal(0, np.std(ml.residuals()), data.shape) # example
    # ...
```

### Discussion points
- Should these global settings be saved when exporting a model to a .pas file? I can imagine seed being relevant to save for reproducibility of that specific model, while environment-specific settings like parallel, numba, or cache probably shouldn't be serialized into the model file.

- In Pandas, global configurations are handled via pd.options, which feels much more intuitive for Python users than rcParams (run commands parameters), which is a bit of a Matplotlib-specific legacy term. It might be worth considering aliasing or migrating rcParams to pastas.options as part of this enhancement.

martinvonk:
### Summary

This PR refactors the Pastas settings system to provide a cleaner, more maintainable API. The key changes are:

- `pastas.options` provides a user-friendly interface with both attribute-style and dict-style access for global settings in Pastas
- **Backward compatibility**: The deprecated `ps.rcParams` still works for now and delegates to the central to `options`

### Usage

**Public**
```python
import pastas as ps

# Attribute-style access
ps.options.seed = 42
print(ps.options.seed)  # 42

# All settings
print(ps.options.seed)       # 358183147
print(ps.options.cache)      # False
print(ps.options.numba)      # True
print(ps.options.parallel)   # False

**Internal:**
```python
from pastas._options import options

# Direct access to central storage
options.cache = False
```


pastas/_config.py
dbrakenhoff, line 1: Why not move everything to options and just import from there? Seems a bit like an extra file we don't really need?

pastas/_config.py
```diff
@@ -0,0 +1,77 @@
...
+        fill_before="mean",
+        fill_after="mean",
+        fill_nan=0.0,
+    ),
+}
+
+# Global settings dictionary
+global_settings: dict[str, Any] = {
+    # Global configuration settings
+    "seed": 358183147,
```
dbrakenhoff: Why a default seed set? Shouldn't we default to None?

pastas/decorators.py
```diff
@@ -339,16 +379,27 @@ def temporarily_disable_cache():
         with ps.temporarily_disable_cache():
             # Caching is disabled within this block
             ml.simulate()
+
+    .. deprecated::
```
dbrakenhoff: This context manager  was used when computing confidence intervals, because we definitely know that caching in that state is not useful. What is the proposed replacement? Just hard-coded calls to `ps.options.cache=False` in the code? And then put it back at the end? Any way we can reinstate these context managers with the new `ps.options` logic?

pastas/options.py
```diff
@@ -0,0 +1,220 @@
...
+    -----
+    The following settings are available:
+
+    - seed : int | None
+        Global random seed for reproducibility. If None, no seed is set.
+    - cache : bool
+        Global enable/disable caching for simulate() calls.
+        Requires cachetools to be installed.
+    - numba : bool
+        Global enable/disable Numba JIT compilation.
```
dbrakenhoff: Maybe elaborate the description to state that turning it off is mostly a debugging option, and not really recommended?

pastas/options.py
```diff
@@ -0,0 +1,220 @@
...
+    - parallel : bool
+        Global enable/disable parallelization.
```
dbrakenhoff: What does this actually control? Only the two `ps.stats._compute_ccf*` functions?

If it relates to numba parallel, I don't think flipping this setting after import will do anything. The decorator values are set at import, and cannot be flipped after without more custom code. We could consider setting `NUMBA_NUM_THREADS=1` to force single thread behavior. 

And I think this setting  is really only for debugging as well.

pastas/options.py
```diff
@@ -0,0 +1,220 @@
...
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from pastas._config import global_settings
+
+
+class OptionsContainer:
```
dbrakenhoff: Can you inherit from Dict to avoid having to program all the special methods yourself? And then add on setting and getting from attributes as well?

dbrakenhoff (commented):
Some questions/comments in the comments. 

Also, is pickling support necessary at all? I don't think any of these settings should affect model results, and therefore do not need to be saved in any way. They just control pastas behavior, not outcomes (hopefully). Unless seeding the random generators is considered something you would want to persist...?

The exception would be `use_block=True`  which isn't yet added here. And that makes sense I think.  In that case the idea would be to have a global setting to modify defaults which would indeed require a default value of use_block=None, to respect the global setting inside the method call. If we want this, this setting does affect outcomes and would need to be saved in the pas file (which it already is). But I would agree that this can just be set in the rfuncs by the users themselves.

pastas/_config.py
martinvonk: I asked my new `Vibe` friend the same question and he was very much against this because he wanted a clear seperation between the internal storage (_config) and public API (options). But personally I don't mind having everything in one file.

pastas/options.py
martinvonk: At the moment yes. But in the future maybe more. So that's why i wanted to make this a global setting like this. This way it can also help with PastaStore where you can then enfore pastas.option.parallel=False always.

pastas/_config.py
martinvonk: I think it is nice to have a SEED. pyemu has this as well which is aslo the same int. It wil help with the Pest solvers in pastas-plugins as well. There you now have a seed attribute to each solver where this seed can be the default.

dbrakenhoff: Are there examples where it is annoying to RNG the same set of random numbers each time? I'm fine with a global seed setting, but wondering whether there is a situation where this would not be what you want?

pastas/stats/core.py
```diff
@@ -331,7 +326,7 @@ def _preprocess(
     return x, t, dt_mu
 
 
-@njit(parallel=True, nogil=True, cache=True)
+@njit(parallel=options["parallel"], nogil=True, cache=options["cache"])
```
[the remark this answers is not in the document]
dbrakenhoff: Yeah, this cache setting should not control numba behavior.

pastas/decorators.py
dbrakenhoff: If you really want to control parallel behavior after import, I think you have to add in the global options value here. So decorate functions with parallel=True, then in our custom decorator, check for the presence of the parallel kwarg, and use the global setting.

dbrakenhoff (commented):
Currently, some of these options control too much and should be more scoped, in my opinion. Also these settings should all support runtime changes (i.e. after import), otherwise there is no reason to expose them to the user. 

- Seed is fine
- Cache should only control simulate caching and nothing else. We don't need to control numba caching.
- Parallel should not control both numba compiling settings, and parallel flags within functions. So we have to choose what this controls. For the latter, using the global values in function definitions will mean the global setting is evaluated at import, and then will not respect changes to the global setting after. So i would remove the parallel option from those functions and just let users decide when they call the function. For numba decorators, the functions can be decorated with parallel=True, but in our custom `njit` decorator we would need to overwrite that with the global setting. 
- timeseries is fine

martinvonk:
Okay so @dbrakenhoff and I had a good chat. We agreed that options should only change Pastas' behavior, but not the outcomes. So that's why seed is now not part of this PR. If we want random stuff to be reproducible we should add a seed attribute to the solver, or maybe to the model in a later pr. 

Also I changes the dict like behavior of `options` because a class that sometimes behaves like a dict is confusing.

raoulcollenteur:
Looks nice. One thing that I think would be good to discuss is whether or not the time series settings is an option. I feel like this should be somewhere else, perhaps as simple as ps.timeseries.settings. When we change the location of the predefined settings, I think I would not put it in options. So what about splitting the timeseries settings dict from the options? (Moving the settings dict to another place could be another PR if you want to make this PR simpler.)

raoulcollenteur (approved):

martinvonk:
> Looks nice. One thing that I think would be good to discuss is whether or not the time series settings is an option. I feel like this should be somewhere else, perhaps as simple as ps.timeseries.settings. When we change the location of the predefined settings, I think I would not put it in options. So what about splitting the timeseries settings dict from the options? (Moving the settings dict to another place could be another PR if you want to make this PR simpler.)

I think having the options / global settings in different places is also a bit cumbersome. But I don't know.. What do you think @dbrakenhoff and @rubencalje ?

mbakker7:
From a general perspective, I would think all settings in one place. But that is also because I don't understand the advantage of putting the timeseries settings somewhere else.

martinvonk:
Okay this is now done after a pastas meetings. We now have:
`ps.options` for the options for parallelization etc.

`ps.timeseries.SETTINGS` for the time series settings. For this I also considered a function: ps.timeseries.get_default_settings` but that would be more verbose.

martinvonk:
Can you review this again @dbrakenhoff ?

pastas/timeseries.py
```diff
@@ -35,6 +34,62 @@
 
 logger = getLogger(__name__)
 
+SETTINGS = {
```
raoulcollenteur: why not lowercase? Uppercase looks really ugly :p

dbrakenhoff:
I agree lower case is prettier. Also a few more remarks:

1. The numba accelerated functions `_compute_ccf_rectangle` and `_comput_ccf_gaussian` should have `parallel=True` always. These functions will benefit from this I think and there really isn't any reason to turn it off. This is not something we want to let the users configure. Question is how do we do that, with the global option being inserted into the decorator... 
2. The other thing is that modfying `ps.timeseries.settings` will modify the dictionary for the rest of that session. Also when you access e.g. `a = ps.settings.timeseries["prec"]` and then modify `a`, the original is also altered. That was always the case I guess, but I think it would be nicer to ensure a copy of the settings dictionaries is returned. Maybe something along these lines:

```python
class CopyDict(dict):
    """A dict that returns a copy of the value when accessed."""

    def __getitem__(self, key):
        return super().__getitem__(key).copy()

settings = CopyDict(_settings)
```

The only "downside" when implementing above is that this fails silently:
```python
ps.timeseries.settings["prec"]["fill_after"] = "i am silently ignored..."
```

But I can live with that and think it is nice that the sub dictionaries are always returned as as copy.

martinvonk:
> 2\. The other thing is that modfying `ps.timeseries.settings` will modify the dictionary for the rest of that session. Also when you access e.g. `a = ps.settings.timeseries["prec"]` and then modify `a`, the original is also altered. That was always the case I guess, but I think it would be nicer to ensure a copy of the settings dictionaries is returned. Maybe something along these lines:

Maybe we should just make a function then? instead of the difficulty with CopyDict:

def get_default_settings(name: Literal["oseries", "prec", "evap", "..."]) -> OseriesSettingsDict | StressSettingsDict:
    ...
    return SETTINGS[name]

martinvonk:
> * The numba accelerated functions `_compute_ccf_rectangle` and `_comput_ccf_gaussian` should have `parallel=True` always. These functions will benefit from this I think and there really isn't any reason to turn it off. This is not something we want to let the users configure. Question is how do we do that, with the global option being inserted into the decorator...

That's the whole point about the parallel toggle right? If you have something around pastas or _compute_ccf_rectangle. You don't want to have everything in parallel, fighting for cores, which slows things down.

raoulcollenteur:
I suggest making another issue to address that issue. It was like this before so it is new and does not have to be fixed here and can be done after 2.,0 as well.

dbrakenhoff:
> That's the whole point about the parallel toggle right? If you have something around pastas or _compute_ccf_rectangle. You don't want to have everything in parallel, fighting for cores, which slows things down.

Right, fair point, forgot about people actually running pastas in parallel. But in that case the Pastas default should be true? Most users won't be running other parallel tools around pastas, and if they are, they will probably be aware that turning it off in pastas is smart? We can make pastastore check this when doing parallel apply calls. Just seems a bit of a waste to not use the free performance if you can have it. But maybe we should check first how much performance gain parallel=True gives for the ccf functions?

raoulcollenteur:
I changed `SETTINGS` to `settings`, but it needs some additional logic because the methods also have a settings argument. However, this works now.

`ps.timeseries.settings`

@martinvonk what do you think? Otherwise we could discuss making a new function:

`ps.timeseries.get_default_settings() `

Which returns a copy and solve this option as well (https://github.com/pastas/pastas/issues/1289).. I can see that might be a preferred option but it is different from what we agreed with the others.

@dbrakenhoff , @mbakker7 @rubencalje what would you prefer?  

`ps.timeseries.settings` or `ps.timeseries.get_default_settings()`

ps. would be nice if Github had a vote option :)

pastas/_options.py
```diff
@@ -0,0 +1,17 @@
...
+from dataclasses import dataclass
+
+
+@dataclass(slots=True)
+class PastasOptions:
+    """Central configuration storage for Pastas global settings."""
+
+    cache: bool = False
+    numba: bool = True
+    parallel: bool = False
```
dbrakenhoff: default to True!

martinvonk:
> `ps.timeseries.get_default_settings() `

I'd say lets just make this function. The default settings were never really made for adjusting. Otherwise we wouldn't allow for the `StressSettingsDict` and `OseriesSettingsDict` to be parsed. If we now change the location, and later decide to make it a function later that's also awkard. So I'd say, just make the function `ps.timeseries.get_default_settings(kind: Literal["prec", "evap", "..."]) -> StressSettingsDict | OseriesSettingsDict:`. Thats easy and clean and more understandable than an immutable dict.

pastas/timeseries.py
```diff
@@ -106,16 +164,14 @@ class TimeSeries:
...
-    >>> ps.rcParams["timeseries"]
+    >>> ps.timeseries.settings
 
     See Also
     --------
     pastas.timeseries.TimeSeries.update_series
         For the individual options for the different settings.
     """
 
-    _predefined_settings = rcParams["timeseries"]
```
dbrakenhoff, line 117: Instead of the module alias, you could use the internal reference to distinguish between internal `settings` and the settings dictionary?

pastas/decorators.py
```diff
@@ -283,16 +326,17 @@ def njit(function: Callable | None = None, **kwargs) -> Callable:
...
         try:
-            if not USE_NUMBA:
+            if not options.numba:
                 return f
             else:
                 from numba import njit
 
-                fnumba = njit(f, **kwargs)
-                return fnumba
+                njit_partial = partial(njit, parallel=options.parallel)
```
dbrakenhoff: We now insert options.parallel into each numba decorator. It might not hurt all that much but we know some numba accelerated functions are not parallel, so setting parallel=True in those functions is a bit weird. Should we not keep the parallel=True decorator in the njit decorators, then replace the setting with our global parallel option instead? That way only truly parallelizable functions can be controlled with our global option, and the rest is left alone?

dbrakenhoff:
> ps. would be nice if Github had a vote option :)

You can launch polls in the discussions area: https://github.com/orgs/pastas/discussions/new?category=polls

I made some changes, according to my comments, hopefully you guys agree this is the way to go?

> @martinvonk what do you think? Otherwise we could discuss making a new function:

>Which returns a copy and solve this option as well (https://github.com/pastas/pastas/issues/1289).. I can see that might be a preferred option but it is different from what we agreed with the others.

This we can discuss as part of #1289 I think.

raoulcollenteur (approved):

martinvonk:
~~Maybe another radical idea: how about making the default settings only available via a property of the TimeSeries class: `ps.TimeSeries.settings`. Then we solve both in one go.~~

Nevermind, this won't work because settings is already an attribute.

Issue #735, opened 2024-04-11
The `noisemodel` keyword argument in ps.Model and `noise` keyword argument in ml.solve are Deprecated as of Pastas 1.5


raoulcollenteur:
This thread is used to explain users the change of deprecating the `noisemodel` in `ps.Model` and `noise` argument in `ml.solve` and `ml.initialize` methods. The error messages are linked to this issue, which is pinned at the top of the issue page.

# Changes to default behavior regarding noise models in pastas>=1.5

Noise models are no longer added to time series models by default. From Pastas 1.5
onwards we are forcing users to explicitly add a noise model to their time series model
should they want to use one.

> [!WARNING]  
> This means that if you are running old code with the new version of
Pastas and haven't specified anything with respect to the noise model in your code
(thereby relying on Pastas defaults) the _**result will change**_ !

## Contents

- [Remind me, what is a noise model?](https://github.com/pastas/pastas/issues/735#user-content-1)
- [Why are we changing the default in Pastas?](https://github.com/pastas/pastas/issues/735#user-content-2)
- [Overview of changes](https://github.com/pastas/pastas/issues/735#user-content-3)
- [Guide for updating to Pastas 1.5](https://github.com/pastas/pastas/issues/735#user-content-4)

<h2 id="1">Remind me, what is a noise model?</h2>

A noise model can be used to try to reduce the residuals (simulation minus
observations) to white noise. When the noise (the result after applying the noise model
to the residuals) meets the criteria of white noise, this means we can make certain
statistical inferences with our time series model. For example, when the noise meets
these criteria, the estimated parameter uncertainty is reliable and can be used to
quantify model uncertainty. For more information about the noise model, check out the 
following notebooks:

- [Diagnostic checking](https://pastas.readthedocs.io/latest/examples/diagnostic_checking.html)
- [Timestep analysis for reducing autocorrelation](https://pastas.readthedocs.io/latest/examples/timestep_analysis.html)
- [Uncertainty quantification](https://pastas.readthedocs.io/latest/examples/uncertainty.html)

<h2 id="2"> Why are we changing the default in Pastas?</h2>

There are several reasons:

 - A noise model is not always necessary in a time series model.
 - We want users to explicitly choose whether they want to use a noise model. This means an extra line of code is sometimes necessary, but this makes it much clearer what the modeler is doing.
 - There are other methods for estimating the uncertainty of the model parameters that
[...]

PR #1270, opened 2026-06-30
Update way to add model components

### LINKED ISSUE #977: Should `add_stressmodel` be removed?
Adding a stress model to a pastas model has always been a two-step approach: First create a stress model, then add it to the pastas model. The idea was that the same stress model (i.e., recharge) can be used for a multiple pastas models. The downside is that the stressmodel doesn't know what pastas model it is added to, which may create undesirable inputs. For example: if the stress model actually depends on the simulated head, or uses the observed head to set initial parameters, then the head needs to be specified for the stress model. I don't think that is what we want. Especially now that we are working on stress models with feedback loops (the simulated head depends on the head), this becomes undesirable. Besides, I wonder how often users create one stress model and then assign it to multiple pastas models. 

### Proposed usage
I propose to add a stress model to a pastas model upon creation. This means the first argument of a stress model is the pastas model to be added to (this is the same syntax as for, for example, a modflow model using flopy, or a tim analytic element model). 

This would obviously be a significant usage change, but would make the scripts shorter and, most importantly, prepare us for future enhancements of pastas that will contain feedback loops. I propose this change for Pastas 2.0.

raoulcollenteur:
# Short description

This PR changes the way the models are constructed, deprecating add_stressmodels, add_constant, add_transform, and add_noisemodel. The new way is similar to Flopy and Timflow, adding model as the first argument of the component that is added. The model is stored so noise, stressmodels, constant, transforms get access to the model. Internally the components are added to the model using _add_method. The former public add_methods are deprecated. 

A warning is thrown when the first argument is not model, pointing to the new behavior. 

Quite the change, but I kinda like it and I think it will be liked more after a while of getting used to it :)

- [x] closes issue #977 
- [x] is documented
- [x] type hints
- [x] tests added or expanded


raoulcollenteur:
I think this is ready for review. One thing I am not sure about is whether we want to use `model` or `ml` as the argument name, thoughts? 

I took me a while to get used to the new way, but I now think it is actually nice.

mbakker7:
Great to see this is moving forward. It will make it easier to make Pastas models and it will open new doors for future development. Stressmodels will know what model they are added to so they have access to the head series. Parameters can be stored with the stress models. Etc. Those new developments can all be added when Pastas 2.0 is out with these new changes.

mbakker7:
I started with the `pastas_modeling` notebook.
I first tried
```
ml = ps.Model(ho, name="first_model")
sm1 = ps.StressModel(stress=recharge, rfunc=ps.Gamma(), name="recharge", settings="prec")
```
And got the nice error
```
TypeError: StressModel.__init__() missing 1 required positional argument: 'model'
```
So I fixed that. Then it looked like
```
ml = ps.Model(ho, name="first_model")
sm1 = ps.StressModel(model=ml, stress=recharge, rfunc=ps.Gamma(), name="recharge", settings="prec")
ml.solve(tmin="1985", tmax="2010", solver=ps.solver.LeastSquares())
```
But now I get the error
```
TypeError: LeastSquares.__init__() missing 1 required positional argument: 'model'
```
I don't think that should be the case. But I tried it anyway
```
ml = ps.Model(ho, name="first_model")
sm1 = ps.StressModel(model=ml, stress=recharge, rfunc=ps.Gamma(), name="recharge", settings="prec")
ml.solve(tmin="1985", tmax="2010", solver=ps.solver.LeastSquares(model=ml))
```
But then I got a longer error message
```
TypeError: least_squares() got an unexpected keyword argument 'callback'
```
So it is not quite working correctly yet.
Maybe I have to update Scipy, but it doesn't work with `LmfitSolve` either. Why does a solver need a model instance anyway? Even if it would need it, then the model can pass it, as a syntax `ml.solve(..., solver=Solver(model=ml)` is odd as it passes `ml` twice.

raoulcollenteur:
Thanks for trying the new code, interesting finding with the solver, I can't reproduce it locally, which spicy version do you have?

Perhaps we should discuss if we want to add the solver in this new way as well or stick to the old method. I think for many solvers however we actually want to pass the model as an argument upon initialization. So another option would be to change to `solver` argument in ml.solve, which seems a bit pointless to me anyway by now... @martinvonk you have thoughts about how to add the solver?

mbakker7:
All right this works:
```
ml = ps.Model(oseries=ho, name="first_model")
sm1 = ps.StressModel(model=ml, stress=recharge, rfunc=ps.Gamma(), name="recharge", settings="prec")
ps.solver.LeastSquares(model=ml) # optional
ml.solve(tmin="1985", tmax="2010")
```
Installed:
```
Pastas     : 2.0.0
Python     : 3.13.14
Numpy      : 2.4.6
Pandas     : 3.0.3
Scipy      : 1.18.0
Matplotlib : 3.11.0
Numba      : 0.65.1
```

I get the `LeastSquares` error with
```
Pastas     : 2.0.0
Python     : 3.13.5
Numpy      : 2.1.3
Pandas     : 2.2.3
Scipy      : 1.15.3
Matplotlib : 3.10.0
Numba      : 0.61.0
```

raoulcollenteur:
I think the best option is to deprecate the `solver` argument in ml.solve now. That way we have one approach to add a solver which is most clear to users I think and most future-proof. 

The `callback` error was from spicy, it was added in 1.16 :) https://docs.scipy.org/doc/scipy-1.16.0/reference/generated/scipy.optimize.least_squares.html

mbakker7:
All right. Just changed the scipy requirements to at least 1.16

mbakker7:
So we want to deprecate setting the solver in `ml.solve` and I agree. 
We could make it still work with just `solver=ps.solver.LeastSquares()` rather than `solver=ps.solver.LeastSquares(model=ml)`. This means that the default model in the solver instances should be `None`. In the `Model.solve` function we must then check whether the solver has a model defined and if not, we define it there. This can all be removed when the deprecation warning runs out. It is a bit ugly, but otherwise there is no deprecation. 
The other option is to just implement this and throw an intelligent error when the `solver` keyword is used in `ml.solve`. I actually like that better. This is for 2.0 anyway.

raoulcollenteur:
Oke I made the final changes and I think the PR is ready for review now. To summarize:

- Model components are now added to the model by adding the model as the first argument (ie. ps.Stressmodel(ml..)
- A warning is given if the first argument is not a Model, pointing the user to the new code pattern
- `solver` is deprecated and solvers are added in the same pattern (like was already the case for some ie. mcmc).
- The old ml.add_methods will throw an error and point the user to the new pattern. I don't think there's a good way to allow both the old and new approach, so I think a hard cut is unfortunately necessary, but that is also why it is 2.0?
- Scripts get shorter now, and new possibilities arise as all model components have direct access to the mother model.

Looking forward to all your thoughts.

mbakker7:
This is really a nice change and an important step forward.

rubencalje:
I think this is a waste of our time, and of the current users of pastas that are used to the way model elements are added.

Besides, I do not understand why is is so hard to make this backward compatible. When a user does not add a model-class as the first argument of a stress model, we can show a warning. As soon as he calls `ml.add_stressmodel` the model should call a method like `_add_to_model` of the stress model, which otherwise would have been called by the `__init__` of the stress model.

mbakker7:
I understand this is a change for the users. But a good change: scripts are shorter. And no strange issues where certain stress models need a model and others don't.

martinvonk:
> Besides, I do not understand why is is so hard to make this backward compatible. When a user does not add a model-class as the first argument of a stress model, we can show a warning. As soon as he calls 'ml.add_stressmodel' the model should call a method like '_add_to_model' of the stress model, which otherwise would have been called by the 'init' of the stress model.

Can you elaborate a bit more on what this (path) would look like @rubencalje ? Am I correct that you're suggesting that: In 2.0 you would add the model keyword argument in StressModel.__init__? and then in 3.0 you would have a change where `model` is the first argument? 

I like this change in the sense that I think that it is necessary. I don't like that it is such a big breaking change. It reminds me of removing the noisemodel by default in 1.0. There we handled it by creating a big warning when importing pastas. Additionally we created a long-read issue #https://github.com/pastas/pastas/issues/735 to explain the change and the reasoning behind it. Maybe we should to that here as well.

rubencalje:
I would still add the `Model` argument as the first argument of StressModels, but use a decorator to make sure the code functions if the first argument is not a `Model` (and the user uses `add_stressmodel` to set the model later on). This decorator needs to intercept the arguments before the original `__init__` is called.

The basic idea is:

1. Inspect the first positional argument.
2. If it is a `Model`, call the wrapped function unchanged.
3. Otherwise, insert `None` as the first argument so that the remaining arguments shift one position.

For example:

```python
def optional_model_argument(func):
    @wraps(func)
    def wrapper(self, *args, **kwargs):
        if "model" in kwargs:
            return func(self, *args, **kwargs)

        if args and isinstance(args[0], Model):
            return func(self, *args, **kwargs)

        logger.warning("In Pastas 3.0, the first argument of a stressmodel has to be a pastas.Model, blablabla")
        return func(self, None, *args, **kwargs)

    return wrapper
```

Then

```python
class StressModel:

    @optional_model_argument
    def __init__(self, model, stress, rfunc, name=None):
        self.model = model
        ...
```

Both of these work:

```python
# New API
sm = StressModel(model, stress, rfunc)

# Old API
sm = StressModel(stress, rfunc)
```

mbakker7:
Could you explain what the real reason is that you want to be able to do both? Am I correct to interpret that you agree that removing the necessity to do `add_stressmodel` is nice, but you want to give the user some time to get used to it before we completely remove it? If so, then the user can also wait a bit with updating to 2.0 until he/she is comfortable with this new feature and then go to 2.0.

rubencalje:
`add_stressmodel` is a bit longer, but was logical all these years, and still is to a lot of users. I do not think this needs to be changed. If a stressmodel needs a model, we can check if the stressmodel was added to a Model or not. This does not require the basis syntax change that you propose.

But if you do want to change the syntax, I think you should first show a warning for a while, so users have time to see the warning before their scripts fail. Therefore, you need to be able to do both.

rubencalje:
I also think `add_stressmodel` is very clear about the action you are performing. It is also clear this can be undone with `del_stressmodel`.

martinvonk:
Just to be clear, you say:

> Both of these work:
> ```python
> # New API
> sm = StressModel(model, stress, rfunc)
> 
> # Old API
> sm = StressModel(stress, rfunc)
> ```

But would
```python
sm = StresssModel(stress=stress, rfunc=rfunc)
```
still raise an error because `model` is not provided? Or would you suggest making all keyword arguments in the `__init__` of the StressModels `None`?

rubencalje:
> Just to be clear, you say:
> 
> > Both of these work:
> > ```python
> > # New API
> > sm = StressModel(model, stress, rfunc)
> > 
> > # Old API
> > sm = StressModel(stress, rfunc)
> > ```
> 
> But would
> 
> ```python
> sm = StresssModel(stress=stress, rfunc=rfunc)
> ```
> 
> still raise an error because `model` is not provided? Or would you suggest making all keyword arguments in the `__init__` of the StressModels `None`?

`sm = StresssModel(stress=stress, rfunc=rfunc)` is equivalent to `sm = StressModel(stress, rfunc)`, so would also work. The current keyword arguments are not changed.

martinvonk:
> sm = StresssModel(stress=stress, rfunc=rfunc) is equivalent to sm = StressModel(stress, rfunc), so would also work. The current keyword arguments are not changed.

I see now, you catch this behavior in the decorator.

pastas/model.py
```diff
@@ -976,16 +979,16 @@ def solve(
...
-        elif self.solver is None:  # add scipy least_squares if no solver provided
+            msg = "solver argument is deprecated since Pastas 2.0. Please use ps.solver.LeastSquares(ml) pattern instead to add a solver."
+            logger.error(msg)
+            raise KeyError(msg)
+
+        # Add default solver if none is provided
+        if self.solver is None:  # add scipy least_squares if no solver provided
             logger.debug("Adding LeastSquares as default solver.")
-            self.add_solver(solver=LeastSquares())
+            self._add_solver(solver=LeastSquares(model=self))
```
rubencalje: So why not just `LeastSquares(model=self)`? That would work for stressmodels, but for solvers, we did not choose to implement it the same way (?), so we still do `self._add_solver(solver=LeastSquares(model=self))`, which feels a bit double, having to specify the model (`self`) twice.

Why is that? Just curious.

raoulcollenteur: Good point, that is unnecessary so we should change that.

rubencalje:
I made some small final changes, and I think it is ready now. The only thing that is still ugly is that the pastas Model of the Solver class is named `ml`. We set the attribute `model` now as well, but do not use it. I think we should use `model` everywhere in the solver class, and make `ml` a property with a depreation-warning. I can do it here, but we can do that in another PR as well.

raoulcollenteur:
> I made some small final changes, and I think it is ready now. The only thing that is still ugly is that the pastas Model of the Solver class is named `ml`. We set the attribute `model` now as well, but do not use it. I think we should use `model` everywhere in the solver class, and make `ml` a property with a depreation-warning. I can do it here, but we can do that in another PR as well.

Let's do this in this PR! I can do this later but if you're already on it that would also be good of course :)

raoulcollenteur:
I made all the changes and double-checked the old pattern still works on some other scripts. All seems good to go now and I'll merge this into Dev after the tests ran successfully :)

raoulcollenteur:
@rubencalje I need a review before I can merge? You have time?

rubencalje (approved):
I quickly looked at your latest addidtions, and they seemed to do exactly what we want. Great!

Issue #1076, opened 2025-12-15
Pastas naming conventions


martinvonk:
### Describe the task
Make sure we are consistent with Pastas naming conventions

raoulcollenteur:
This is an Issue not a PR ;-)

martinvonk:
Found another one in the timeseries utils where we use both `index` and `tindex` for a `pandas.DatetimeIndex`

raoulcollenteur:
Guess we can close this after merging the last PR

PR #1277, opened 2026-07-16
Rename `index` argument to `tindex` for DatetimeIndex

martinvonk:
# Short description
Fix for DatetimeIndex that we call `index` and `tindex` across the codebase.

- [x] closes issue #1076


martinvonk:
I did not add a deprecation here because it is relatively new @rubencalje. Is that fine by you?

rubencalje:
> I did not add a deprecation here because it is relatively new @rubencalje. Is that fine by you?

fine

rubencalje (approved):

PR #1278, opened 2026-07-16
Support the old way to add a stressmodel/noisemodel/transform/constant/solver again

rubencalje:
# Short description
This PR makes sure the changes in PR #1270 are backwards compatible. So a stressmodel/noisemodel/transform/constant/solver can be generated without specifying the Pastas Model as the first argument, and ml.add_stressmodel, ml.add_noisemodel, etc function again. With both functionalities, the user will be shown a warning, that together will inform the user of the new way to add these model elements (by specifying the model as the first argument).

- [x] a test is added to test the old way of adding a stressmodel: `test_add_stressmodel_indirectly`

### DIFF
--- a/pastas/decorators.py
+++ b/pastas/decorators.py
@@ -130,18 +130,15 @@ def check_argument_model(function: Callable) -> Callable:
 
     @wraps(function)
     def _make_model_component(self, *args, **kwargs):
-        if args:
-            from pastas.model import Model
-
-            ml = args[0]
-            if not isinstance(ml, Model):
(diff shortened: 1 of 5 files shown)


rubencalje:
Hmm, because I am not tagetting the dev-branch, but the other PR, the test are not running. @raoulcollenteur, if you are ok with these changes, maybe we can merge it and then I fix possible broken tests in your PR?

raoulcollenteur (approved):
Should be 2.4 I think but let's change that in the other PR.

PR #1284, opened 2026-07-17
Move notebook execution from ReadTheDocs to GitHub Actions

### LINKED ISSUE #1259: Pass artifacts from GH Action to RTD
### Describe the task

We are now running the notebooks both on GH Actions and RTD. This doesn't make any sense, so we should figure out how to pass on the results from GHA to RTD. @dbrakenhoff mentioned this is how flops does it as well so we can take a look at that.

raoulcollenteur:
## Summary

Move notebook execution from ReadTheDocs to GitHub Actions to avoid duplicate execution.

## Changes

- **New workflow**: `.github/workflows/execute_notebooks.yml` executes all example and benchmark notebooks in GitHub Actions and uploads them as artifacts
- **Updated `doc/conf.py`**: Added `rtds_action` extension and disabled notebook execution in RTD (`nb_execution_mode = "off"`)
- **Updated `.readthedocs.yaml`**: Added `rtds_action` dependency and `GITHUB_TOKEN` environment variable
- **Updated `pyproject.toml`**: Added `rtds_action` to RTD dependencies and updated tox `all` environment to exclude notebooks
- **Updated `test_unit_pytest.yml`**: Updated job name to reflect that notebooks are now excluded

## How It Works

1. GitHub Actions executes all notebooks and uploads them as an artifact named `notebooks-for-<commit-sha>`
2. ReadTheDocs uses the `rtds_action` Sphinx extension to download the artifact during the build
3. The executed notebooks overwrite the original notebooks in the repository
4. `myst_nb` uses the pre-executed notebooks (since `nb_execution_mode = "off"`)

## Benefits

- Notebooks are only executed once (in GitHub Actions)
- Faster ReadTheDocs builds
- Consistent execution environment
- Better error handling (failures caught in CI before RTD build)

## Notes

- The `test_benchmark_notebooks.yml` workflow is still active but is now redundant. It can be removed in a future cleanup.
- Local documentation builds will work without the artifact (`rtds_action_error_if_missing = False`)
- The executed notebooks are retained for 7 days in GitHub Actions artifacts


raoulcollenteur:
Starting to suspect to might actually work :)

.github/workflows/test_unit_pytest.yml
```diff
@@ -42,7 +42,7 @@ jobs:
             python: "3.14"
             toxenv: py314
             experimental: false
-          - name: Test suite for all unit tests including Notebooks
+          - name: Test suite for all unit tests (excluding notebooks)
```
martinvonk: if you exclude the notebooks in the pytest routine, this will generate a lower coverage

rubencalje:
Great idea! Is this still work in progress?

Currently, the documentation for this PR does not show any output in the notebooks, but you probably know this allready.

raoulcollenteur:
Yes! Just one last piece of the puzzle is to trigger the RTD build after the execute_notebook ran with another scheduled GH Action, and then this should actually be done :)

raoulcollenteur:
Now it should trigger the RTD built after the notebooks are excuted, let's see..

dbrakenhoff (approved):
Looks good! Is it working properly now?

raoulcollenteur:
uploading the artifacts yes, but somehow triggering the built not yet... this seems like a hard nut to crack if I'm honest. Perhaps a better way is to use GH CI to also built the Docs and then upload to RTD... but then I think we might loose the nice options from RTD...

To be continued. Might not make 2.0 though.

pyproject.toml
```diff
@@ -132,7 +133,7 @@ commands = [
 ]
 
 [tool.tox.env.all]
-description = "run all unit tests and notebooks"
+description = "run all unit tests (excluding notebooks)"
```
martinvonk, line 136: It's fine to just not run the notebooks anymore and accept a lower coverage percentage. But this testing routine is now the same as py311-py314. So then we should just get rid of this testing routine.

martinvonk (changes requested):
Only one comment about the testing routine for the coverage that we should address

raoulcollenteur:
I converted it back to draft because I just can;t get it to work for now. But for future reference:

1. run GH Action to run execute the notebooks and create artifacts
2. Run GH Action to start the built on RTD, somehow using the artifacts

Seems the most obvious way. But configuring this is not obvious.

We might also close this PR because nothing substantial is done in the end here.

martinvonk:
Can't you upload the html artifact to a branch. That's how GitHub pages does it as well. And then let readthedocs download the html from the branch?

raoulcollenteur:
Any luck here @martinvonk ?

raoulcollenteur:
@martinvonk shall we just close this PR? I feel like if we want to do this it's better to start fresh and only copy what we learned here.

PR #1290, opened 2026-07-17
add first draft of pastas 2.0 note

dbrakenhoff:
Do not merge!

This is just to work on the note to accompany the release of pastas 2.0

### DIFF
--- a/doc/benchmarks/check_response_functions.ipynb
+++ b/doc/benchmarks/check_response_functions.ipynb
@@ -1483,7 +1483,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "pastas (3.13.5)",
+   "display_name": "Python 3",
    "language": "python",
    "name": "python3"
   },
@@ -1497,7 +1497,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.13.5"
+   "version": "3.13.11"
   }
  },
  "nbformat": 4,
--- a/doc/examples/changing_responses.ipynb
+++ b/doc/examples/changing_responses.ipynb
@@ -281,7 +281,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "pastas_dev",
+   "display_name": "Python 3",
    "language": "python",
    "name": "python3"
   },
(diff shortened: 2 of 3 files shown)


rubencalje:
Looks good! One thing that needs to be included:
- the model settings are reused when you call the `solve` method. So when you do `ml.solve(tmin='2000')` the next time you call `ml.solve()`, tmin is set to '2000'. In the past tmin would be set back to the default value (the first date of your oseries).

mbakker7:
Are you sure that the links at the top of the note work? They didn't seem to work on my computer. I could click on them, but nothing happened.

dbrakenhoff:
> Are you sure that the links at the top of the note work? They didn't seem to work on my computer. I could click on them, but nothing happened.

I think they might work once I post them, but not entirely sure yet :).

mbakker7:
> > Are you sure that the links at the top of the note work? They didn't seem to work on my computer. I could click on them, but nothing happened.
> 
> I think they might work once I post them, but not entirely sure yet :).

I read up on that a bit, and it sounds like it will work on Github. And this won't be on ReadTheDocs, will it? Not entirely sure that it will work there.

dbrakenhoff:
> And this won't be on ReadTheDocs, will it? Not entirely sure that it will work there.

Yes, only github!

martinvonk:
Note to self before i get of the bus: add something on the stresses in StressModels no longer being stored in the stress list attribute, but under the keyword argument name.

martinvonk:
Maybe we can turn this into a discussion now that the release candidate is out? We can still edit discussions I think. Also you we can add a small note on how to install a release candidate:
`pip install --pre pastas`

dbrakenhoff:
Yeah do we want to do that now with release-candidate info, or do we wait till the official 2.0 release and communicate it then? I'm fine with both, and good idea to make it a pinned discussion. If we decide on a strategy, I can make the discussion, I think I still have some local edits to the note at the moment.

martinvonk:
> Yeah do we want to do that now with release-candidate info, or do we wait till the official 2.0 release and communicate it then? I'm fine with both, and good idea to make it a pinned discussion. If we decide on a strategy, I can make the discussion, I think I still have some local edits to the note at the moment. 

I'd prefer to do it now. Then people installing the release candidate know what has changed via the overview. Otherwise it is pretty hard to find.

Maybe we can also actively communicate the discussion for pastas 2.0 and ask people to test the release candidate.

martinvonk:
Can we close this now since https://github.com/orgs/pastas/discussions/1324 is live?

Issue #1168, opened 2026-04-19
Make TarsoModel `d1` and ThresholdTransform `1` an additive parameter


martinvonk:
### Describe the proposed feature
Currently, the parameters `d1` and `1` from TarsoModel and ThresholdTransform are absolute parameters. This means that they indicate the absolute level above which another response / damping occurs. 

However, I think it would be better if the parameter would be an additive parameter to `constant_d`/`d0`. This makes it easier to provide initial bounds that indicate, for instance, that the level should always be above the constant_d. Now it can be the case that the d1 and 1 parameter are lower than the d0 and constant_d.

So for the TarsoModel: d1_old = d0 + d1_new
And for ThresholdTransform: 1_old = constant_d + 1_new

Maybe we should also rename them, removing "d" they don't indicate a base elevation anymore.

Happy to discuss thoughts on this. If it should be an option or not etc.

raoulcollenteur:
@rubencalje what do you think of this idea of making it an additive parameter? Would that improve the performance on the model you tested in Italy @mbakker7 ? 

ps. naming was changed already.

rubencalje:
The additive parameter seems like a good idea if you assume that the second drainage level is always higher than the first drainage level.

However, these models are often applied in areas where the groundwater level is high relative to the surface elevation. In such locations, seepage is common, and the base drainage level may even be above the surface level. Under wet conditions, the second drainage level becomes active to prevent the groundwater level from rising further. In these cases, the second drainage level may actually be lower than the first drainage level.

I suspect this situation occurs quite frequently (also in the model of @mbakker7), which means that representing the second drainage level as an additive parameter is not a good choice.

martinvonk:
Note that additive does not mean it can only be larger than a certain value. It means it is added on top of the other parameter. The additive parameter could also become negative (dependent on the bounds) which means that you can stimulate levels where you a second drainage base below the first one.

dbrakenhoff:
I guess the advantage of that could then be that you can optionally enforce d1>d2 or the other way around? Is that something a user would consider doing?

rubencalje:
Good points, you can then bound the difference, or make sure the difference is always negative or positive. But I also like the idea of assigning absolute values to the two drainage base parameters, as they may not be very correlated and users can derive them visually easily (at least the second one). Maybe we can add an option `additive` to TarsoModel and/or ThresholdTransform?

Issue #1167, opened 2026-04-19
Let TarsoModel `d0` make use of model `constant_d`


martinvonk:
### Describe the proposed feature
TarsoModel now makes use of two "drainage level" parameters. d0 and d1. d0 is the base elevation and d1 is the level above which the threshold nonlinearity occurs. But since we also have a constant in the model, can't we make use of the constant_d of the model, instead of d0?

I think this would make it simpler and assist with not having to delete the constant with every TarsoModel.

raoulcollenteur:
I still really like this idea if it would work. Is it not simple to implement and test, and, if good, release with 2.0?

martinvonk:
> I still really like this idea if it would work. Is it not simple to implement and test, and, if good, release with 2.0?

I think this should now be pretty easy since the TarsoModel knows the model (and its .constant) exists.

martinvonk:
from #1291 
> This is actually quite complicated because the parameters during optimzation are not known to the model. So you can't just do: model.get_parameters in the TarsoModel to get the model.constant parameter value.

dbrakenhoff:
> > This is actually quite complicated because the parameters during optimzation are not known to the model. So you can't just do: model.get_parameters in the TarsoModel to get the model.constant parameter value.

Can't you update the name column for the `constant_d` parameter in `Model._parameters`? Or does the pastas world break apart when you try that?

PR #1291, opened 2026-07-20
Let `TarsoModel` make use of the `Model` if parsed

martinvonk:
# Short description
This PR updates the TarsoModel such that the Model.constant can be used. This way you don't need to call, model.del_constant before adding a TarsoModel.

- [x] related to issue #1167 and #1168 
- [x] is documented
- [x] type hints
- [x] tests added or expanded


martinvonk:
Something I don't really understand; whe changed the behavior of Pastas when initializing a StressModel in #1270. One of the benefits was that the TarsoModel could then use the model(.oseries). However, the TarsoModel does not make use of the model now. So I changed that here as well as it is required for the constant. But I thought that was fixed in #1270.

dbrakenhoff:
> However, the TarsoModel does not make use of the model now. So I changed that here as well as it is required for the constant. But I thought that was fixed in https://github.com/pastas/pastas/pull/1270.

Did you consider backwards compatibility here? E.g. we want to continue to support the old `oseries` input until 2.4?

martinvonk:
So I gave up on making the model constant part of the TarsoModel for now. This is actually quite complicated because the parameters during optimzation are not known to the model. So you can't just do: model.get_parameters in the TarsoModel to get the model.constant parameter value.

For now this is an improvement because if the Model is parsed the TarsoModel actually can use it to estimate the initial parameters. This actuall was one of the main reasons to make the syntax change. So now it is actually happening.

Also I made the example notebook simpler.

pastas/stressmodels.py
```diff
@@ -2488,24 +2488,40 @@ def __init__(
...
+            self.dmin = float(dmin)
+            self.dmax = float(dmax)
+        elif model is not None:
+            oseries = model.oseries.series
+            self.dmin = float(oseries.min())
+            self.dmax = float(oseries.max())
+        elif oseries is not None:
+            self.dmin = float(oseries.min())
+            self.dmax = float(oseries.max())
+        elif dmin is None or dmax is None:
```
raoulcollenteur: I think this should be self.dmin and self.dmax now

raoulcollenteur: Alternatively you could remove and self.dmin/dmax in the lines above and assign to self after all the check like before.

raoulcollenteur (approved):
I think this should be good after you fixed that small thing for the warning

PR #1294, opened 2026-07-21
fix #1285

### LINKED ISSUE #1285: Recursion error when calling ps.LeastSquares? And noise=True in LmfitSolve?
I am getting recursion errors when calling `ps.LeastSquares` on the dev branch. 

And I am getting an error running an LmfitSolve() because noise=True even though the model has no noisemodel. 

I think I have the fixes, but I need to know no pending PRs are already addressing these issues?

dbrakenhoff:
- fix deprecated aliases in init to avoid recursion error
- remove noise arg in lmfit

### DIFF
--- a/pastas/__init__.py
+++ b/pastas/__init__.py
@@ -36,7 +36,7 @@ from pastas.rfunc import (
     Polder,
     Spline,
 )
-from pastas.solver import EmceeSolve, LeastSquares, LmfitSolve, likelihood, timer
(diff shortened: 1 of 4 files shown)


raoulcollenteur (approved):

raoulcollenteur:
Makes sense! Strange that I never to an error or warning here locally.. Let's merge after the tests ran successfully.

PR #1295, opened 2026-07-21
remove .py files and unused .pas files

### LINKED ISSUE #1292: (Re)move .py files in docs/examples folder
### Describe the task

Right now, all the examples are documented in ipynb files. However, we have a few historical files in .py format. I suggest to remove those that are double, and change those we want to keep to .ipynb formats and make them available on the docs as well.

### Remove (already in .ipynb)
- [x] example.py
- [x] example_snow.py
- [x] example_step.py
- [x] example_tarso.py
- [x] example_uncertainty.py
- [x] example_well.py
- [x] example_transform.py
- [x] example_freq_obs.py
- [x] example_time_weighted_resample.py

### To ipynb
- [x] example_logger.py

### Other (Reason)
- [x] example_spline.py ( to docstring, example very short anyway no explanation plus not a feature we want to promote)
- [x] example_timestep_weighted_resample.py (deprecated method)
- [x] example_docs.py (only one to keep, because it is the example shown on the docs :))

raoulcollenteur:
# Short description

Remove redundant .py files in examples. 

- [x] closes issue #1292
- [x] is documented


raoulcollenteur:
This PR is ready for review. Issue #1292 lists the changes. Basically most were double anyways, one new notebook was added, and one example added to an existing notebook. I think this also makes the tests slightly quicker again as there is less to run :)

martinvonk (commented):
Can you quantify the coverage before and after this change? Maybe we can transform (some of these) these into tests if they are not covered anymore. 

But i agree getting rid of these files is nice.

raoulcollenteur:
You're starting to miss Codacy? ;-)

I suspect it actually doesn't change because all of the removed stuff was already somewhere else as well. 

I'll run a local test and compare with the current dev branch

dbrakenhoff (approved):

raoulcollenteur:
I updated the GH CI action to print the coverage, probably comes in handy in the future as well : https://github.com/pastas/pastas/actions/runs/30792748279/job/91619567487

It now reports 85%, which is similar to the dev branch but it is difficult to compare because many files are removed that had an almost 100% coverage.

I made a final small change to remove the change in the time_series notebook in the user guide, because I now found that code in prepare_timeseries ipynb as well.

PR #1299, opened 2026-07-30
Support real and complex in numba recharge funcs

dbrakenhoff:
This allows numba to compile njitted functions in real or complex depending on the input types. This is faster when inputs are all real. Previous iteration caused njitted functions to always perform complex arithmetic even for only real inputs. 

Squash and merge when reviewed,


pastas/recharge.py
```diff
@@ -383,41 +381,49 @@ def get_root_zone_balance(
...
+        sr = np.zeros(n + 1, dtype=dtype)
+        sr[0] = 0.5 * srmax
+        ea = np.zeros(n, dtype=dtype)
+        r = np.zeros(n, dtype=dtype)
+        q = np.zeros(n, dtype=dtype)
+
+        lp_scaled = lp * srmax
 
         for t in range(n):
-            # Make sure the solution is larger than 0.0 and smaller than sr
```
martinvonk, line 395: maybe add back the comments, i think those are nice

martinvonk (approved):
looks good! maybe add back the comments in the code because those help.

PR #1300, opened 2026-07-30
Improve pastas 2.0 performance

dbrakenhoff:
# Short description
- Reinstate sim_index cache through internal `ml._sim_index` attribute (view is available through `ml.sim_index` but cannot be set). Sim index is recomputed on change to tmin/tmax/freq/warmup and cleared whenever `ml.set_settings` is called. 
- Reinstate time_offset caching through a similar internal attribute `ml.time_offset` to avoid looping over all stressmodels on each access. 
- Introduce cached interpolation indices and weights during simulation. 
- Avoid computing residuals twice when using a noisemodel by adding a `res` kwarg to `ml.noise()` and `ml.noise_weights()`
- Access internal _settings instead of property to avoid repeated copies of settings dict. 
- Avoid building named tuples each time `.stress` attributes are accessed
- Avoid always casting index unit to `"us"` and set units when building an index. 
- Compute simulate using numpy arrays (only in `ml.simulate()`, stressmodels still return pandas series) and build pandas object afterwards.


pastas/model.py
```diff
@@ -524,27 +515,46 @@ def simulate(
...
             p = p.values
 
-        sim = Series(data=np.zeros(sim_index.size, dtype=float), index=sim_index)
+        sim_values = np.zeros(
+            sim_index.size, dtype=np.complex128 if np.iscomplexobj(p) else np.float64
+        )
 
         istart = 0  # Track parameters index to pass to stressmodel object
+        sim_index_min = sim_index.min()
+        sim_index_first = sim_index[0]
```
martinvonk: sim_index_min and sim_index_first are hopefully the same.

pastas/model.py
```diff
@@ -524,27 +515,46 @@ def simulate(
...
-                tmin=sim_index.min(),
+                tmin=sim_index_min,
                 tmax=tmax,
                 freq=freq,
                 dt=dt,
             )
-            sim = sim.add(contrib)
+            same_grid = (
+                contrib.index.size == sim_index.size
+                and contrib.index.size > 0
```
martinvonk: Can't we do: sim.index.difference(contrib.index).empty() or something?

pastas/model.py
```diff
@@ -524,27 +515,46 @@ def simulate(
...
+                and contrib.index[0] == sim_index_first
+                and contrib.index[-1] == sim_index_last
+            )
+            if not same_grid:
+                raise ValueError(
+                    f"Simulation of stressmodel {sm.name} does not match the simulation "
+                    "index. Check the settings of the stressmodel and/or the model "
+                    "settings."
+                )
+            sim_values += contrib.to_numpy(copy=False)
```
martinvonk: just out of curiosity:
is += really fastest, or could we also do a numpy.stack([..., ...]).sum()

pastas/model.py
```diff
@@ -1347,48 +1379,104 @@ def set_oseries(
...
+        """Time offset."""
+        return (
+            self._get_time_offset(self.settings["freq"])
+            if self._time_offset is None
+            else self._time_offset
+        )
+
+    @time_offset.setter
+    def time_offset(self, _: Timedelta) -> None:
+        raise AttributeError("time_offset is a read-only property.")
```
martinvonk: this is not necessary right with a property?

pastas/model.py
```diff
@@ -1347,48 +1379,104 @@ def set_oseries(
...
+            return sim_index
+
+    @property
+    def sim_index(self) -> DatetimeIndex:
+        """Simulation index."""
+        return self._sim_index
+
+    @sim_index.setter
+    def sim_index(self, _: DatetimeIndex) -> None:
+        raise AttributeError("sim_index is a read-only property.")
```
martinvonk: same here

pastas/model.py
```diff
@@ -1347,48 +1379,104 @@ def set_oseries(
...
+                tmin=tmin - warmup, tmax=tmax, freq=freq, time_offset=self.time_offset
+            )
+            if settings_match:
+                self._sim_index = sim_index
+            return sim_index
+
+    @property
+    def sim_index(self) -> DatetimeIndex:
+        """Simulation index."""
+        return self._sim_index
```
martinvonk, line 1441: do we want sim_index to be public?

pastas/model.py
```diff
@@ -1347,48 +1379,104 @@ def set_oseries(
...
+
+    @sim_index.setter
+    def sim_index(self, _: DatetimeIndex) -> None:
+        raise AttributeError("sim_index is a read-only property.")
+
+    def _get_interpolation_indices_weights(self, sim_index, obs_index) -> Series:
```
martinvonk: missing the type hints, also maybe we could do without this method. additionally, DatetimeIndex should have keyword argument name `tindex`

pastas/model.py
```diff
@@ -1583,7 +1669,9 @@ def get_parameters(self, name: str | None = None) -> ArrayLike:
         """
         # select parameters from appropriate stressmodel or noisemodel
         parameters = (
-            self._parameters.query("name == @name") if name else self._parameters
+            self._parameters.loc[self._parameters["name"] == name]
```
martinvonk, line 1635: rip query 💔

pastas/model.py
```diff
@@ -2117,7 +2204,7 @@ def _get_file_info(self) -> dict[str, Any]:
 
         try:
             file_info["owner"] = getlogin()
-        except Exception as e:
+        except Exception as e:  # noqa: BLE001
```
martinvonk: I guess we could find out what error happens here.. Instead of using the bare Exception

pastas/timeseries_utils.py
```diff
@@ -902,8 +901,46 @@ def resample(
...
-    if hasattr(tindex, "as_unit"):  # pandas >= 3.0
+    if hasattr(tindex, "as_unit") and tindex.unit != "us":  # pandas >= 3.0
         # In pandas 3.0, the default resolution for newly created datetime-like objects
         # is datetime64[us] (microseconds)
         tindex = tindex.as_unit("us")
     return tindex.view("int64")
+
+
+def _get_interpolation_weights(
+    sim_index: DatetimeIndex, obs_index: DatetimeIndex
```
martinvonk: see comment about `tindex`

pastas/stressmodels.py
```diff
@@ -2840,12 +2845,13 @@ def __init__(
         self.rfunc2 = rfunc2
         self.tchange = Timestamp(tchange)
         self.set_init_parameters()
+        self._nt = namedtuple("StressesTuple", ["stress"])
```
martinvonk: isn't this one inherited from super()?

pastas/solver/objective_function.py
```diff
@@ -39,10 +39,9 @@ def misfit(
         The calculated residuals or noise, optionally with separate components.
     """
     # Get the residuals or the noise
+    rv = ml.residuals(p)
```
martinvonk: maybe rename `rv` to `res`

dbrakenhoff:
Summary, pastas 2.0 mostly faster, and fit results equal or better :smiley: 

Comparison in solve time between 2.0 vs 1.14:

comparison_scatter_solve_time_2 0 0_vs_1 14 0

Comparison of model fit:

comparison_scatter_nse_2 0 0_vs_1 14 0

Details:
comparison_model_grouped_bar_solve_time

dbrakenhoff:
@martinvonk the misfit function uses `ml=` instead of `model=` still, perhaps you can fix that as well? I thought we decided on `model=` everywhere?

pastas/model.py
martinvonk: replaced by sim.index.equals(contrib.index)

pastas/model.py
martinvonk: fixed

pastas/model.py
martinvonk: += is faster :)

pastas/model.py
martinvonk: removed

pastas/model.py
martinvonk: removed

pastas/model.py
martinvonk: fixed the type hint but still public..

pastas/model.py
martinvonk: removed this method and put the logic in residuals. though we could consider making: simulate_on_obs()

pastas/model.py
```diff
@@ -531,12 +529,7 @@ def simulate(
                 freq=freq,
                 dt=dt,
             )
-            same_grid = (
-                contrib.index.size == sim_index.size
-                and contrib.index.size > 0
-                and contrib.index[0] == sim_index_first
-                and contrib.index[-1] == sim_index_last
-            )
+            same_grid = sim_index.equals(contrib.index)
```
dbrakenhoff: I believe this same_grid method was a vaguely AI guided performance choice, I would assume the index comparison is pretty fast, but I'm not sure. The other risk (though I think that has been mitigated by casting to us everywhere in pastas) is that index units might be different, though I'm not sure the equals check also checks for that?

pastas/model.py
martinvonk: fixed by OSError

pastas/timeseries_utils.py
martinvonk: fixed

pastas/stressmodels.py
martinvonk: fixed

pastas/solver/objective_function.py
martinvonk: fixed

pastas/model.py
martinvonk: Maybe we could also do:
same_grid = np.array_equal(sim_index.values, contrib.index.values)

dbrakenhoff:
(sorry, pressed the wrong button accidentally, didn't mean to perform the merge)

pastas/model.py
martinvonk: Maybe we could also just remove this check because we only check to raise a an error. But you would get an error on the dimensions anyway.

martinvonk (approved):
I approve. Nice changes :)

Would be good if @raoulcollenteur also had a look.

pastas/model.py
dbrakenhoff: Martin and I agreed that we can get rid of this check. The only downside is potentially if someone makes a stressmodel with a simulate that returns results for a subset of the modeling period. In that case the arrays don't align and you will get some a numpy mismatching arrays error: `ValueError: operands could not be broadcast together with shapes ... ` 

We semi-replaced it with a NaN check on the contributions. This gives you an informative error message if rfunc parameters cause the step response to blow up and produce infinities or whatever. 

All other error handling is taken care of by `stress.update_series()` which ensures the stress spans the modeling period, and if it cannot, raises an informative error.

pastas/model.py
dbrakenhoff: correct, I didnt realize that :P

pastas/model.py
dbrakenhoff: Yes, because sim_index used to be public, we decided to leave it there for now. Don't think it is ever used, but who knows.

raoulcollenteur (approved):

raoulcollenteur:
Thanks for all the work on this one! Looks good to me, I like passing the res in the noise, we should have done that before (also to prevent double logger messages).

raoulcollenteur:
> Summary, pastas 2.0 mostly faster, and fit results equal or better 😃
> 
> Comparison in solve time between 2.0 vs 1.14:
> 
> comparison_scatter_solve_time_2 0 0_vs_1 14 0
> Comparison of model fit:
> 
> comparison_scatter_nse_2 0 0_vs_1 14 0
> Details: comparison_model_grouped_bar_solve_time

We'll take this one for the Pastas 2.0 presentation ;-)

PR #1310, opened 2026-08-03
Create simulate method that returns simulation on observations index

### LINKED ISSUE #1306: Add model simulate method that returns interpolated simulation values on observations
Now this logic is entirely in `Model.residuals`.  However, for some solvers and methods it would also be usefull is this logic is available elsewhere. Also it would clean up the code.

### Expected usage

```python
def simulate_interpolated(self, ..., ...)
    obs = ml.observations(tmin, tmax)
    sim = ml.simulate(tmin, tmax)
    # interpolate linearly on obs
    ...

def residuals()
    sim = self.simulate_interpolated()
    res = obs.subtract(sim)
    ...
```

martinvonk:
# Short description
This PR introduces the `simulate_on_observations` method. It returns the simulated values at the same timestamps as the `model.obsevations`. If there are observations at a higher frequency than the model frequency, linear interpolation is applied. If not than no interpolation is required. 

This cleans the `residuals` method up significantly. Also it is useful for solvers that only need the simulated values, and not the residuals.

Closes issue #1306


martinvonk:
We can also merge this later. But I think it is nice to have in 2.0. Especially when updating the `pastas-plugins` for pastas 2.0 as well. 

I'm open for better names. Though I think: `simulate_interpolated` a bad name because interpolation is not what happens per se.

raoulcollenteur:
Can you run a check to make sure this PR does not make it slower?

martinvonk:
I don't notice any difference locally for a simple model solve:
dev: ~0.075 s
sim_int: ~0.075 s

raoulcollenteur (approved):

PR #1317, opened 2026-08-07
Make deprecations for Pastas 2.0 consistent

### LINKED ISSUE #1276: `DeprecationWarning` not shown to users by default, change to `FutureWarning`
`DeprecationWarning` has a special status in Python: it is intended for developers, not end users. Since Python 3.7 (following PEP 565), it is ignored by default for imported libraries, which is why most Pastas users will never see it.

Because we want users to see the warning, I think we should change the warning type to `FutureWarning` in the `PastasDeprecationWarning` and `deprecate_args_or_kwargs` decorators. Or we should change the warnings to logger.warn, so users see it, and it is included in log-files people may generate with their sctrips.

martinvonk:
# Short description
Makes the deprecations for Pastas 2.0 consistent:
- Deprecation is either at version 2.4 for future depreactions
- Else it is version 2.0 or earlier

Rename `PastasDeprecationWarning` to `deprecate_class_func_or_method` because we don't use DeprecationWarning anymore but future warnign. Also it is more consistent with the `deprecate_args_or_kwargs`. 

- [x] closes issue #1276, #701

### DIFF
--- a/pastas/__init__.py
+++ b/pastas/__init__.py
@@ -12,7 +12,7 @@ from pastas import check, extensions, forecast, solver, stats
 from pastas._options import options
 from pastas.dataset import list_datasets, load_dataset
 from pastas.decorators import (
-    PastasDeprecationWarning,
+    deprecate_class_func_or_method,
     get_use_cache,
     get_use_numba,
     set_use_cache,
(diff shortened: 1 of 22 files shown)


martinvonk:
This fixes two TODO's from the #1176:
- fix all the deprecations from "2.3.0" to "2.4.0"
- check the list of #701 

I also renamed `PastasDeprecationWarning` to `deprecate_class_func_or_method` but we could also not do that. Or find a better name.

dbrakenhoff (approved):

PR #1322, opened 2026-08-12
fix error in formulas on rtd hantush example

### LINKED ISSUE #1321: Some equations in the Hantush example notebook are not shown in rtd
The latex below cannot be displayed in readthedocs and the error: 'Erroneous nesting of equation structures' is shown instead.

```
$$
\begin{align*}
A &= A^\prime \text{K}_0 \left( 2 r \sqrt{b^\prime} \right)\\
a &= a \\
b &= r^2 b^\prime
\end{align*}
$$
```

OnnoEbbens:
According to AI this should fix it

### DIFF
--- a/doc/examples/hantush_response.ipynb
+++ b/doc/examples/hantush_response.ipynb
@@ -90,11 +90,11 @@
     "HantushWellModel function are:\n",
     "\n",
     "$$\n",
-    "\\begin{align*}\n",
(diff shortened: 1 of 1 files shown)


OnnoEbbens:
This worked, the formulas are now visible on rtd

martinvonk (approved):

PR #1330, opened 2026-08-14
Fix brentq failure in Hantush response check for small gain values

### LINKED ISSUE #1328: Error in true response tmax check with extremely small gain parameter in WellModel
I have a WellModel with a very small parameter A. When I solve the model I get an error because the memory check tries to use `brentq` to get the tmax. This method fails for very small numbers.

### Code to reproduce
minimal reproduction
```python
import pastas as ps
ps.Hantush(approximate_tmax=False).get_tmax([-3.33659342e-180 , 3.99179273e-002 , 3.65719918e+004])
```

full model reproduction (see attached model file (rename .pas to .txt to make it work)).

[ml.txt](https://github.com/user-attachments/files/31029062/ml.txt)

```python
from pastas.io import load
ml = load('ml.pas')
ml_swm.solve(freq="14D", tmin="1990-01-01", tmax="2025-01-01")
```

Pastas     : 2.0.0rc1
Python     : 3.13.13
Numpy      : 2.4.6
Pandas     : 2.3.3
Scipy      : 1.18.0
Matplotlib : 3.11.1
Numba      : 0.67.0rc1

The error originates in the check module. I cannot find a way to turn of the check module when solving a model, maybe that would be nice to have as well.

martinvonk:
# Short description
The true response length is calculated in the `check` module for `Hantush` and `HantushWellModel`. The true length corresponds to the point where the step response reaches the gain multiplied by the specified cutoff.

However, when the gain $A$ is extremely small, numerical underflow causes the `brentq` root-finding algorithm to fail. Since the cutoff calculation is normalized, the actual gain value is not required. Setting $A=1.0$ resolves the underflow issue and allows `brentq` to converge reliably.

- [x] closes issue #1328

### DIFF
--- a/pastas/rfunc.py
+++ b/pastas/rfunc.py
@@ -1059,14 +1059,21 @@ class Hantush(RfuncBase):
         tmax = wrightomega(log_z).real * a
         return tmax
 
-    def _f_step(self, t: float, A: float, a: float, b: float, cutoff: float) -> float:
-        """Objective function for root finding (t varies, other params fixed)."""
+    def _f_step(self, t: float, a: float, b: float, cutoff: float) -> float:
+        """Objective function for root finding (t varies, other params fixed).
+
+        This function computes the difference between the step response at time t
+        and the specified cutoff. It is used in root-finding algorithms to determine
+        the time t at which the step response reaches the cutoff value.
+        """
         t_arr = np.array([t], dtype=float)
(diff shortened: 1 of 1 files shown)


dbrakenhoff (approved):

OnnoEbbens (commented):
good job!

PR #1336, opened 2026-08-24
Enhance Emcee initialization with nsteps and update related notebooks

### LINKED ISSUE #1335: Set `nsteps` in Emcee init
### Describe the task
This can now only be parsed to `model.solve` which is inconsistent with the LeastSquares solver.

martinvonk:
# Short description
Makes behavior across the solvers more consistent

- [x] closes issue #1335


martinvonk:
@raoulcollenteur can we move the GaussianLikelihood and the GausianLikelihoodAr1 into `objective_function.py`? Now they are in `likelihood.py` which gives the awkward code:

```python
from pastas.solver.likelihood import GaussianLikelihood

ps.Emcee(
   objfunction=GaussianLikelihood,
)
```
instead of 
```python
from pastas.solver.objective_function import GaussianLikelihood

ps.Emcee(
   objfunction=GaussianLikelihood,
)
```

I think the latter is much more logical.

martinvonk:
> @raoulcollenteur can we move the GaussianLikelihood and the GausianLikelihoodAr1 into `objective_function.py`? Now they are in `likelihood.py` which gives the awkward code:
> 
> ```python
> from pastas.solver.likelihood import GaussianLikelihood
> 
> ps.Emcee(
>    objfunction=GaussianLikelihood,
> )
> ```
> 
> instead of
> 
> ```python
> from pastas.solver.objective_function import GaussianLikelihood
> 
> ps.Emcee(
>    objfunction=GaussianLikelihood,
> )
> ```
> 
> I think the latter is much more logical.

I now did this in https://github.com/pastas/pastas/pull/1336/commits/0db20e68ffbd5998129110ccf8f4a8ae7b8e15e4. Feel free to revert the commit. We could maybe also create an `objective` directory inside the `solver` directory with an `function.py` and `likelihood.py` if you want to keep things seperate.

raoulcollenteur:
Thanks for the PR. Looks good, but I'd like to keep the likelihood function in a separate file for now. I understand it feels strange to have `obj_function=likelihood` but otherwise the solution would be to change `obj_function` to likelihood instead. However, I'd say we just stick to how it is right now. 

So, please revert https://github.com/pastas/pastas/commit/0db20e68ffbd5998129110ccf8f4a8ae7b8e15e4 and then we can merge this.

raoulcollenteur (approved):
after reverting https://github.com/pastas/pastas/commit/0db20e68ffbd5998129110ccf8f4a8ae7b8e15e4, I approve.

martinvonk:
How about this @raoulcollenteur current implementation?
This now works:
`ps.objective.misfit`
`ps.objective.likelihood`
`ps.objective.GaussianLikelihood`
`ps.objective.likelihood.GaussianLikelihood`

raoulcollenteur (approved):

